PR #4, opened 2022-09-23
Fix SIZES for torch tensors

sgugger:
This PR fixes the SIZE in bytes of `torch.int64` and adds a few other dtypes. With it I can serialize and deserialize a classic BERT model.


Narsil (approved):
LGTM

PR #5, opened 2022-09-26
Adding slice support for lazy loading.

Narsil:


safetensors/src/lib.rs
```diff
@@ -285,6 +301,27 @@ pub enum Dtype {
     F64,
 }
 
+impl Dtype {
+    /// Gives out the size (in bytes) of 1 element of this dtype.
+    pub fn size(&self) -> usize {
+        match self {
```
Narsil, line 307: We should probably re-expose that to Python so we don't have to write it over there  (jsut to avoid maintaining 2 lists)

safetensors/src/slice.rs
```diff
@@ -0,0 +1,416 @@
...
+};
+
+/// Error representing invalid slicing attempt
+#[derive(Debug)]
+pub enum InvalidSlice {
+    TooManySlices,
+}
+
+#[derive(Debug)]
+pub enum TensorIndexer {
```
Narsil, line 13: This is taken from https://github.com/LaurentMazare/tch-rs/blob/f691db1708c12db02f0836348f6ba341d3a0c40d/src/tensor/index.rs

I removed the unnecessary parts (for now) as it simplifies the code for now.

I'm leaving parts in comments because they might become important again (`[:, 2]` is `Select` for instance and could be useful as it would avoid a reshape, still unsure if I want to support here or in python only.
`IndexSelect` is ` [:, torch.Tensor(2,3 , 5, 6)]`  which I'm not sure we need/want support (first of all this exact form would require tensor support, as done in `tch-rs`. So heavy adaptation is probably needed.
In most cases in distributed settings I would say `tensor[:, ::3]` would be the most complex query and it would be written `(0..n).step_by(3)` in rust and is actually a `StepBy` object.

Anyway, all this complexity is left for later.

safetensors/src/slice.rs
```diff
@@ -0,0 +1,416 @@
...
+    view: &'data TensorView<'data>,
+    indices: Vec<(usize, usize)>,
+}
+
+impl<'data> SliceIterator<'data> {
+    pub(crate) fn new(
+        view: &'data TensorView<'data>,
+        slices: Vec<TensorIndexer>,
+    ) -> Result<Self, InvalidSlice> {
+        // Make sure n. axis does not exceed n. of dimensions
```
Narsil, line 207: This is the meat of the algorithm

safetensors/src/slice.rs
```diff
@@ -0,0 +1,416 @@
...
+    }
+}
+
+impl<'data> Iterator for SliceIterator<'data> {
+    type Item = &'data [u8];
+
+    fn next(&mut self) -> Option<Self::Item> {
+        // TODO
+        let (start, stop) = self.indices.pop()?;
+        Some(&self.view.data[start..stop])
```
Narsil, line 280: Because we calculate all indices in advance this becomes extremely simple.

I think this is OK in first approximation, but for iterators it's probably nicer if we can avoid the full calculation and storage upfront and implement only the next logic.

This seems simpler for now. We will fix if there is actually a performance issue I think

mishig25:
Could you also add a slicing test on python bindings?

safetensors/src/slice.rs
```diff
@@ -0,0 +1,418 @@
...
+}
+
+impl_from_range!(Range<usize>);
+impl_from_range!(RangeFrom<usize>);
+impl_from_range!(RangeFull);
+impl_from_range!(RangeInclusive<usize>);
+impl_from_range!(RangeTo<usize>);
+impl_from_range!(RangeToInclusive<usize>);
+
+pub trait IndexOp<'data, T> {
```
sgugger, line 70: I'm not seeing where this trait is used for now.
Really not a big fan of the super short name `i` for its main method.

Narsil:
> Could you also add a slicing test on python bindings OR there are more things need to be figured out atm for Python?

https://github.com/huggingface/safetensors/pull/6

Here is what I started. The API is definitely not nice now, but at least it works.
The biggest issue to make this nice on python side, is that PyO3 custom structs are not allowed to have lifetimes (because Python does not know aobut lifetimes)

In order to make slices nice to use we need some kind of object to slice into `tensor[:, ::3]` for instance but since `tensor`cannot be `TensorView` we need another mecanism to get it working.

sgugger (commented):
Looking nice!

safetensors/src/slice.rs
Narsil: The trait is here only to support "multiple signature" functions (which isn't allowed).

This works by creating a trait and 1 function that works on anyone that has that trait, and then implement the trait for all types of arguments we want to handle.

I really copied this: https://docs.rs/tch/latest/tch/index/index.html which aims to make something as easy to use as pytorch.
And we need some sort of trait because `Range`, `RangeFull` at least are different types, so we cannot have the sliced data accept a concrete type directly.

The "no trait" route is doing an enum:
```rust
enum TensorIndexer{
    Full(RangeFull)
    Range(Range<usize>)
    ...
}
```
which does work but it much more tedious on the caller.

As for the name of the function, we definitely can make it more explicit here. It's intentionally small in `tch-rs` because it's meant to be used directly and mimic Python indexing (I can dig it up, but using real slicing in Rust, has some kind of limitations because of the trait it asks to implement).

Here it's the same. For Python users, they should never see it, but for rust users doing something like

```rust
let tensorview =  SafeTensors::deserialize(&content).unwrap().tensor("mytensor");
let buffer: Vec<u8> = tensorview.i( (.., ..2,3..4) ).unwrap()..flat_map(|b| b.to_vec()).collect();
```

Feels better than the non trait version:
```rust
let tensorview =  SafeTensors::deserialize(&content).unwrap().tensor("mytensor");
let buffer: Vec<u8> = tensorview.get_sliced_data( vec![
  TensorIndexer::Narrow(Bound::Unbounded, Bound::Unbounded),  
  TensorIndexer::Narrow(Bound::Unbounded, Bound::Excluded(2)),
  TensorIndexer::Narrow(Bound::Included(3), Bound::Excluded(4)),
).unwrap()..flat_map(|b| b.to_vec()).collect();
```

So purely ergonomics. Naming it `slice` (or some other name) is really doable. I kind of enjoy `i` in `tch-rs` though.

sgugger: Thanks for the explanation! Would prefer slice to `i` personally (anything more descriptive than `i` ;-) ) but that's my personal taste.

Narsil: Moved to `slice`. No hugely strong opinions yet.

safetensors/src/lib.rs
```diff
@@ -159,6 +162,9 @@ impl<'data> SafeTensors<'data> {
             std::str::from_utf8(&buffer[8..8 + n]).map_err(|_| SafeTensorError::InvalidHeader)?;
         let metadata: Metadata = serde_json::from_str(string)
             .map_err(|_| SafeTensorError::InvalidHeaderDeserialization)?;
+        // TODO validate metadata
```
McPatate, line 165: Are you planning on having versions of safetensors files ?

McPatate (approved):

PR #6, opened 2022-09-26
Python slice API proposition

Narsil:
Much better API:

```python
with PySafeFile(self.local, framework="pt") as f:
      tensor = f["test"][:, :, 1:2]
```
Design:
- **Context manager**
This is meant to open a file, we need to close it, and this is the Python way (lazy load on a real buffer instead of a file doesn't have real advantages since torch is already able to not copy a single thing with striding in the first place)
- **Framework arg**:
I played with other ideas, as if the code was framework agnostic it would be easier on users on any kind of failure and to adapt directly within Python.
However, `memoryview` doesn't handle dtype, and `bfloat16` is only handled by pytorch, so using it as intermediary is not really as viable as I thought.
-  **f["test"]**
This object is ~equivalent to `TensorView<'data>` except we cannot have a lifetime on it, so it owns the Mmap instead.
This is *not* as `torch.Tensor` at this point, only when you slice into does it become `torch.Tensor`.
- **[:, :, 1:2]**
This is where we can calculate how to slice into the data, creating the buffer iteratively (so only once) as `PyByteArray` so we do actually avoid allocating the entire buffer here (hopefully). In order to avoid shape shenanigans, we simply return the `torch.Tensor` directly.

Pros:
- handles a lot of boilerplate for caller.
- Feels more pythonesque

Cons:
- Unsure about soundness of keeping Mmap struct (in PySafeFile), it's meant to be used as context manager, so it should clean up after itself.
- Lots more boilerplate in the binding code.

### DIFF
--- a/bindings/python/tests/test_pt_comparison.py
+++ b/bindings/python/tests/test_pt_comparison.py
@@ -1,5 +1,5 @@
 import unittest
-from safetensors.torch import save_file, load_file, load
+from safetensors.torch import save_file, load_file, load, to_dtype
 from huggingface_hub import hf_hub_download
 import torch
 import datetime
@@ -30,6 +30,31 @@ class SafeTestCase(unittest.TestCase):
         save_file(self.data, outfilename)
         print()
         print("Serialization (safe) took ", datetime.datetime.now() - start)
+
+
+class SliceTestCase(unittest.TestCase):
+    def setUp(self):
+        self.tensor = torch.arange(6, dtype=torch.float32).reshape((1, 2, 3))
+        self.data = {"test": self.tensor}
+        self.local = "./tests/data/out_safe_pt_mmap.bin"
+        # Need to copy since that call mutates the tensors to numpy
+        save_file(self.data.copy(), self.local)
+
+    def test_deserialization_slice(self):
+        from safetensors.safetensors_rust import deserialize_file_slice
+
+        slices = (slice(None), slice(None), slice(1, 2))
+
+        item = deserialize_file_slice(self.local, "test", slices)
+        self.assertEqual(
+            item["data"],
+            b"\x00\x00\x80?\x00\x00\x80@",
+        )
+
+        dtype = to_dtype(item["dtype"])
+        tensor = torch.frombuffer(item["data"], dtype=dtype).view(1, 2, 1)
+        self.assertTrue(torch.equal(tensor, torch.Tensor([[[1.0], [4.0]]])))
+        self.assertTrue(torch.equal(tensor, self.tensor[:, :, 1:2]))
 
 
 class TorchTestCase(unittest.TestCase):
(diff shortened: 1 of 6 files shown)


bindings/python/tests/test_pt_comparison.py
```diff
@@ -32,6 +32,31 @@ def test_serialization_safe(self):
...
+        self.local = "./tests/data/out_safe_pt_mmap.bin"
+        # Need to copy since that call mutates the tensors to numpy
+        save_file(self.data.copy(), self.local)
+
+    def test_deserialization_slice(self):
+        from safetensors.safetensors_rust import deserialize_file_slice
+
+        slices = (slice(None), slice(None), slice(1, 2))
+
+        item = deserialize_file_slice(self.local, "test", slices)
```
Narsil, line 48: This is the bad API.

sgugger:
> f["test"] This object is ~equivalent to TensorView<'data> except we cannot have a lifetime on it, so it owns the Mmap instead. This is not as torch.Tensor at this point, only when you slice into does it become torch.Tensor.

Ah, that's problematic as we also have a use case to get individual weights directly (probably much more than getting a slice) and doing 
```py
with PySafeFile(self.local, framework="pt") as f:
      tensor = f["test"][:]
```
does not seem like a super nice API for such a use case.

Other than this, the API looks super nice.

Narsil:
> Ah, that's problematic as we also have a use case to get individual weights directly (probably much more than getting a slice) and doing

Don't think there's an easy way out of this because a `torch.Tensor` has to own a buffer, and we can't figure out the minimal buffer before you slice (in case of lazy loading).

For loading all the tensors `load_file(filename)` should be the go-to IMO.

Also we probably need to add something like `f.keys()` to display available tensors to users.

Narsil:
> No I don't want to load all the tensors and blow up my RAM.

I see and understand (actually there seem to exist mecanism where we can send directly from disk to GPU https://devblogs.microsoft.com/directx/directstorage-api-available-on-pc/ that might be usable in the future )

What about something like this:

```python
with PySafeFile(self.local, framework="pt") as f:
      tensor = f["test"] # direct pytorch tensor
```

```python
with PySafeFile(self.local, framework="pt") as f:
      tensor = f.slice("test")[:, :, 1:2]
```

Wdyt ? Seems it solves most issues right ?

sgugger:
That works too!

Narsil:
I don't like `PySafeFile` but I'm hesitant to call it `open`. `safeopen` ? Any other suggestions ?

sgugger:
We should avoid using a Python reserved name since users might do a `from safetensors import *`. `safe_open` sounds good!

julien-c:
agree with proposed name, it is a nice name

Narsil:
I made the changes.

The big caveat I see with this proposal is that 
```
with PySafeFile(self.local, framework="pt") as f:
      tensor = f["test"][:, :, 1:2]

with PySafeFile(self.local, framework="pt") as f:
      tensor = f.slice("test")[:, :, 1:2]
```
Are only slightly different will the optimization benefits is missed for the first version.

@sgugger what do you think about `f.tensor("test")` instead of the direct slice (might make documenting + discoverability better).

sgugger:
Maybe `f.get_tensor` and `f.get_slice` then? But okay for me if there is no direct index, especially to avoid confusion like the ones you pointed out.

PR #9, opened 2022-09-27
PoC for additional metadata

sgugger:
This adds a new `Dict[str, str]` to the metadata in a `SafeTensors` objects, that can save any user metadata associated to some weight file in this format.

The main use right now is to save the format used when saving a model (PyTorch, TensorFlow etc) so I know on the Transformers side if a conversion of weights should be applied or not.

### DIFF
--- a/bindings/python/tests/test_simple.py
+++ b/bindings/python/tests/test_simple.py
@@ -1,5 +1,5 @@
 import unittest
-from safetensors.numpy import save, load, save_file, load_file
+from safetensors.numpy import save, load, save_file, load_file, read_metadata_in_file
 from safetensors.torch import load_file as load_file_pt, save_file as save_file_pt
 import numpy as np
 from huggingface_hub import hf_hub_download
@@ -14,15 +14,16 @@ class TestCase(unittest.TestCase):
 
         self.assertEqual(
             out,
-            b"""<\x00\x00\x00\x00\x00\x00\x00{"test":{"dtype":"I32","shape":[2,2],"data_offsets":[0,16]}}\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00""",
+            b"""[\x00\x00\x00\x00\x00\x00\x00{"__metadata__":{"format":"np"},"test":{"dtype":"I32","shape":[2,2],"data_offsets":[0,16]}}\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00""",
         )
 
         data[1, 1] = 1
         out = save({"test": data})
+        print(out)
 
         self.assertEqual(
             out,
-            b"""<\x00\x00\x00\x00\x00\x00\x00{"test":{"dtype":"I32","shape":[2,2],"data_offsets":[0,16]}}\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\00""",
+            b"""[\x00\x00\x00\x00\x00\x00\x00{"__metadata__":{"format":"np"},"test":{"dtype":"I32","shape":[2,2],"data_offsets":[0,16]}}\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00""",
         )
 
     def test_deserialization(self):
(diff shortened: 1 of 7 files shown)


safetensors/src/lib.rs
```diff
@@ -96,14 +97,14 @@ fn prepare<'hash, 'data>(
         tensors.push(tensor);
     }
 
-    let metadata: Metadata = Metadata(hmetadata);
+    let metadata: Metadata = Metadata(data_info.clone(), hmetadata);
```
sgugger, line 100: New at rust, so not sure if here the best is to explicitly clone or deref with `*` which might move the dictionary?

safetensors/src/lib.rs
```diff
@@ -204,14 +206,19 @@ impl<'data> SafeTensors<'data> {
     /// The tensor returned is merely a view and the data is not owned by this
     /// structure.
     pub fn names(&self) -> Vec<&'_ String> {
-        self.metadata.0.iter().map(|(name, _)| name).collect()
+        self.metadata.1.iter().map(|(name, _)| name).collect()
+    }
+
+    /// Returns additional metadata passed along by the user.
+    pub fn get_metadata(&self) -> HashMap<String, String> {
+        self.metadata.0.clone()
```
sgugger, line 214: Here add to clone otherwise rust was complaining that the value was moved. Could maybe be avoided by making the filed public in the metadata (which would have to be public as well). I don't expect this to be a big object anyway.

Narsil:
Thank you ! 

And congrats for the first rust PR !

I would suggest moving the structure `Metadata` to an actual struct, and ideally without breaking change (not because we can't at this point everything is up in the air but more because it feels more JSONy to have a "special" field)

Something like this:

```rust
#[derive(Debug, Deserialize, Serialize)]
pub struct Metadata {
  #[serde(skip_serializing_if = "Option::is_none")]
  #[serde(rename = "__metadata__")]
  metadata: Option<HashMap<String, String>>,
  #[serde(flatten)]
  tensors: HashMap<String, TensorInfo>,
}
  ```
  
  Here is one example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e7561254982212418f956549a538d2d8
  
  Another example (Option vs empty hashmap) : https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e7561254982212418f956549a538d2d8
  
  Aside from that the rest looks really nice !

File: safetensors/src/lib.rs (hunk 1 above)
Comment (Narsil) [reply safetensors/src/lib.rs:100]: You cannot deref.

More explicitly, if you deref the borrow checker is going to complain.
You have in your possession a `ref` so doing  `*data_info` is OK, but that's the owned value, which you are trying to give to `Metadata`, which this is not yours to give as you don't have ownership ! 

Here we don't really have much of a choice to clone here.
Other option would be to handle the lifetimes everywhere which is too complex for the benefit IMO.

File: safetensors/src/lib.rs [hunk 3]
```diff
@@ -171,7 +173,7 @@ impl<'data> SafeTensors<'data> {
     /// structure.
     pub fn tensors(&self) -> Vec<(String, TensorView<'_>)> {
         let mut tensors = vec![];
-        for (name, info) in &self.metadata.0 {
+        for (name, info) in &self.metadata.1 {
```
Narsil, line 176: Changing for a real struct will make that more readable.

`Metadata(HashMap<String, Tensor>)` is just an easy shortcut to name a "raw" type so that you can differentiate.

For instance you could have

`User(String)`
`Email(String)`

```
fn set_email(email: String)   # Potential bug, `set_email("myuser")` 
vs
fn set_email(email: Email)
```
So that you now have 2 different types which will make your code a tiny bit more readable + type safe (with the added cost of adding `.0` in some places).

Not something worth it most of the time.

safetensors/src/lib.rs
Narsil: You could also return a reference to the map instead:

```rust
pub fn get_metadata(&self) -> &HashMap<String, String> {
        &self.metadata.0
```
Should work.

Since this struct already has an associated lifetime (the lifetime of the buffer) you might need more lifetime annotations to explicitly tell the compiler that both lifetimes are the same.

```rust
pub fn get_metadata(&'data self) -> &'data HashMap<String, String> {
        &self.metadata.0
```

Since `SafeTensors` cannot live more than `'data` , if you are handing out a reference to someone, then that reference cannot outlive YOU (and therefore neither the longest `'data`) .

In general the borrow checker has simple rules which elide the lifetimes, but it might not work here.

Narsil (approved):
LGTM !

bindings/python/tests/test_simple.py
```diff
@@ -51,6 +51,8 @@ def test_numpy_example(self):
         loaded = load_file("./out.bin")
         self.assertTensorEqual(tensors, loaded, np.allclose)
 
+        metadata =
```
Narsil, line 54: Oops je pense ?

sgugger: Lol yes! Went too fast to remove.

PR #10, opened 2022-09-27
Big revamp:

Narsil:
- Remove `Tensor` to fuse with `TensorView` (the price to pay for shape
  cloning should be small enough that it shouldn't make too much of a 
  difference (less surface is easier to manage).
- Added `TensorView` sanity check at instantiation.
- Remove `deserialize_file`, now that we have `safe_open` facility.
- Tried to remove `torch` dependency on `numpy` (Still need the
`torch.Tensor` to bytes without copying facility for now).
- Revamped the speed tests to show metrics side by side. They need
to be taken with extreme caution, timings are never reproducible.


bindings/python/py_src/safetensors/torch.py
```diff
@@ -1,43 +1,69 @@
...
+    "U8": torch.uint8,
+    "BOOL": torch.bool,
+}
+
+
+def _getdtype(dtype_str: str) -> torch.dtype:
+    return _TYPES[dtype_str]
+
+
+def _view2np(safeview) -> Dict[str, torch.Tensor]:
```
sgugger, line 49: It's more of a view2torch here :-)

safetensors/src/tensor.rs
```diff
@@ -0,0 +1,451 @@
...
+    fn new(metadata: HashMap<String, String>, tensors: HashMap<String, TensorInfo>) -> Self {
+        Self { metadata, tensors }
+    }
+
+    /// Gives back the tensor metadata
+    pub fn tensors(&self) -> &HashMap<String, TensorInfo> {
+        &self.tensors
+    }
+
+    /// Gives back the tensor metadata
```
sgugger, line 172:
```suggestion
    /// Gives back the metadata
```

bindings/python/src/lib.rs
```diff
@@ -54,17 +65,22 @@ fn prepare<'a>(tensor_dict: HashMap<String, &'a PyDict>) -> PyResult<HashMap<Str
 fn serialize<'a, 'b>(
     py: Python<'b>,
     tensor_dict: HashMap<String, &'a PyDict>,
+    tensor_data: Option<HashMap<String, String>>,
```
sgugger, line 68: It's nothing to do with tensors here, so would just call it `metadata`.

bindings/python/src/lib.rs
```diff
@@ -54,17 +65,22 @@ fn prepare<'a>(tensor_dict: HashMap<String, &'a PyDict>) -> PyResult<HashMap<Str
...
     let pybytes = PyBytes::new(py, &out);
     Ok(pybytes)
 }
 
 #[pyfunction]
-fn serialize_file<'a>(tensor_dict: HashMap<String, &'a PyDict>, filename: &str) -> PyResult<()> {
+fn serialize_file(
+    tensor_dict: HashMap<String, &PyDict>,
+    filename: &str,
+    tensor_data: Option<HashMap<String, String>>,
```
sgugger, line 80: Same here

sgugger (approved):
Left a few nits but LGTM!

bindings/python/py_src/safetensors/torch.py
Narsil: Done.

bindings/python/src/lib.rs
Narsil: `Extra_data` ?

It's Metadata within Metadata at some point. 
(Here it's not confusing because metadata does not exist, but if we can have the name throughout the code base it's easier to read I think)

Narsil: I kept `metadata`. I'm being to nit-picky here (the field is name `__metadata__` which is makes sense. we have 2 kinds of metadata but it's not that horrible I think.

Narsil:
@sgugger Merging, feel free to try the new 

```python
with safe_open(..) as f:
    metadata = f.metadata()
```

PR #14, opened 2022-09-29
Use np.prod for compatibility with Python 3.7

sgugger:
`math.prod` was only introduced in Python 3.8, so using it constrains `safetensors` to that min version. This PR provides an alternative.

### DIFF
--- a/bindings/python/py_src/safetensors/torch.py
+++ b/bindings/python/py_src/safetensors/torch.py
@@ -84,7 +84,7 @@ def _tobytes(tensor: torch.Tensor) -> bytes:
     import ctypes
     import numpy as np
 
-    length = math.prod(tensor.shape)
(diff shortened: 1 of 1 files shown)


bindings/python/py_src/safetensors/torch.py
```diff
@@ -84,7 +84,7 @@ def _tobytes(tensor: torch.Tensor) -> bytes:
     import ctypes
     import numpy as np
 
-    length = math.prod(tensor.shape)
+    length = np.prod(tensor.shape).item()
```
Narsil, line 87: Ok, I had removed it because PT doesn't technically require numpy (I still had to keep it do get the underlying buffer without copying)

Narsil (approved):

PR #17, opened 2022-10-04
Refusing to save non contiguous tensors now.

Narsil:
Refusing to save non contiguous tensors now.

It causes weird issues.

1/ It is hard to make it play well with lazy-loading since you would
have 2 call sites, 1 needing to refer to some mother tensor (and
potentially a whole graph of them) that might not yet exist.

```python
x = torch.zeros((20, 20))
y = x[:, 0]

torch.save([x, y], filename)

x1, y1 = torch.load(filename)
y1.fill_(1)
```

This exists because some two different tensors can refer to the same
buffer with different striding, and pytorch save can persist them.

This library and format doesn't enable that so we should just
discard them.

2/ There was a bug when saving from something else than `cpu`
because the pointer would point to a GPU address for instance
so every tensor needs to be brought back before saving
(Unless there's a better way).

3/ Disable saving sparse tensors.

Forcing contiguous tensors from PT.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,17 +33,19 @@ repos:
             "--",
             "-Dwarnings",
           ]
-  - repo: https://github.com/ambv/black
-    rev: 21.4b2
+  - repo: https://github.com/psf/black
+    rev: 22.3.0
     hooks:
       - id: black
         name: "Python (black)"
-        args: ["--line-length", "100", "--target-version", "py35"]
+        args: ["--line-length", "119", "--target-version", "py35"]
         types: ["python"]
-  - repo: https://github.com/ambv/black
-    rev: 21.4b2
+  - repo: https://gitlab.com/pycqa/flake8
+    rev: 3.8.3
     hooks:
-      - id: black
-        name: "Python Pyi (black)"
-        args: ["--line-length", "100", "--target-version", "py35"]
-        types: ["pyi"]
+      - id: flake8
+        args: ["--config", "bindings/python/setup.cfg"]
+  - repo: https://github.com/pre-commit/mirrors-isort
+    rev: v5.7.0 # Use the revision sha / tag you want to point at
+    hooks:
+      - id: isort
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,49 @@
+repos:
+  - repo: https://github.com/Narsil/pre-commit-rust
+    rev: 0c016cee78144d06d906fccc7715d607a946ca5c
+    hooks:
+      - id: fmt
+        name: "Rust (fmt)"
+        args: ["--manifest-path", "safetensors/Cargo.toml", "--"]
+      - id: clippy
+        name: "Rust (clippy)"
+        args:
+          [
+            "--manifest-path",
+            "safetensors/Cargo.toml",
+            "--all-features",
+            "--all-targets",
+            "--",
+            "-Dwarnings",
+          ]
+  - repo: https://github.com/Narsil/pre-commit-rust
+    rev: 0c016cee78144d06d906fccc7715d607a946ca5c
+    hooks:
+      - id: fmt
+        name: "Python (fmt)"
+        args: ["--manifest-path", "bindings/python/Cargo.toml", "--"]
+      - id: clippy
+        name: "Python (clippy)"
+        args:
+          [
+            "--manifest-path",
+            "bindings/python/Cargo.toml",
+            "--all-features",
+            "--all-targets",
+            "--",
+            "-Dwarnings",
+          ]
+  - repo: https://github.com/ambv/black
+    rev: 21.4b2
+    hooks:
+      - id: black
+        name: "Python (black)"
+        args: ["--line-length", "100", "--target-version", "py35"]
+        types: ["python"]
+  - repo: https://github.com/ambv/black
+    rev: 21.4b2
+    hooks:
+      - id: black
+        name: "Python Pyi (black)"
+        args: ["--line-length", "100", "--target-version", "py35"]
+        types: ["pyi"]
--- a/bindings/python/py_src/safetensors/torch.py
+++ b/bindings/python/py_src/safetensors/torch.py
@@ -1,8 +1,8 @@
-from .safetensors_rust import serialize_file, serialize, safe_open, deserialize
-import math
 from typing import Dict, Optional
 
 import torch
+
+from .safetensors_rust import deserialize, safe_open, serialize, serialize_file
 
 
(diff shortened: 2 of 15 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,49 @@
...
+          [
+            "--manifest-path",
+            "bindings/python/Cargo.toml",
+            "--all-features",
+            "--all-targets",
+            "--",
+            "-Dwarnings",
+          ]
+  - repo: https://github.com/ambv/black
+    rev: 21.4b2
```
sgugger, line 37: Let's use 22.3 here to be consistent with the other OSS libraries of HuggingFace (maybe Tokenizer needs an update if those pre-commit hooks come from there)

.pre-commit-config.yaml
```diff
@@ -0,0 +1,49 @@
...
+        args:
+          [
+            "--manifest-path",
+            "bindings/python/Cargo.toml",
+            "--all-features",
+            "--all-targets",
+            "--",
+            "-Dwarnings",
+          ]
+  - repo: https://github.com/ambv/black
```
sgugger, line 36: There is a redirect, but the repo is in the psf org
```suggestion
  - repo: https://github.com/psf/black
```

.pre-commit-config.yaml
```diff
@@ -0,0 +1,49 @@
...
+    rev: 21.4b2
+    hooks:
+      - id: black
+        name: "Python (black)"
+        args: ["--line-length", "100", "--target-version", "py35"]
+        types: ["python"]
+  - repo: https://github.com/ambv/black
+    rev: 21.4b2
```
sgugger, line 44: Same comments here.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,49 @@
...
+            "--all-targets",
+            "--",
+            "-Dwarnings",
+          ]
+  - repo: https://github.com/ambv/black
+    rev: 21.4b2
+    hooks:
+      - id: black
+        name: "Python (black)"
+        args: ["--line-length", "100", "--target-version", "py35"]
```
sgugger, line 41: Could we use 119 as a line length (same as Transformers, Datasets, Accelerate, Diffusers)?

bindings/python/py_src/safetensors/torch.py
```diff
@@ -7,7 +7,11 @@
 
 def save(tensors: Dict[str, torch.Tensor], metadata: Optional[Dict[str, str]] = None) -> bytes:
     flattened = {
-        k: {"dtype": str(v.dtype).split(".")[-1], "shape": v.shape, "data": _tobytes(v)}
+        k: {
+            "dtype": str(v.dtype).split(".")[-1],
+            "shape": v.shape,
+            "data": _tobytes(v, k),
+        }
```
sgugger, line 14: Will need to be reverted if we do use 119 as a max length.

bindings/python/py_src/safetensors/torch.py
```diff
@@ -21,7 +25,11 @@ def save_file(
     metadata: Optional[Dict[str, str]] = None,
 ):
     flattened = {
-        k: {"dtype": str(v.dtype).split(".")[-1], "shape": v.shape, "data": _tobytes(v)}
+        k: {
```
sgugger, line 28: Same here.

bindings/python/py_src/safetensors/torch.py
```diff
@@ -80,7 +88,26 @@ def _view2torch(safeview) -> Dict[str, torch.Tensor]:
...
-def _tobytes(tensor: torch.Tensor) -> bytes:
+def _tobytes(tensor: torch.Tensor, name: str) -> bytes:
+
+    try:
+        if not tensor.is_contiguous():
+            raise ValueError(
+                f"""You are trying to save a non contiguous tensor: `{name}` which is not allowed. It either means
+            you are trying to save tensors which are reference of each other in which case it's recommended to save
+            only the full tensors, and reslice at load time, or simply call `.contiguous()` on your tensor to pack it
+            before saving."""
```
sgugger, line 99:
```suggestion
                f"You are trying to save a non contiguous tensor: `{name}` which is not allowed. It either means "
                "you are trying to save tensors which are reference of each other in which case it's recommended to save "
                "only the full tensors, and reslice at load time, or simply call `.contiguous()` on your tensor to pack it "
                "before saving."
```
If using triple quotes here, the new lines and indentation will show with the warning. It's better to use multiple lines like the suggestion.

bindings/python/py_src/safetensors/torch.py
```diff
@@ -80,7 +88,26 @@ def _view2torch(safeview) -> Dict[str, torch.Tensor]:
...
+            )
+    except RuntimeError:
+        # This occurs with sparse tensors
+        raise ValueError(
+            f"""You are trying to save a sparse tensor: `{name}` which this library does not support. You can make it a dense
```
sgugger, line 104: Same comment.

bindings/python/tests/test_pt_comparison.py
```diff
@@ -24,6 +24,22 @@ def test_odd_dtype(self):
         self.assertTrue(torch.equal(data["test2"], reloaded["test2"]))
         self.assertTrue(torch.equal(data["test3"], reloaded["test3"]))
 
+    @unittest.skipIf(not torch.cuda.is_available(), "Cuda is not available")
+    def test_gpu(self):
+        data = {
+            "test": torch.arange(4).view((2, 2)).to("cuda:0"),
+        }
```
sgugger, line 31: Should fit in one line.

bindings/python/tests/test_pt_comparison.py
```diff
@@ -73,6 +89,16 @@ def setUp(self):
         # Need to copy since that call mutates the tensors to numpy
         save_file(self.data.copy(), self.local)
 
+    def test_cannot_serialize_a_non_contiguous_tensor(self):
+        tensor = torch.arange(6, dtype=torch.float32).reshape((1, 2, 3))
+        x = tensor[:, :, 1]
+        data = {"test": x}
+        self.assertFalse(
+            x.is_contiguous(),
+        )
```
sgugger, line 98: Same here.

sgugger (approved):
Thanks a lot for the PR! I left comments on the style mainly.

I think it's important to align the version of black and its options so that people can use the same env to contribute to the whole Hugging Face ecosystem.

.pre-commit-config.yaml
Narsil: Yep it needs an update.

I also need to create the `Makefile` (it's just that `make` doesn't work on Windows)

Narsil:
> I think it's important to align the version of black and its options so that people can use the same env to contribute to the whole Hugging Face ecosystem.

100% agree, `tokenizers` was lagging behind, I'll update it too.

Issue #15, opened 2022-09-30
Question about parallelism/rayon


mishig25:
Questions regarding parallelism:

1. If I'm not mistaken, both tensor serialization & deserialization operations should be parallelizeable. Is this assumption correct? For example, I was thinking this line below can be replaced with [par_extend](https://docs.rs/rayon/latest/rayon/iter/trait.ParallelExtend.html#tymethod.par_extend) https://github.com/huggingface/safetensors/blob/86e356fed6b75d4b0a06ee446ab94f1317eafc2e/safetensors/src/tensor.rs#L59 
2. If point 1 is valid, does it make sense to do so? Or is it not worth it (i.e. speed up might not be great OR we don't want to introduce dependency on rayon)?

Thanks so much! 🤗 
cc: @Narsil

Narsil:
Hi @mishig25 ,

Have you tried it ? I'm curious to see the results.

Since `buffer` is mutable I expect this will not be trivially parallelizable. 
It's probably technically doable in this function since all buffer sizes are known at this point.

For writing to disk `serialize_file` I anticipate, the limiting factor would be the actual disk speeds not CPUs, and random disk access is I think worse than serialized one even for SSDs. 

`rayon` is a big dependency and I wouldn't add it without solid reason. One of the biggest merits of this lib is the smallish size and very low dependencies (currently only `serde` and `serde_json`).

PR #19, opened 2022-10-06
Implement benchmarking using Criterion

mishig25:
### TLDR: add benchmarking for `serialization` & `deserialziation`

Not sure if we will do parallelization https://github.com/huggingface/safetensors/issues/15
But either way, I thought it is probably a good idea to add some benchmarking.

I'm using [criterion](https://github.com/bheisler/criterion.rs), which is a popular rust benchmarking framework (also used on [tokenizers](https://github.com/huggingface/tokenizers/blob/0f20b0823a41de7aa918e90e58a3398b024f4b73/tokenizers/Cargo.toml#L76)).

Can be run as:
```
cargo bench
```
Produces output as:
```
Serlialize 10_MB
time:   [427.15 µs 436.21 µs 446.56 µs]                             
change: [-0.3956% +0.7567% +1.9834%] (p = 0.32 > 0.05)
No change in performance detected.

Deserlialize 10_MB                     
time:   [5.5303 µs 5.5634 µs 5.5997 µs]                                
change: [-0.4513% +0.5147% +1.5265%] (p = 0.32 > 0.05)
No change in performance detected.
```
Read more about [cargo bench here](https://doc.rust-lang.org/cargo/commands/cargo-bench.html).
Note: since criterion is a dev-dependency, it does not get added in the prod build

### DIFF
--- a/safetensors/Cargo.toml
+++ b/safetensors/Cargo.toml
@@ -8,3 +8,10 @@ edition = "2021"
 [dependencies]
 serde = {version = "1.0", features = ["derive"]}
 serde_json = "1.0"
+
+[dev-dependencies]
+criterion = "0.3"
+
+[[bench]]
+name = "benchmark"
+harness = false
--- a/safetensors/benches/benchmark.rs
+++ b/safetensors/benches/benchmark.rs
@@ -1,0 +1,55 @@
+use criterion::{black_box, criterion_group, criterion_main, Criterion};
+use safetensors::tensor::*;
+use std::collections::HashMap;
+
+// Returns a sample data of size 2_MB
+fn get_sample_data() -> (Vec<u8>, Vec<usize>, Dtype) {
+    let shape = vec![1000, 500];
+    let dtype = Dtype::F32;
+    let n: usize = shape.iter().product::<usize>() * dtype.size(); // 4
+    let data = vec![0; n];
+
+    (data, shape, dtype)
+}
+
+pub fn bench_serialize(c: &mut Criterion) {
+    let (data, shape, dtype) = get_sample_data();
+    let n_layers = 5;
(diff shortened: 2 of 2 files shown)


sgugger (approved):
For what my Rust beginner's opinion is worth, LGTM!
Thanks for adding benchmarks!

Narsil (approved):
LGTM.

Very nice !
Just FYI, there are some "speed" tests directly in Python.
For this sort of job, the disk access is likely to be an extremely important part of the overall speed of the program so it's important to keep that in mind.

Btw, if you want to check an idea for speeding things up, I think it could be in the load order of tensors. (Only PT is really faster afaik).

If we could order `safe_open("..").keys()` to be in increasing offsets, it's likely that it's going to be faster to load the overall array.
I haven't really tested in, but the speeds are so flaky right now that I'm guessing part of it is the randomization linked to the hashmap.

mishig25:
> here are some "speed" tests directly in Python.
@Narsil could you elaborate on this? 🤗 

> For this sort of job, the disk access is likely to be an extremely important part of the overall speed of the program so it's important to keep that in mind.

Is there anything I should do in this PR or is it good to merge?

Narsil:
`pytest -sv tests/`  within `bindings/python` and you can see some performance measurements. (Fully serializing a gpt2 model to disk and back).

Narsil:
> Is there anything I should do in this PR or is it good to merge?

Go right ahead ! Thanks !

mishig25:
> pytest -sv tests/ within bindings/python and you can see some performance measurements. (Fully serializing a gpt2 model to disk and back).

oh I see, thanks a lot! I was looking specifically for a test with name `*speed*`

PR #23, opened 2022-10-19
Explaining the `__metadata__` in the spec.

Narsil:


sgugger (approved):
Thanks for documenting it!

PR #25, opened 2022-10-19
Creating a few range of attacks against classic formats.

Narsil:
- pickle (ACE affecting PT)
- zip (Denial of Service affecting numpy + PT)
- marshal (affecting Keras, not directly in the h5 file format)

- Created 8 proposals to abuse the current format.
  Found 2 kind of flaws that could also cause some form of denial of
  service). PRs coming up.

PR #26, opened 2022-10-19
Creating a few range of attacks against classic formats.

Narsil:
- pickle (ACE affecting PT)
- zip (Denial of Service affecting numpy + PT)
- marshal (affecting Keras, not directly in the h5 file format)

- Created 8 proposals to abuse the current format.
  Found 2 kind of flaws that could also cause some form of denial of
  service). PRs coming up.

### DIFF
--- a/attacks/README.md
+++ b/attacks/README.md
@@ -1,0 +1,59 @@
+The purpose of this directory is to showcase various attacks (and creating your own).
+
+
+# Torch Arbitrary code execution
+
+Try it out. This will create a seemingly innocuous `torch_ace.pt` file.
+```
+python torch_ace_create.py
+python torch_ace_get_pwned.py
+```
+
+# Tensorflow (Keras) Arbitrary Code execution (does not affect `transformers`)
+
+Try it out. This will create a seemingly innocuous `tf_ace.h5` file.
+```
+python tf_dos_create.py
+python tf_dos_get_pwned.py
+```
+
+# Torch Denial of Service (OOM kills the running process)
+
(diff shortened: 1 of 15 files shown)


attacks/README.md
```diff
@@ -0,0 +1,59 @@
+The purpose of this directory is to showcase various attacks (and creating your own).
```
McPatate, line 1: Are you planning to rebase  after merging #25 ?

safetensors/src/tensor.rs
```diff
@@ -91,19 +97,36 @@ pub struct SafeTensors<'data> {
...
+        buffer: &'in_data [u8],
+    ) -> Result<(usize, Metadata), SafeTensorError>
     where
         'in_data: 'data,
     {
         let arr: [u8; 8] = [
             buffer[0], buffer[1], buffer[2], buffer[3], buffer[4], buffer[5], buffer[6], buffer[7],
         ];
-        let n = u64::from_le_bytes(arr) as usize;
+        let n: usize = u64::from_le_bytes(arr)
```
McPatate, line 110: do we still need a `u64` here ?

McPatate (approved):
lgtm !

sgugger (approved):
LGTM for what it's worth!
Super interesting to see the various attack files!

safetensors/src/tensor.rs
Narsil: Yes, `usize` is plateform dependant (mostly `u64` nowadays).

`usize` is the size of a pointer. Which we need for indexing the buffer.

The cases where this would be incorrect seems pretty niche now, but still.

PR #29, opened 2022-10-20
Updating convert script + fixing index.

Narsil:


bindings/python/convert.py
```diff
@@ -31,6 +44,8 @@ def convert_multi(model_id):
         commit_message="Adding `safetensors` variant of this model",
         create_pr=True,
     )
+    for local in local_filenames:
+        os.remove(local)
```
sgugger, line 48: Looks like the kind of things that should go in a `finally` and the rest should be in a `try` block, to avoid leaving the user with files in case of error during the script.

bindings/python/convert.py
```diff
@@ -48,6 +63,7 @@ def convert_single(model_id):
         path_in_repo=local,
         repo_id=model_id,
     )
+    os.remove(local)
```
sgugger, line 66: Same here.

bindings/python/convert.py
```diff
@@ -65,7 +81,10 @@ def convert_single(model_id):
     )
     args = parser.parse_args()
     model_id = args.model_id
-    try:
-        convert_multi(model_id)
-    except EntryNotFoundError:
+    api = HfApi()
+    info = api.model_info(model_id)
+    filenames = set(s.rfilename for s in info.siblings)
+    if "pytorch_model.bin" in filenames:
```
sgugger, line 87: Much cleaner!

bindings/python/convert.py
```diff
@@ -65,7 +81,10 @@ def convert_single(model_id):
...
         convert_single(model_id)
+    else:
+        convert_multi(model_id)
```
sgugger, line 90: This could also return the path to the lfile/list of files so the try/finally is just around those instructions and does not take a whole function.

Narsil: But if the function crashes and has not try except within the body, then there's no way to capture the llist of files right ?

bindings/python/convert.py
Narsil: Updated. I didn't meant to make this the most bullet proof but you're right so I added this.

Narsil:
@sgugger can you approve if you think it's good enough please ? :)

sgugger (approved):
Ah sorry, forgot to click the green button!

PR #31, opened 2022-10-20
Less GIL locks speeds things up apparently ?

Narsil:
This PR reduces the locations where the GIL is taken, which at least in appearance speeds up significantly the load speed for PT.
I am now on my local computer ~3x faster to load on CPU.

The PR stems from realizing that trying to load on GPU directly, was consuming a LOT of CPU (it still does).
GPU loading was much slower to load everything on CPU then moving to GPU than directly loading on GPU (tensor per tensor).

GIL definitely seems at play here.

So `safe_open` now receives a `device` argument to that the various functons can return tensors directly on the correct location.

No check or concern for TF or Flax has been taken yet (numpy cannot allocate on GPU).
When we do we'll probably have to make `Device` uniform across library.

### DIFF
--- a/bindings/python/src/lib.rs
+++ b/bindings/python/src/lib.rs
@@ -156,14 +156,16 @@ struct safe_open {
     metadata: Metadata,
     offset: usize,
     framework: Framework,
+    device: String,
     mmap: Arc<Mmap>,
 }
 
 #[pymethods]
 impl safe_open {
     #[new]
-    fn new(filename: &str, framework: Framework) -> PyResult<Self> {
+    fn new(filename: &str, framework: Framework, device: Option<String>) -> PyResult<Self> {
         let file = File::open(filename)?;
+        let device = device.unwrap_or_else(|| "cpu".to_string());
 
         // SAFETY: Mmap is used to prevent allocating in Rust
         // before making a copy within Python.
@@ -186,6 +188,7 @@ impl safe_open {
             metadata,
             offset,
             framework,
+            device,
             mmap: Arc::new(buffer),
         })
     }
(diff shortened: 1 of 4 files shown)


bindings/python/tests/test_pt_comparison.py
```diff
@@ -46,36 +46,63 @@ def test_sparse(self):
...
         pt_time = datetime.datetime.now() - start
         print()
         print(f"Deserialization (Safe) took {safe_time}")
         print(f"Deserialization (PT) took {pt_time} (Safe is {pt_time/safe_time} faster)")
+        for k, v in weights.items():
+            tv = tweights[k]
+            self.assertTrue(torch.allclose(v, tv))
+            self.assertEqual(v.device, torch.device("cpu"))
+
+    def test_deserialization_safe_gpu(self):
```
sgugger, line 75: This test should have some kind of decorator to skip it if there is no GPU (will be necessary to get the CI green).

sgugger (approved):
Nice add!

bindings/python/src/lib.rs
```diff
@@ -288,25 +306,34 @@ impl PySafeSlice {
...
+            self.info.dtype,
+            &newshape,
+            array,
+            &self.device,
+        )
     }
 }
 
 fn create_tensor(
-    py: Python,
```
mishig25, line 305: what is the advantage/reason of removing `py` arg & using `Python::with_gil(|py| {` ?

Narsil: It is the core of the thing.
I must say I don't completely understand but it comes from here:

https://github.com/PyO3/pyo3/issues/1056
https://github.com/PyO3/pyo3/issues/1547

Basically you want to hold the gil as little as possible to give a chance to pyo3 to release it's objects (which then makes the memory available for follow up tensor creation)

mishig25: oh I see 👍

PR #33, opened 2022-10-24
[WIP] Make GPU loading faster by removing all extra CPU copies.

Narsil:


bindings/python/src/lib.rs
```diff
@@ -211,6 +211,144 @@ struct safe_open {
...
+    unsafe {
+        let cuda_memcpy: libloading::Symbol<
+            unsafe extern "C" fn(
+                device_ptr: u64,
+                src_ptr: *const std::ffi::c_void,
+                src_len: usize,
+            ) -> u32,
+        > = cudart.get(b"cudaMemcpy").map_err(|e| {
+            exceptions::PyException::new_err(format!("Couldn't find cudaMemcpy {e:?}",))
+        })?;
```
co42, line 330: I'm not sure of the cost of that but shouldn't it be done only once ? In `__enter__` maybe

bindings/python/src/lib.rs
```diff
@@ -262,25 +412,47 @@ impl safe_open {
...
-            &self.framework,
-            info.dtype,
-            &info.shape,
-            array,
-            &self.device,
-        )
+        match (&self.device, &self.framework, &self.cudart) {
+            (Device::Cuda(_), Framework::Pytorch, Some(cudart)) => {
+                Python::with_gil(|py| -> PyResult<PyObject> {
+                    let module = PyModule::import(py, intern!(py, "torch"))?;
```
co42, line 418: Do we need to do this import every time we get a tensor ?

bindings/python/src/lib.rs
Narsil: This is a reference, and references with a lifetime and those are disallowed in `pyclass`.
There might be ways those.

bindings/python/src/lib.rs
Narsil: You can't use statics for the same reason is that all Python variables have a python lifetime associated with it.

However, I just discovered : https://docs.rs/pyo3/latest/pyo3/once_cell/struct.GILOnceCell.html !

bindings/python/src/lib.rs
Narsil: FYI I timed the actual libloading, it takes `70us` roughly..  (That's actually fetching the entire library, not just this method)

bindings/python/src/lib.rs
```diff
@@ -211,6 +211,144 @@ struct safe_open {
...
+            let out = cuda_memcpy(
+                (data_ptr + offset) as u64,
+                slice.as_ptr() as *const std::ffi::c_void,
+                len,
+            );
+            // SAFETY: Here we have a correct library, we successfully called memcpy,
+            // but somehow the call failed. This is really worrying since Pytorch is
+            // responsible for allocating the memory.
+            if out != 0 {
+                panic!(
```
davidhewitt, line 342: Should this be `std::process::abort`? It is possible (although discouraged) to catch the `PanicException` translated by PyO3. We don't allow users to name it directly, but it derives from `BaseException` so some `catch` blocks can halt them.

bindings/python/src/lib.rs
```diff
@@ -211,6 +211,144 @@ struct safe_open {
...
+        > = cudart.get(b"cudaMemcpy").map_err(|e| {
+            exceptions::PyException::new_err(format!("Couldn't find cudaMemcpy {e:?}",))
+        })?;
+        for slice in iterator {
+            let len = slice.len();
+            let out = cuda_memcpy(
+                (data_ptr + offset) as u64,
+                slice.as_ptr() as *const std::ffi::c_void,
+                len,
+            );
```
davidhewitt, line 337: Personally I like to reduce the scope of the `unsafe`, so I'd just put it around here and the `cudart.get` call.

```suggestion
            // Safety: pointer and len are guaranteed valid as they originate from the same slice
            let out = unsafe { cuda_memcpy(
                (data_ptr + offset) as u64,
                slice.as_ptr() as *const std::ffi::c_void,
                len,
            ) };
```

davidhewitt (commented):
No problem, happy to offer ideas (although I can't always guarantee a speedy reply!).

FFI is both inherently unsafe and also difficult to verify. For pure-rust `unsafe` there's tools such as miri to check for UB, I think with FFI the best you can do is run tools like Valgrind but I'm unconvinced they'll give much value here.

One suggestion I do have - the way you search for the cuda runtime lib is a bit spooky, I think it might not match the one that torch has already loaded? I had a quick look but didn't see a torch API to get the loaded cuda runtime name (although it might still exist). I wonder if possible to enumerate already loaded libraries to find the cuda lib - you can then use `libloading` to get a new handle and I _think_ that'll also be guaranteed to avoid arbitrary code execution because the lib will already have been initialised?

bindings/python/src/lib.rs
Narsil: Thanks for the tip ! 
`std::process::abort` I think.

bindings/python/src/lib.rs
Narsil: I agree in general. Here I tried to encapsulate both fetching the symbol and the various calls, as basically they all influence each other.

Narsil:
>  the way you search for the cuda runtime lib is a bit spooky

100% agree, I tinkered with "cleaner approaches", but I found it quite hard to code. 
`torch._C` is their c library. 
It is linked to `libtorch_python.so`, which is linked to `libtorch_cuda_cpp.so` which is linked to `libcudart-xxxx.so.xxx`.

Afaik, the "link" is resolved at runtime in the actual process and unknown to the program here. So there is still some manual resolution to be done. Meaning reading environement LD_LIBRARY_PATH and so on. Which I don't feel very confident about.

This current resolution is not great because if someone manages to highjack this location and send a bogus library named `libcudart` then it will indeed enable escape.
But I feel like if an attacker managed to do this, well security is gone already.

bindings/python/src/lib.rs
Narsil: I tried to save directly `cuda_memcpy` within the GILOnceCell, but I wasn't able to do it. Well I was able to compile it, but I triggered  the panic when using `cuda_memcpy` afterwards. So something must have been very wrong.

It was something like:

```rust
fn find_memcpy() -> Symbol<'static, MemcpyFn>{
   let symbol = Python::with_gil(|py| {
         let module = get_module(py, &TORCH_MODULE);
         let lib = find_cudart(module);
         let lib = Box::leak(Box::new(lib));
         let symbol = lib.get("cudaMemcpy");
         symbol
     })
     symbol
}
....
// At init
CUDA_MEMCPY.get_or_init(py, || find_memcpy());
```

However somehow it fails to properly call it.
`Box::leak` prevents the  `Drop` from being run right ? So the lib should not get unloaded ?

For instance:
```
The implementation of thread-local variables is extremely platform specific and uses of such variables that work on e.g. Linux may have unintended behaviour on other targets.
```
So maybe we can't really leak.

`Library` itself doesn't seem to be thread safe, maybe we should load on every access to be safe ? (It's ok speedwise)

davidhewitt:
> But I feel like if an attacker managed to do this, well security is gone already.

I mostly agree with this point; I guess that what this might enable is something like a privilege escalation if attacker used user X to write the malicious lib and then your process with user Y loads it.

Is it perhaps asking upstream in torch if they have or are willing to expose a way to get the currently loaded lib?

bindings/python/src/lib.rs
davidhewitt: Hmm, that's unfortunate. It's not clear to me what might be going wrong, that looks like it should be fine to me.

I'm not aware of any thread-local variable here?

If torch is dynamically loading the cuda runtime, I wondered if it's possible that initialising too early here might interact badly with torch? I'm just speculating wildly at this point.

davidhewitt: If you're not worried about the loading call, you could always cache what you wanted from torch / filesystem (e.g. cuda location) and then the per-call cost is just load & lookup the symbol.

Narsil:
> Is it perhaps asking upstream in torch if they have or are willing to expose a way to get the currently loaded lib?

I'll try to do that. Old torch libs still won't work though.
Good point about privilege escalation, didn't think about that.

bindings/python/src/lib.rs
Narsil: Yes, this is what I'll end up doing I think.

Narsil: Also in `torch==1.13` they splitted cuda to it's own lib, so ....

Narsil: Ok, figure it out, loading `torch._C` directly actually works. `libloading` uses `dlopen` under the hood and is able to resolve `cudaMemcpy` directly, so no need for ELF shenanigans and it's safe since it's already loaded in the torch python namespace !

Narsil:
@davidhewitt 

I am finding super weird issues (and slightly worrying).

```rust
  let data_ptr_fn = tensor.getattr("data_ptr")?;
    let data_ptr: usize = data_ptr_fn.call0()?.extract()?;

    let out = unsafe {
        cuda_memcpy(
            data_ptr.try_into()?,
```

This works, but this doesn't:

```rust
  let data_ptr_fn = tensor.getattr("data_ptr")?;
    let data_ptr: u64 = data_ptr_fn.call0()?.extract()?;

    let out = unsafe {
        cuda_memcpy(
            data_ptr,
```
It fails with some weird cuda error. Either invalid argument, or memcpy invalid direction (??).
This doesn't make a lot of sense to me.
The values look exactly when I compare them everything really seems to be equivalent.

Is there anything you might be aware of ? 

Also referencing the `Symbol<_>` seems to be unsafe (just moving the calls from `cuda_memcpy` to `´&cuda_memcpy` in the loop).

davidhewitt:
Where did you get the definition of the `cudaMemcpy` symbol?

Looking at [the docs on Nvidia](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1gc263dbe6574220cc776b45438fc351e8) (which might be wrong, I'm not an expert here), it seems like the definition you want is:

```rust
#[repr(C)]
enum cudaMemcpyKind {
    cudaMemcpyHostToHost = 0,
    cudaMemcpyHostToDevice = 1,
    cudaMemcpyDeviceToHost = 2,
    cudaMemcpyDefault = 3
}

type MemcpyFn =
     unsafe extern "C" fn(dest: *mut c_void, src_ptr: *const c_void, src_len: libc::size_t, kind: cudaMemcpyKind) -> u32;
```

In particular you should be using pointer for the first argument in case you are running on a 32-bit system, and additionally you are missing the `kind` argument. Without this you've got UB as the FFI function call will be setting up an incorrect stack frame for the C function arguments. The `kind` parameter is unset and will likely generate exactly the kind of errors you are reporting.

Looks like with `usize` input you're just getting luck that the stack is set up correctly, with `u64` not so lucky.

davidhewitt:
You might want to consider creating a `#[repr(C)] enum` for the return value too, as I'm unsure if the size is guaranteed to be `u32` on all platforms. https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1g3f51e3575c2178246db0a94a430e0038

Narsil:
Thanks ! Super helpful.

For some reason I was using https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MEM.html#group__CUDA__MEM_1g4d32266788c440b0220b1a9ba5795169 and using its signature instead of `cudaMemcpy`.

That does explain the corrupted stack ! And the issues when I tried to leak the reference. The corruption was not because of the symbol but because the stack changed and the signature was wrong.

PR #34, opened 2022-10-25
Actually pass along metadata

sgugger:
When adding the metadata, I wasn't very careful in the integrations I did not test :sweat_smile: 

This PR fixes it.


Narsil (approved):
Oops :) 

Thanks for the fix.

PR #35, opened 2022-10-26
Some PyO3 specifics.

Narsil:
Move `Device` to real struct (makes management of it much easier).

Use `intern!` as per : https://pyo3.rs/main/doc/pyo3/macro.intern.html

### DIFF
--- a/bindings/python/src/lib.rs
+++ b/bindings/python/src/lib.rs
@@ -226,18 +226,36 @@ impl safe_open {
 
                     let data_ptr_fn = tensor.getattr("data_ptr")?;
                     let data_ptr: usize = data_ptr_fn.call0()?.extract()?;
-                    use rustacuda::memory::{CopyDestination, DeviceBuffer, DevicePointer};
                     unsafe {
-                        let data_ptr = DevicePointer::wrap(&mut *(data_ptr as *mut u8));
+                        #[derive(Debug)]
+                        pub struct CudaError(String);
 
-                        let mut buffer = DeviceBuffer::from_raw_parts(data_ptr, data.len());
-                        buffer.copy_from(data).map_err(|e| {
-                            exceptions::PyException::new_err(format!(
-                                "Error when Copying to cuda {:?}",
-                                e
-                            ))
-                        })?;
-                        std::mem::forget(buffer);
+                        pub fn check_cuda(err: cuda_sys::CUresult) -> Result<(), CudaError> {
+                            if err == cuda_sys::cudaError_enum_CUDA_SUCCESS {
+                                Ok(())
+                            } else {
+                                unsafe {
+                                    let mut str_ptr = std::ptr::null();
+
+                                    cuda_sys::cuGetErrorString(err, &mut str_ptr);
+                                    let string = std::ffi::CStr::from_ptr(str_ptr)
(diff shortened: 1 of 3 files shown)


bindings/python/src/lib.rs
```diff
@@ -154,26 +154,68 @@ impl<'source> FromPyObject<'source> for Framework {
...
+                if tokens.len() == 2 {
+                    let device: usize = tokens[1].parse()?;
+                    Ok(Device::Cuda(device))
+                } else {
+                    Err(exceptions::PyException::new_err(format!(
+                        "device {name} is invalid"
+                    )))
+                }
+            }
+            name => Err(exceptions::PyException::new_err(format!(
```
sgugger, line 182: Can we add one last match where name can be converted to a usize? Since we can pass int/str for a device usually in PyTorch.

bindings/python/tests/test_pt_comparison.py
```diff
@@ -53,38 +53,77 @@ def setUp(self):
             save_file(data.copy(), self.local)
 
     def test_deserialization_safe(self):
-        start = datetime.datetime.now()
+        W = 1
+        N = 1
```
sgugger, line 57: Still not a fan of one-letter variable names ;-)

bindings/python/tests/test_pt_comparison.py
```diff
@@ -53,38 +53,77 @@ def setUp(self):
...
+        for k, v in weights.items():
+            tv = tweights[k]
+            self.assertTrue(torch.allclose(v, tv))
+            self.assertEqual(v.device, torch.device("cpu"))
+
+    @unittest.skipIf(not torch.cuda.is_available(), "Cuda is not available")
+    def test_deserialization_safe_gpu(self):
         # First time to hit disk
-        load_file(self.local)
+        load_file(self.local, device="cuda:0")
```
sgugger, line 90: We could test here that passing 0 works as well as `"cuda:0"`.

bindings/python/tests/test_pt_comparison.py
```diff
@@ -94,6 +133,29 @@ def test_deserialization_safe_gpu(self):
             self.assertTrue(torch.allclose(v, tv))
             self.assertEqual(v.device, torch.device("cuda:0"))
 
+    @unittest.skipIf(torch.cuda.device_count() < 2, "Only 1 device available")
+    def test_deserialization_safe_device_1(self):
+        # First time to hit disk
+        load_file(self.local, device="cuda:1")
```
sgugger, line 139: Same here for 1

sgugger (commented):
Nice additions, thanks! To be more intuitive for PyTorch users, it would be great if the device argument could also accept an int.

Also in a follow up we might need to add support for XLA devices (like there is for MPS).

bindings/python/src/lib.rs
Narsil: Sure ! Good idea.
I already tested "mps", it works but it's slower to create tensors 1 by 1 on "mps" than creating them on CPU and then moving them to "mps".. .Don't really know why. (`torch.load(.., map_location="mps")` doesn't work too afaik)

sgugger (approved):

PR #37, opened 2022-10-26
Moving tensor validation (length wise) directly in validate.

Narsil:
That way we have a single point of validation.

Either when you create `Metadata` when you are serializing. Or when we deserialize the JSON we have to use `validate()` and will validate all edge cases.


sgugger (approved):
LGTM!

PR #39, opened 2022-10-26
Support for Python 3.7

sgugger:
`importlib.metadata` only exists in Python 3.8 and higher, this PR fixes the package for Python 3.7 (note that a new release will be necessary).


sgugger:
Yes sorry it was added automatically by VSCode in my previous PR 🤦

PR #42, opened 2022-10-28
Use BTreeMap instead of HashMap

thomasw21:
@julien-c reported that the `sha256` from the same model would generate different values when running the saving script twice. My guess is that this is due to the use of `HashMap` which doesn't guarantee ordering when generating the json string of it. In order to fix this @OlivierDehaene suggested to use `serde::collections::BTreeMap` instead. We also add a test that didn't pass on main.

### DIFF
--- a/bindings/python/py_src/safetensors/torch.py
+++ b/bindings/python/py_src/safetensors/torch.py
@@ -84,7 +84,6 @@ def _view2torch(safeview) -> Dict[str, torch.Tensor]:
 
 
 def _tobytes(tensor: torch.Tensor, name: str) -> bytes:
-
     try:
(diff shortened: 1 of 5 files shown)


thomasw21:
I'm not sure why but the test pass for me locally. It just seems no to be raising at this step: https://github.com/huggingface/safetensors/blob/bb05b526273c6665a7f08b5c0f85a6e977c71e8d/bindings/python/py_src/safetensors/torch.py#L96-L102

This doesn't really seem related to this PR does it?

Narsil:
New `torch` version `1.13` it seems.

Narsil (approved):
LGTM

PR #45, opened 2022-10-29
Supporting torch==1.13.0

Narsil:


bindings/python/py_src/safetensors/torch.py
```diff
@@ -84,22 +84,20 @@ def _view2torch(safeview) -> Dict[str, torch.Tensor]:
...
-        if not tensor.is_contiguous():
-            raise ValueError(
-                f"You are trying to save a non contiguous tensor: `{name}` which is not allowed. It either means you"
-                " are trying to save tensors which are reference of each other in which case it's recommended to save"
-                " only the full tensors, and reslice at load time, or simply call `.contiguous()` on your tensor to"
-                " pack it before saving."
-            )
-    except RuntimeError:
-        # This occurs with sparse tensors
+    if tensor.layout != torch.strided:
```
thomasw21, line 87: Nit: [The doc](https://pytorch.org/docs/stable/tensor_attributes.html#torch-layout) says that currently two are supported, so it's natural to say if it's not dense, it's sparse. To be future proof, why no make it so that `tensor.layout == torch.sparse_coo` then raise ValueError due to sparsity, if it's dense then do the rest of the path, and if it's something else, then raise NotImplemented?

thomasw21 (approved):
A small nit, but this looks good to me. Tested on my mac M1 with torch 1.12.1 and it worked fine.

PR #46, opened 2022-10-31
Testing torch==1.10 (lowest available on transformers with `frombuffer`) too.

Narsil:


thomasw21:
Is it normal that the CI doesn't run `Install Python` ?

Narsil:
CI doesn't run when source files are not modified (actually the rust test themselves shouldn't be run either on PRs that don't modify source code.

thomasw21 (approved):
LGTM! I don't see anywhere any requirements file (except the dev one). Should we enforce minimum torch version somewhere in the codebase? Otherwhise it'll crash with the same error. Maybe in `torch.py`?

Narsil:
Hmm neither `torch` nor `tensorflow` nor `jax` should be required (like `transformers`) therefore, no `requirements.txt`.

I could definitely add an error when using a too old version of `torch`.

thomasw21:
>I could definitely add an error when using a too old version of torch.

Yeah that's what I meant actually. Not mandatory, but it's easier to understand if the error tell you "You're using torch 1.9. We only support >10" than "Attribute `from_buffer` doesn't exist in `torch`"

PR #47, opened 2022-10-31
Add convert all

Narsil:
Hardened script.

- Check actual inference against `transformers`
- Uses directory to prevent clash on space.

### DIFF
--- a/bindings/python/convert.py
+++ b/bindings/python/convert.py
@@ -2,17 +2,31 @@ import argparse
 import json
 import os
 import shutil
+from tempfile import TemporaryDirectory
+from collections import defaultdict
+from inspect import signature
+from typing import Optional, List
 
 import torch
 
-from huggingface_hub import CommitOperationAdd, HfApi, hf_hub_download
+from huggingface_hub import CommitOperationAdd, HfApi, hf_hub_download, get_repo_discussions
 from huggingface_hub.file_download import repo_folder_name
 from transformers import AutoConfig
 from transformers.pipelines.base import infer_framework_load_model
 from safetensors.torch import save_file
 
 
-def check_file_size(sf_filename, pt_filename):
+def shared_pointers(tensors):
+    ptrs = defaultdict(list)
+    for k, v in tensors.items():
+        ptrs[v.data_ptr()].append(k)
+    failing = []
+    for ptr, names in ptrs.items():
+        if len(names) > 1:
+            failing.append(names)
+    return failing
+
+def check_file_size(sf_filename: str, pt_filename: str):
     sf_size = os.stat(sf_filename).st_size
     pt_size = os.stat(pt_filename).st_size
 
@@ -23,13 +37,13 @@ def check_file_size(sf_filename, pt_filename):
          """)
 
 
-def rename(pt_filename) -> str:
+def rename(pt_filename: str) -> str:
     local = pt_filename.replace(".bin", ".safetensors")
     local = local.replace("pytorch_model", "model")
     return local
 
 
-def convert_multi(model_id):
+def convert_multi(model_id: str) -> List["CommitOperationAdd"]:
     filename = hf_hub_download(repo_id=model_id, filename="pytorch_model.bin.index.json")
     with open(filename, "r") as f:
         data = json.load(f)
@@ -58,12 +72,20 @@ def convert_multi(model_id):
     return operations
 
 
-def convert_single(model_id, folder):
+def convert_single(model_id: str, folder: str) -> List["CommitOperationAdd"]:
     sf_filename = "model.safetensors"
     filename = hf_hub_download(repo_id=model_id, filename="pytorch_model.bin")
     loaded = torch.load(filename)
 
     local = os.path.join(folder, sf_filename)
+    shared = shared_pointers(loaded)
+    for shared_weights in shared:
+        for name in shared_weights[1:]:
+            loaded.pop(name)
+
+    # For tensors to be contiguous
+    loaded = {k: v.contiguous() for k, v in loaded.items()}
+
     save_file(loaded, local, metadata={"format": "pt"})
 
     check_file_size(local, filename)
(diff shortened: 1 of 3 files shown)


bindings/python/convert.py
```diff
@@ -14,59 +31,95 @@ def rename(pt_filename) -> str:
...
+def convert(api, model_id):
+    info = api.model_info(model_id)
+    filenames = set(s.rfilename for s in info.siblings)
+
+    folder = repo_folder_name(repo_id=model_id, repo_type="models")
+    os.makedirs(folder)
+    new_pr = None
+    try:
+        operations = None
+        if "model.safetensors" in filenames or "model_index.safetensors.index.json" in filenames:
```
julien-c, line 103: maybe at some point we'll want a `--force` flag (to correct already-merged incorrect safetensors weights for instance)

bindings/python/convert.py
```diff
@@ -14,59 +31,95 @@ def rename(pt_filename) -> str:
...
+    _, pt_model = infer_framework_load_model(model_id, config)
+
+    input_ids = torch.arange(10).long().unsqueeze(0)
+    sf_logits = sf_model(input_ids)
+    pt_logits = pt_model(input_ids)
+    torch.testing.assert_close(sf_logits, pt_logits)
+    print(f"Model {model_id} is ok !")
+
+
+def convert(api, model_id):
```
julien-c, line 94: i guess you're not a fan of types but I feel like here at least the return type `CommitInfo` could be nice :)

Up to you though

bindings/python/convert.py
```diff
@@ -14,59 +31,95 @@ def rename(pt_filename) -> str:
...
+    info = api.model_info(model_id)
+    filenames = set(s.rfilename for s in info.siblings)
+
+    folder = repo_folder_name(repo_id=model_id, repo_type="models")
+    os.makedirs(folder)
```
julien-c, line 99: this line means concurrency *on one single model* is still = 1 (the mkdir call will fail if the previous one hasn't been removed), but i guess that's almost by design

Another option is to just mkdir a temporary dir

bindings/python/convert.py
Narsil: Good suggestion !

bindings/python/convert.py
Narsil: Haha I actually love types, I just hate when it makes code unreadable, or when they are not properly enforced (meaning there's no guarantee coming with them, so they might as well not exist)

bindings/python/convert.py
Narsil: True !

Narsil: Modified to use tempdir. Might also help if `/tmp` is mapped to ramfs.

bindings/python/convert.py
```diff
@@ -1,72 +1,189 @@
...
+    sf_params = sf_model.state_dict()
+
+    pt_shared = shared_pointers(pt_params)
+    sf_shared = shared_pointers(sf_params)
+    if pt_shared != sf_shared:
+        raise RuntimeError("The reconstructed model is wrong, shared tensors are different {shared_pt} != {shared_tf}")
+
+    sig = signature(pt_model.forward)
+    input_ids = torch.arange(10).unsqueeze(0)
+    pixel_values = torch.randn(1, 3, 224, 224)
```
ydshieh, line 117: Might be better to use the values from the config file. (`image_size`), and cast to `float`.
That's why I see in some docstring, as well as
https://github.com/huggingface/transformers/blob/b047472650cba259621549ac27b18fd2066ce18e/tests/test_modeling_common.py#L1747-L1748

bindings/python/convert.py
```diff
@@ -1,72 +1,189 @@
...
+        kwargs["image"] = pixel_values
+
+
+    if torch.cuda.is_available():
+        pt_model = pt_model.cuda()
+        sf_model = sf_model.cuda()
+        kwargs = {k: v.cuda() for k, v in kwargs.items()}
+
+    pt_logits = pt_model(**kwargs)[0]
+    sf_logits = sf_model(**kwargs)[0]
```
ydshieh, line 144: I would suggest to test against all hidden states with `output_hidden_states=True` (whenever possible)

like
https://github.com/huggingface/transformers/blob/b047472650cba259621549ac27b18fd2066ce18e/src/transformers/commands/pt_to_tf.py#L292

bindings/python/convert.py
```diff
@@ -1,72 +1,197 @@
...
+def convert_single(model_id: str, folder: str) -> List["CommitOperationAdd"]:
+    sf_filename = "model.safetensors"
+    filename = hf_hub_download(repo_id=model_id, filename="pytorch_model.bin")
+    loaded = torch.load(filename)
+
+    local = os.path.join(folder, sf_filename)
+    shared = shared_pointers(loaded)
+    for shared_weights in shared:
+        for name in shared_weights[1:]:
+            loaded.pop(name)
```
sgugger, line 88: I don't think it's a good idea to delete those from the state dict. It will definitely surprise users using `safetensors` independently from Transformers, and even in Transformers this will lead to new warnings for users and possible failures as I think a good half of the models don't have tied weights properly implemented.

Narsil: This is checked against during the reload, so the PR won't be opened if that's the case.

I found FSMT in the top 100 that's impacted.

Duplicating the weights in disk WILL crash fine-tuning anyway no ?
So it would be better to implement tying weights, no ? (And crash early when it isn't .)

sgugger: I don't think you are checking against new warnings.

In both cases I don't see where there is going to be crash. In one instance (weights duplicated) inference will work and fine-tuning will have disconnected weights, in the other (weights not duplicated) inference will not work (since the weights won't be found) and fine-tuning will have disconnected weights.

In any case I would trust the PyTorch state dict as the absolute source of trust.

PR #51, opened 2022-11-01
Adding M1 (mps) benchmarks (going first through CPU seems faster though? but PT is just as slow going directly on MPS.).

Narsil:


McPatate (approved):
Do you need me to test it works ok ?

PR #56, opened 2022-11-02
Found out about `torch.asarray` + `torch.Storage` combo.

Narsil:
Extremely fast (I'm guessing it's just pointer arithmetic and actual
load time would it when actually using the tensors.)

```
benches/test_pt.py::test_pt_pt_load_cpu PASSED
benches/test_pt.py::test_pt_sf_load_cpu PASSED


------------------------------------------------------------------------------------- benchmark: 2 tests ------------------------------------------------------------------------------------
Name (time in ms)            Min                 Max                Mean            StdDev              Median                IQR            Outliers       OPS            Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_pt_sf_load_cpu       4.4699 (1.0)        9.8395 (1.0)        5.5089 (1.0)      0.9217 (1.0)        5.2233 (1.0)       0.7899 (1.0)         24;13  181.5238 (1.0)         175           1
test_pt_pt_load_cpu     162.7417 (36.41)    183.1321 (18.61)    173.9028 (31.57)    8.4244 (9.14)     174.7930 (33.46)    13.4904 (17.08)         2;0    5.7503 (0.03)          6           1
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

Legend:
  Outliers: 1 Standard Deviation from Mean; 1.5 IQR (InterQuartile Range) from 1st Quartile and 3rd Quartile.
  OPS: Operations Per Second, computed as 1 / Mean
=============================================================================================================== 2 passed, 4 deselected in 8.96s ================================================================================================================
```

I checked against transformers and I get:

```
============== DUMMY ====================
Loaded in 0:00:03.180369
Setting `pad_token_id` to `eos_token_id`:50256 for open-end generation.
[{'generated_text': 'test.\n\nThe first thing to note is that the "new" version of the app is not'}]
Ran in 0:00:04.339221
============== NO ALLOC ====================
Lookup in 0:00:00.000292
model from config in 0:00:00.581647
tokenizer in 0:00:00.641293
Loaded in 0:00:00.648781
Loaded weights in 0:00:00.654212
Loaded on model in 0:00:00.658909
Setting `pad_token_id` to `eos_token_id`:50256 for open-end generation.
Ran in 0:00:01.828281
[{'generated_text': 'test.\n\nThe first thing to note is that the "new" version of the app is not'}]
Setting `pad_token_id` to `eos_token_id`:50256 for open-end generation.
Ran in 0:00:02.982728
[{'generated_text': 'test.\n\nThe first thing to note is that the "new" version of the app is not'}]
```
And without `safetensors`

```
============== DUMMY (PT) ====================
Loaded in 0:00:04.272222
Setting `pad_token_id` to `eos_token_id`:50256 for open-end generation.
[{'generated_text': 'test.\n\nThe first thing to note is that the "new" version of the app is not'}]
Ran in 0:00:05.432200
```

https://gist.github.com/Narsil/57ddec904ee795cc6e78eb13562f0c36

I am not sure where the 500-600ms come from when loading the model with `no_init_weights` from accelerate, I'm guessing it's `torch` internals but I'm not sure.

the `NO_ALLOC` case is more of a showcase but maybe at some point we can manage to integrate into `transformers`.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -12,7 +12,7 @@ jobs:
         os: [ubuntu-18.04, macos-latest, windows-latest]
         # Lowest and highest, no version specified so that 
         # new releases get automatically tested against
-        torch_version: [torch==1.10, torch]
+        torch_version: [torch==1.11, torch]
     steps:
       - name: Checkout repository
         uses: actions/checkout@v1
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -12,7 +12,7 @@ jobs:
         os: [ubuntu-18.04, macos-latest, windows-latest]
         # Lowest and highest, no version specified so that 
         # new releases get automatically tested against
-        torch_version: [torch==1.11, torch]
+        torch_version: [torch==1.10, torch]
     steps:
       - name: Checkout repository
         uses: actions/checkout@v1
--- a/bindings/python/benches/test_pt.py
+++ b/bindings/python/benches/test_pt.py
@@ -81,7 +81,7 @@ def test_pt_sf_load_gpu(benchmark):
         assert torch.allclose(v, tv)
 
 
-@pytest.mark.skipif(not torch.backends.mps.is_available(), reason="requires mps")
+@pytest.mark.skipif(not hasattr(torch.backends, "mps") or not torch.backends.mps.is_available(), reason="requires mps")
 def test_pt_pt_load_mps(benchmark):
     # benchmark something
     weights = create_gpt2(12)
@@ -95,7 +95,7 @@ def test_pt_pt_load_mps(benchmark):
         assert torch.allclose(v, tv)
 
 
-@pytest.mark.skipif(not torch.backends.mps.is_available(), reason="requires mps")
+@pytest.mark.skipif(not hasattr(torch.backends, "mps") or not torch.backends.mps.is_available(), reason="requires mps")
 def test_pt_sf_load_mps(benchmark):
     # benchmark something
     weights = create_gpt2(12)
--- a/bindings/python/src/lib.rs
+++ b/bindings/python/src/lib.rs
@@ -384,6 +384,47 @@ enum Storage {
     TorchStorage(GILOnceCell<PyObject>),
 }
 
+#[derive(Debug, PartialEq, Eq, PartialOrd)]
+struct Version {
+    major: u8,
+    minor: u8,
+    patch: u8,
+}
+
+impl Version {
+    fn new(major: u8, minor: u8, patch: u8) -> Self {
+        Self {
+            major,
+            minor,
+            patch,
+        }
+    }
+
+    fn from_string(string: String) -> Result<Self, &'static str> {
+        let mut parts = string.split('.');
+        let major_str = parts.next().ok_or("Torch major version missing")?;
+        let minor_str = parts.next().ok_or("Torch minor version missing")?;
+        let patch_str = parts.next().ok_or("Torch path version missing")?;
+        let mut patch_parts = patch_str.split('+');
+        let patch_str = patch_parts.next().ok_or("Torch path version missing")?;
+
+        let major = major_str
+            .parse()
+            .map_err(|_| "Python major version not an integer")?;
+        let minor = minor_str
+            .parse()
+            .map_err(|_| "Python minor version not an integer")?;
+        let patch = patch_str
+            .parse()
+            .map_err(|_| "Python patch version not an integer")?;
+        Ok(Version {
+            major,
+            minor,
+            patch,
+        })
+    }
+}
+
 /// Opens a safetensors lazily and returns tensors as asked
 ///
 /// Args:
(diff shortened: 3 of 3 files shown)


thomasw21 (commented):
Can you share your lib versions? One thing I'm wondering is this something that's integrated? https://github.com/huggingface/accelerate/pull/699

Also makes me think that https://github.com/pytorch/pytorch/pull/82603 has been merged and should be part of 1.13 so we can probable update this part to increase the speed: https://github.com/huggingface/accelerate/pull/699/commits/f1939bccf57dc5b8246d88089d5137306b679c43

Narsil:
torch==1.13 (the insane speedup is available for every version >= 1.11).

Basically.

```python
# Untyped required for slicing into it.
storage = torch.ByteStorage.from_file("model.safetensors).untyped()

tensor = torch.assarray(storage[start: stop], dtype=torch.uint8).view(dtype=dtype).reshape(shape)
```
Is basically what goes on.

sgugger (commented):
Quick question compared to what was there before: it looks like this loads the whole checkpoint in RAM (as a torch tensor) then takes a slice of it when we request a tensor. Whereas before, when asking for a given tensor, we were only loading that particular tensor in RAM I believe?

Narsil:
@sgugger  it seems that this  `Storage.from_file` memory maps the file (which is why it's instant).
mmap does NOT read to RAM by default, but it will but things in RAM when needed. Everything is handled by the OS which explains why it's faster in most settings.
Also it won't put the whole file in memory if not needed.

Meaning that there shouldn't be allocation until we actually use this storage. mmap should be in charge of allocating only what's needed from the disk.
It uses `shared=False` which supposedly doesn't attempt to write back anything on file if the tensor gets modifiedhttps://pytorch.org/docs/stable/storage.html#torch.TypedStorage.from_file.

thomasw21:
Just FYI: while installing on my setup (mac m1) I had the following error:
```
============== DUMMY ====================
NOTE: Redirects are currently not supported in Windows or MacOs.
Loaded in 0:00:05.769029
Setting `pad_token_id` to `eos_token_id`:50256 for open-end generation.
[{'generated_text': 'test.\n\nThe first thing to note is that the "new" version of the app is not'}]
Ran in 0:00:06.126767
============== NO ALLOC ====================
Lookup in 0:00:00.478613
model from config in 0:00:07.093166
tokenizer in 0:00:12.818499
Loaded in 0:00:12.831270
Loaded weights in 0:00:12.845461
Loaded on model in 0:00:12.847576
Setting `pad_token_id` to `eos_token_id`:50256 for open-end generation.
[1]    30634 bus error  python test.py
```

Don't think we should merge before having a better idea of why that is.

Narsil:
> Just FYI: while installing on my setup (mac m1) I had the following error:

Confirmed failure within PyTorch:

```python
import torch
from huggingface_hub import hf_hub_download
import os
import json
import mmap
from accelerate.big_modeling import init_empty_weights
import datetime

from transformers import AutoModelForCausalLM, AutoConfig, AutoTokenizer, pipeline
from huggingface_hub import hf_hub_download


def load_file(filename, device):
    with open(filename, mode="r", encoding="utf8") as file_obj:
        with mmap.mmap(file_obj.fileno(), length=0, access=mmap.ACCESS_READ) as m:
            header = m.read(8)
            n = int.from_bytes(header, "little")
            metadata_bytes = m.read(n)
            metadata = json.loads(metadata_bytes)

    size = os.stat(filename).st_size
    storage = torch.ByteStorage.from_file(filename, shared=False, size=size).untyped()
    offset = n + 8
    return {name: create_tensor(storage, info, offset) for name, info in metadata.items() if name != "__metadata__"}


DTYPES = {"F32": torch.float32}
device = "cpu"


def create_tensor(storage, info, offset):
    dtype = DTYPES[info["dtype"]]
    shape = info["shape"]
    start, stop = info["data_offsets"]
    return torch.asarray(storage[start + offset : stop + offset], dtype=torch.uint8).view(dtype=dtype).reshape(shape)


start = datetime.datetime.now()
filename = hf_hub_download("gpt2", filename="model.safetensors")

with init_empty_weights():
    config = AutoConfig.from_pretrained("gpt2")
    model = AutoModelForCausalLM.from_config(config).eval()
    print(f"model from config in {datetime.datetime.now() - start}")
    tokenizer = AutoTokenizer.from_pretrained("gpt2")
    print(f"tokenizer in {datetime.datetime.now() - start}")
    generator = pipeline("text-generation", model=model, tokenizer=tokenizer, max_new_tokens=20, do_sample=False)
    generator.device = torch.device(device)
    print(f"Loaded in {datetime.datetime.now() - start}")

weights = load_file(filename, device=device)
print(f"Loaded weights in {datetime.datetime.now() - start}")
model = generator.model
parameters = dict(model.named_parameters())

for name, tensor in weights.items():
    if name.endswith(".attn.bias"):
        continue
    full_name = f"transformer.{name}"
    module_name, param_name = full_name.rsplit(".", 1)
    module = model.get_submodule(module_name)
    current_tensor = parameters[full_name]
    module._parameters[param_name] = tensor
    if name == "wte.weight":
        model.lm_head._parameters["weight"] = tensor
model.to(device=device)

print(f"Loaded on model in {datetime.datetime.now() - start}")

out = generator("test")
print(f"Ran in {datetime.datetime.now() - start}")
print(out)
out = generator("test")
print(f"Ran in {datetime.datetime.now() - start}")
print(out)
```

This fails without any calls to `safetensors` so it's really a Storage bug.
And we're not using any private method as far as I can see except for `model.lm_head._parameters[...]`

PR #58, opened 2022-11-03
Supporting `safe_open` for TF and Flax.

Narsil:


bindings/python/src/lib.rs
```diff
@@ -717,40 +727,60 @@ fn create_tensor(
...
+        let shape = shape.to_vec();
+        let shape: PyObject = shape.into_py(py);
+        let mut tensor: &PyAny = tensor.getattr(intern!(py, "reshape"))?.call1((shape,))?;
+        if device != &Device::Cpu {
+            let device: PyObject = device.clone().into_py(py);
+            let kwargs = PyDict::new(py);
+            tensor = tensor
+                .getattr(intern!(py, "to"))?
+                .call((device,), Some(kwargs))?;
+        }
```
sgugger, line 756: This should probably be done only in `Framework` is `Framework::Pytorch`

sgugger (approved):
Thanks for working on this!

bindings/python/src/lib.rs
Narsil: Yes ! I need to handle the other framework differently (not sure if this lib should handle it or not though).

PR #59, opened 2022-11-04
Adding the convert scripts which will now prevent converting models

Narsil:
in case they will trigger warnigns in the `transformers` side.
Even if the model is perfectly fine, core maintainers fear an influx
of opened issues.

This is perfectly legit.
On the `transformers` side fixes are on the way: https://github.com/huggingface/transformers/pull/20042

We can wait for this PR to hit `main` before communicating super widely.

In the meantime this script of convertion will now prevent converting
models that would trigger such warnings (so the output of the script
**will** depend on the `transformers` freshness.

### DIFF
--- a/bindings/python/convert.py
+++ b/bindings/python/convert.py
@@ -2,18 +2,18 @@ import argparse
 import json
 import os
 import shutil
-from tempfile import TemporaryDirectory
 from collections import defaultdict
 from inspect import signature
-from typing import Optional, List
+from tempfile import TemporaryDirectory
+from typing import List, Optional
 
 import torch
 
-from huggingface_hub import CommitOperationAdd, HfApi, hf_hub_download, get_repo_discussions
+from huggingface_hub import CommitInfo, CommitOperationAdd, Discussion, HfApi, hf_hub_download
 from huggingface_hub.file_download import repo_folder_name
+from safetensors.torch import save_file
 from transformers import AutoConfig
 from transformers.pipelines.base import infer_framework_load_model
-from safetensors.torch import save_file
 
 
 class AlreadyExists(Exception):
@@ -30,15 +30,18 @@ def shared_pointers(tensors):
             failing.append(names)
     return failing
 
+
 def check_file_size(sf_filename: str, pt_filename: str):
     sf_size = os.stat(sf_filename).st_size
     pt_size = os.stat(pt_filename).st_size
 
     if (sf_size - pt_size) / pt_size > 0.01:
-        raise RuntimeError(f"""The file size different is more than 1%:
+        raise RuntimeError(
+            f"""The file size different is more than 1%:
          - {sf_filename}: {sf_size}
          - {pt_filename}: {pt_size}
-         """)
+         """
+        )
 
 
 def rename(pt_filename: str) -> str:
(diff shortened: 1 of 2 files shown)


bindings/python/convert.py
```diff
@@ -97,16 +103,17 @@ def convert_single(model_id: str, folder: str) -> List["CommitOperationAdd"]:
...
 
-    _, pt_model = infer_framework_load_model(model_id, config)
-    _, sf_model = infer_framework_load_model(folder, config)
+    _, (pt_model, pt_infos) = infer_framework_load_model(model_id, config, output_loading_info=True)
+    _, (sf_model, sf_infos) = infer_framework_load_model(folder, config, output_loading_info=True)
 
-    pt_model = pt_model
-    sf_model = sf_model
+    if pt_infos != sf_infos:
+        raise ValueError(f"different infos {model_id}")
```
sgugger, line 116: Maybe add some info on the differences?

bindings/python/convert.py
```diff
@@ -156,7 +163,7 @@ def previous_pr(api: "HfApi", model_id: str, pr_title: str) -> Optional["Discuss
             return discussion
 
 
-def convert(api: "HfApi", model_id: str, force: bool=False) -> Optional["CommitInfo"]:
+def convert(api: "HfApi", model_id: str, force: bool = False) -> Optional["CommitInfo"]:
```
sgugger, line 166: Didn't catch this before but we shouldn't use any api object and just the top level methods (`create_commit` is directly in the main init).

sgugger (approved):
Thanks for adding this!

bindings/python/convert.py
Narsil: It's quite noisy which is why I didn't include it (I mean infos can be quite numerous even on pure PT for transformers).

Doing a proper diff could help mitigate that but also relatively verbose code. I'll give it a stab.

bindings/python/convert.py
Narsil: Why ?

I figured it was actually better to pass around an HfApi codewise, so you can include a specific token (for instance to create conversions directly).
The original space convertion was passing around a Token, but I figure `HfApi` was even easier to transport around and there's less risk to forget to include `token=token` in whatever call.

sgugger: I don't think anyone who uses hugginface_hub uses the HfApi object.

sgugger: You should include a kwarg for token instead IMO

julien-c: > I don't think anyone who uses hugginface_hub uses the HfApi object.

I do  :)

julien-c: (this feature of setting a token on HfApi is quite recent though, but it's useful IMO)

sgugger: Let me rephrase, I don't think many people except Julien use `HfApi` :-p

Narsil: I did use it naturally actually an I think it's pretty cool to encapsulate common client options (most important of which the token).

I really think the "oh but you forgot to pass the token" is a real failure case, and it did happen when I used `get_repo_discussions` where I failed to pass the token and it would fail on private models on the space.

bindings/python/convert.py
Narsil: Diff added.

bindings/python/convert.py
Narsil: I'll go ahead and merge this.

Please open an issue if you feel so strongly about the `HfApi` object :)

PR #70, opened 2022-11-16
Fix doc build actions

mishig25:


mishig25:
@Narsil could you please check if everything is looking good at: https://moon-ci-docs.huggingface.co/docs/safetensors/pr_70/en/index

PR #78, opened 2022-11-17
Fixing unimplemented into real exception.

### LINKED ISSUE #76: Don't panic, use `NotImplementedError`
1. it is semantically incorrect
2. A dirty workaround has to be used, since `pyo3_runtime.PanicException` cannit be imported since `pyo3_runtime` is not a real module.

```python
def getPanicExceptionType():
    try:
        safetensors.safetensors_rust.serialize({"tensor_name": {"dtype": "test", "shape": [0, 0], "data": b""}}, None)
    except BaseException as ex:
        return ex.__class__

PanicException = getPanicExceptionType()
```

Narsil:
Fixes #76

### DIFF
--- a/bindings/python/src/lib.rs
+++ b/bindings/python/src/lib.rs
@@ -64,7 +64,9 @@ fn prepare(tensor_dict: HashMap<String, &PyDict>) -> PyResult<BTreeMap<String, T
                         "float64" => Dtype::F64,
                         "bfloat16" => Dtype::BF16,
                         dtype_str => {
-                            unimplemented!("Did not cover this dtype: {}", dtype_str)
+                            return Err(exceptions::PyException::new_err(format!(
+                                "dtype {dtype_str} is not covered",
+                            )));
                         }
                     }
                 }


KOLANICH:
It is worse than the way it used to be: now it throws a generic exception.

1. It should throw `NotImplementedError`, or even better, a custom exception class derived from it.
2. `PanicException` should also be exported, sofor us not to do perversions to get its class.

Narsil:
As I mentionned, no effort was given (yet) to classify all error types.

I can switch to `PyNotImplementedError` but I feel like doing a full sweep on all current `Exception` might be worthwile to sanitize.
For now, I'd rather have a generic exception everywhere than doing it only in 1 place.

It's possible to have custom exceptions in PyO3 relatively easily: https://pyo3.rs/latest/exception.html

In general, I personally favor reusing as much as possible native exceptions since users don't have to import them and if the meaning is correct then it does work.

Also simple is better.

As for "this is now worse" I do not agree. At least we're not panicking now.

KOLANICH:
>reusing as much as possible native exceptions

Subclasses are also caught. If one derives

```python
class FormatIsNotImplementedError(NotImplementedError):
    pass
```

, then 

```python
try:
    raise FormatIsNotImplementedError
except: NotImplementedError:
    pass
```
will catch it.

(`NotImplementedError` itself derives from `Exception`).

Narsil:
> Subclasses are also caught. If one derives

I know but when an exception occurs, you see its class, not where it's derived from. It makes it harder to catch if you want to do so, because now you have to figure out where it is actually defined. (Just to see if it derives from something else, or most often just catching it without really looking).

Subclassing does have use cases, I'm just saying subclassing everything does not make things necessarily better.

KOLANICH:
>I know but when an exception occurs, you see its class, not where it's derived from. It makes it harder to catch if you want to do so, because now you have to figure out where it is actually defined.

Well, if one wants to catch a generic type instead of precise one, he can guess that `FormatIsNotImplementedError` derives from `NotImplementedError`, because it is the only sensible to name that exception class this way if it is the case.

sgugger (approved):
Works for me!

Narsil:
> he can guess

When you have an error deep in prod that's not super easily reproduceable, well you don't want to do that (at least I don't).


But I do understand what you mean.

PR #79, opened 2022-11-17
Torch fixes

KOLANICH:


Narsil:
Thank you for this.

Do you mind adding a test for `load`. There used to be one, I must have deleted it somehow or it's not passing during the CI which would be concerning.

KOLANICH:
It seems that `safetensors.pytorch.load` has never been tested.

bindings/python/tests/test_pt.py
```diff
@@ -1,33 +1,48 @@
...
-from safetensors.torch import load_file, save_file
+from safetensors.torch import load, load_file, save, save_file
 
+this_dir = Path(__file__).parent
+data_dir = this_dir / "data"
 
 class TorchTestCase(unittest.TestCase):
-    def test_odd_dtype(self):
-        data = {
+    PT_MMAP_FILE = str(data_dir / "out_safe_pt_mmap_small.safetensors")  # FixMe: must accept Path!
```
Narsil, line 13: Do you mind not changing the style of this code ?

We don't need `pathlib` so it's not used on purpose (accepting it as an argument for users that prefer it, is on the other hand desirable).
I personally dislike pathlib because I have encountered bad bugs with it, and I feel it provides little value overall.

Overall your code hurts the very essence of what makes good tests according to me:

**Tests must be readable with no further context and the reader must able by purely reading this code if it's correct or not.**
If this property is not respected, and a reader cannot infer if the code is valid or not simply, then how is it better than the code it is supposed to test. If one has to think hard about correctness, then let's validate directly the actual functional code and skip the test.
That being said this is a principle, and sometimes it's not applicable. This is not the case here imho.

- This implies that repetition in tests is more than welcome, because as a reader I don't have to read 2, 10 or 100 functions to understand what the test is doing, all the context is there right in the current test I am reading.
- Here you introduce multiple levels of indirection, `cls` and `self._class` which I have no clue as a reader in which state they are in when reading locally. Did anyone modify them during the tests ? I have no clue.

I am happy to modify this code for you if you don't want to do it.

KOLANICH:
>I am happy to modify this code for you if you don't want to do it.

Feel free to do it.

>Tests must be readable with no further context and the reader must able by purely reading this code if it's correct or not.

IMHO my tests are in some sense more readable because the similarity and differencies between the tests

```python
def test_odd_dtype(self):
        self.verify_loaded_data(load_file(self.__class__.PT_MMAP_FILE))

    def test_load(self):
        self.verify_loaded_data(load(self.__class__.MMAP_SERIALIZATION))
```

is made visible: one loads data from the file, another one loads data from the buffer, and in other aspects they are the same. One doesn't have to check the piece of code you propose to inline within the both tests, one should do it only once, and that it matches the name we assign to it.

The code can be made even better by passing the data to those functions via arguments instead of via `cls`, to highlight, that the args are the same.

>Here you introduce multiple levels of indirection, cls and self.__class__ which I have no clue as a reader in which state they are in when reading locally.

1. `self.__class__.` is redundant (just `self.` will work fine), but `__class__` is intentionally used to highlight that this variable is owned by the class.

2. Class-owned variables are usually not meant to be mutated by test methods (and if we see a class variable, we usually assumme it is not mutated), but are meant to be mutated by `setUpClass` and `tearDownClass`. If they are mutated by test methods, it makes tests interdependent, but this violates the principles of unit testing.

>`out_safe_pt_mmap_small.safetensors`

IMHO this file name should be changed. There is nothing related to  memory mapping within the file itself.

>We don't need pathlib so it's not used on purpose. I personally dislike pathlib because I have encountered bad bugs with it, and I feel it provides little value overall.

I use `pathlib` since 2015. The bugs in my code related to the use of `pathlib` that I remember:

1. stacking a path looking like an absolute one over a `Path` will result into the absolute path.
2. `Path(".").parent` doesn't work. If one has to make the path absolute before that with `.resolve().absolute()`.
3. If one needs to get relative paths via `relative_to`, then again `resolve().absolute()` before the operation is needed.

Well, not so many, IMHO. In other aspects `pathlib.Path` is far more convenient than `os.path`.

>If this property is not respected, and a reader cannot infer if the code is valid or not simply, then how is it better than the code it is supposed to test.

I don't think about the testing this way. I think about testing not as about pinning the implementation under test to exact examples, but as about using 2 (or more, more the better!) different implementations that are the same in the important aspects ot the desired system operation. And because the impls are different, it is unlikely they both fail the same way. So if behaviour of one impl is different from the behaviour of another impl, then either one of them is wrong, or we test matching on non-essential behaviour that should not be required to match.

So, having considered my arguments, feel free to modify the code the way you feel is the right balance between high-levelness and DRYness and the low-levelness you insist on. Feel free to squash commits and do what you want. You'll have to create an own PR, so feel free to close this one.

Narsil:
> Feel free to do it.

Thanks.

> IMHO my tests are in some sense more readable because the similarity and differencies between the tests
> 

I used to think the same, but as test code gets more complex I have seen MANY examples where the tests was passing and yet the bug was introduced because the test was subtly incorporating the bug too. The easiest example is :

```python
for a, b in zip(calculated, expected):
    self.assertEqual(a, b)
```

Make `calculated` the empty list and boom test is still passing and you never see the bug. Yes, testing for the length before would solve that, but it's rarely done or thought about, and this is just the easiest example I can think of.
Removing indirections has always solved those issues in the long term in the projects I've worked on.

> 
> is made visible: one loads data from the file, another one loads data from the buffer, and in other aspects they are the same. One doesn't have to check the piece of code you propose to inline within the both tests, one should do it only once, and that it matches the name we assign to it.


>     2. Class-owned variables are usually not meant to be mutated by test methods (and if we see a class variable, we usually assumme it is not mutated), but are meant to be mutated by `setUpClass` and `tearDownClass`. If they are mutated by test methods, it makes tests interdependent, but this violates the principles of unit testing.
> 
> 
> > `out_safe_pt_mmap_small.safetensors`
> 
> IMHO this file name should be changed. There is nothing related to memory mapping within the file itself.

True

> 
> > We don't need pathlib so it's not used on purpose. I personally dislike pathlib because I have encountered bad bugs with it, and I feel it provides little value overall.
> 
> I use `pathlib` since 2015. The bugs in my code related to the use of `pathlib` that I remember:
> 
>     1. stacking a path looking like an absolute one over a `Path` will result into the absolute path.
> 
>     2. `Path(".").parent` doesn't work. If one has to make the path absolute before that with `.resolve().absolute()`.
> 
>     3. If one needs to get relative paths via `relative_to`, then again `resolve().absolute()` before the operation is needed.

I had multiple, the worst offenders were code that tried to do something different based on type (either path or string) and using multiple libs that used either one or the other was just hard. Worst kind of bug was the silent code (everything seems to be working but it wasn't doing was it was supposed to do). Putting everything to string just works.

Also when some value is of the wrong type the errors start making no sense with the `/` operation.

Again this is very personal, and I know a lot of devs find it really nice.
  
> Well, not so many, IMHO. In other aspects `pathlib.Path` is far more convenient than `os.path`.

Let's agree to disagree.

> > If this property is not respected, and a reader cannot infer if the code is valid or not simply, then how is it better than the code it is supposed to test.
> 
> I don't think about the testing this way. I think about testing not as about pinning the implementation under test to exact examples, but as about using 2 (or more, more the better!) different implementations that are the same in the important aspects ot the desired system operation. And because the impls are different, it is unlikely they both fail the same way. So if behaviour of one impl is different from the behaviour of another impl, then either one of them is wrong, or we test matching on non-essential behaviour that should not be required to match.

I have never written code twice in the same language (therefore under the same test suite) and used testing to verify "matching" property.
I don't have the time for it. And I have never ever seen two distinct codebases that really fulfilled any sort of matching spec. Litterally none. Regex engines are always different somehow, browsers well we know what's there. Audio/video players are also never reaching the same results that would pass each other test suites. Python interpreters/compilers etc..
And it seems like a very edge case to me.

Tests are built to prove that when you make changes to a codebase (either fixing a bug or adding a feature)  you are not breaking the old stuff. It's not the same codebase anymore, so there should be something changing. You just want to fix part of it.
If that part is unreadable, or needs any form of context which might induce readers into erroneously having confidence the code works then I think it's overall detrimental. (Like the zip example, if the test is green, I really think that my code is working, which it is not since I return an empty list). 

 
> So, having considered my arguments, feel free to modify the code the way you feel is the right balance between high-levelness and DRYness and the low-levelness you insist on. Feel free to squash commits and do what you want. You'll have to create an own PR, so feel free to close this one.

Thanks I will do.
And thanks for all the contribution. 
I may not agree with all the points you raised or some aspects of code management, but I'm really glad you contributed, you did raised a lot of valid concern/points and fixed some holes.

Narsil:
Closed in favor of https://github.com/huggingface/safetensors/pull/90

PR #80, opened 2022-11-17
[docs] Loading speed benchmark

mishig25:
Turning [the speed benchmark snippets](https://gist.github.com/Narsil/c452fbe0cd41c1d4ed226ad441135e40) into doc page

### DIFF
--- a/docs/source/speed.mdx
+++ b/docs/source/speed.mdx
@@ -35,15 +35,13 @@ Download safetensors & torch weights for gpt2:
 >>> # Make sure the experimental gpu loading feature is off since we are testing CPU here
 >>> if "SAFETENSORS_FAST_GPU" in os.environ: del os.environ["SAFETENSORS_FAST_GPU"]
 
->>> device = "cpu"
-
 >>> start_st = datetime.datetime.now()
->>> weights = load_file(sf_filename, device=device)
+>>> weights = load_file(sf_filename, device="cpu")
 >>> load_time_st = datetime.datetime.now() - start_st
 >>> print(f"Loaded safetensors {load_time_st}")
 
 >>> start_pt = datetime.datetime.now()
->>> weights = torch.load(pt_filename, map_location=device)
+>>> weights = torch.load(pt_filename, map_location="cpu")
 >>> load_time_pt = datetime.datetime.now() - start_pt
 >>> print(f"Loaded pytorch {load_time_pt}")
 
@@ -61,18 +59,16 @@ on CPU, safetensors is faster than pytorch by: 76.6 X
 >>> # it's been tested on many different environments
 >>> os.environ["SAFETENSORS_FAST_GPU"] = "1"
 
->>> device = "cuda:0"
-
 >>> # CUDA startup out of the measurement
 >>> torch.zeros((2, 2)).cuda()
 
 >>> start_st = datetime.datetime.now()
->>> weights = load_file(sf_filename, device=device)
+>>> weights = load_file(sf_filename, device="cuda:0")
 >>> load_time_st = datetime.datetime.now() - start_st
 >>> print(f"Loaded safetensors {load_time_st}")
 
 >>> start_pt = datetime.datetime.now()
->>> weights = torch.load(pt_filename, map_location=device)
+>>> weights = torch.load(pt_filename, map_location="cuda:0")
 >>> load_time_pt = datetime.datetime.now() - start_pt
 >>> print(f"Loaded pytorch {load_time_pt}")
 
--- a/docs/source/speed.mdx
+++ b/docs/source/speed.mdx
@@ -22,6 +22,13 @@ Let's start by importing all the packages that will be used:
 >>> import torch
 ```
 
+Download safetensors & torch weights for gpt2:
+
+```py
+>>> sf_filename = hf_hub_download("gpt2", filename="model.safetensors")
+>>> pt_filename = hf_hub_download("gpt2", filename="pytorch_model.bin")
+```
+
 ### CPU benchmark
 
 ```py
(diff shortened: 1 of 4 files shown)


docs/source/speed.mdx
```diff
@@ -0,0 +1,83 @@
...
+```py
+>>> sf_filename = hf_hub_download("gpt2", filename="model.safetensors")
+>>> pt_filename = hf_hub_download("gpt2", filename="pytorch_model.bin")
+```
+
+### CPU benchmark
+
+```py
+>>> # Make sure the experimental gpu loading feature is off since we are testing CPU here
+>>> if "SAFETENSORS_FAST_GPU" in os.environ: del os.environ["SAFETENSORS_FAST_GPU"]
```
Narsil, line 36: We don't care about the value here. Let's remove this to make the example clearer I think.

docs/source/speed.mdx
```diff
@@ -0,0 +1,83 @@
...
+
+>>> device = "cpu"
```
Narsil, line 38: I think it might be easier to put this directly within the calls.
Saves just one line but :) (It was there because I used the same script for both gpu and cpu :) )

docs/source/speed.mdx
```diff
@@ -0,0 +1,83 @@
...
+>>> load_time_st = datetime.datetime.now() - start_st
+>>> print(f"Loaded safetensors {load_time_st}")
+
+>>> start_pt = datetime.datetime.now()
+>>> weights = torch.load(pt_filename, map_location=device)
+>>> load_time_pt = datetime.datetime.now() - start_pt
+>>> print(f"Loaded pytorch {load_time_pt}")
+
+>>> print(f"on CPU, safetensors is faster than pytorch by: {load_time_pt/load_time_st:.1f} X")
+Loaded safetensors 0:00:00.004015
```
Narsil, line 51: As much as we have to do clickbaity stuff on social media to get visibility unfortunately, I'd rather be conservative here, and more importantly explain WHY.

So we can keep the values, but we have to explain a little on what hardware it was taken, and that it works because it's relying on disk cache.

Probably something after the code like:

"'"
This speedup is due to the fact that this library avoids unnecessary copies by mapping the file directly. It is actually possible to do on [pure pytorch](https://gist.github.com/Narsil/3edeec2669a5e94e4707aa0f901d2282).

The current show speedup was gotten on X (os) Y (CPU) Z(disk type).
"""

Wdyt ?

docs/source/speed.mdx
```diff
@@ -0,0 +1,83 @@
...
+
+
+### GPU benchmark
+
+```py
+>>> # This is required because this feature hasn't been fully verified yet, but 
+>>> # it's been tested on many different environments
+>>> os.environ["SAFETENSORS_FAST_GPU"] = "1"
+
+>>> device = "cuda:0"
```
Narsil, line 64: Same remark, maybe let's fuse it within the lines directly

docs/source/speed.mdx
```diff
@@ -0,0 +1,83 @@
...
+>>> start_pt = datetime.datetime.now()
+>>> weights = torch.load(pt_filename, map_location=device)
+>>> load_time_pt = datetime.datetime.now() - start_pt
+>>> print(f"Loaded pytorch {load_time_pt}")
+
+>>> print(f"on GPU, safetensors is faster than pytorch by: {load_time_pt/load_time_st:.1f} X")
+Loaded safetensors 0:00:00.165206
+Loaded pytorch 0:00:00.353889
+on GPU, safetensors is faster than pytorch by: 2.1 X
+```
```
Narsil, line 83: Same thing I would really like to explain why.

"""
The speedup works because this library is able to skip unecessary CPU allocations. It is unfortunately not replicable in pure pytorch as far as we know. The library works by memory mapping the file, creating the tensor empty with pytorch and calling `cudaMemcpy` directly to move the tensor directly on the GPU.
"""

Narsil (approved):
LGTM but left some comments to hopefully improve.

docs/source/speed.mdx
mishig25: https://github.com/huggingface/safetensors/pull/80/commits/2881a94a162e92ebbbaca0bd21e8a2d9e079cedb

docs/source/speed.mdx
mishig25: https://github.com/huggingface/safetensors/pull/80/commits/2881a94a162e92ebbbaca0bd21e8a2d9e079cedb

docs/source/speed.mdx
mishig25: https://github.com/huggingface/safetensors/pull/80/commits/25d95f94dd763009d3f82def17951e21bfdb410f

docs/source/speed.mdx
mishig25: wdyt ? https://github.com/huggingface/safetensors/pull/80/commits/0ee6bd225dd0ed1adbde2f26d0338095a0cb75a8

docs/source/speed.mdx
mishig25: https://github.com/huggingface/safetensors/pull/80/commits/0ee6bd225dd0ed1adbde2f26d0338095a0cb75a8

I'm not sure how to find disk info on google colab, which I;m using

docs/source/speed.mdx
Narsil: I think you forgot the disk type.

We just need NVMe vs SDD vs HD I think. (the numbers are gotten when everything is already read, but still)

mishig25: https://github.com/huggingface/safetensors/pull/80#discussion_r1026256406

docs/source/speed.mdx
Narsil: Then it's OK :)

docs/source/speed.mdx
```diff
@@ -0,0 +1,88 @@
...
+
+The speedup works because this library is able to skip unecessary CPU allocations. It is unfortunately not replicable in pure pytorch as far as we know. The library works by memory mapping the file, creating the tensor empty with pytorch and calling `cudaMemcpy` directly to move the tensor directly on the GPU.
+The currently shown speedup was gotten on:
+* OS: Ubuntu 18.04.6 LTS Y (CPU) Z(disk type).
```
Narsil, line 85:
```suggestion
* OS: Ubuntu 18.04.6 LTS.
```

PR #81, opened 2022-11-17
Fix `std::ffi::c_uint not found in std::ffi`

KOLANICH:


KOLANICH:
>you need to update your rust version.

I'd prefer this lib to have compatibility to older rust versions. The one installed in my system from packages is 1.59 (and I refuse to use `rustup`).

Narsil:
I'd really rather stick to `ffi`. Should switch to `core::ffi` actually.

Even the docs suggest that https://doc.rust-lang.org/std/os/raw/index.html

Sorry but I'm not going to start backporting this lib. If you don't want to use rust stable this is perfectly fine, but don't force it upon others.

Issue #75, opened 2022-11-17
[Critical] Design defect: endianness is not stored and can be arbitrary


KOLANICH:
```python
import numpy as np
import safetensors.numpy

a = safetensors.numpy.save({"a": np.array(range(6), dtype='>u4')})
b = safetensors.numpy.save({"a":np.array(range(6), dtype='<u4')})
print(a)
print(b)
```

```python
b'7\x00\x00\x00\x00\x00\x00\x00{"a":{"dtype":"U32","shape":[6],"data_offsets":[0,24]}}\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x02\x00\x00\x00\x03\x00\x00\x00\x04\x00\x00\x00\x05'
b'7\x00\x00\x00\x00\x00\x00\x00{"a":{"dtype":"U32","shape":[6],"data_offsets":[0,24]}}\x00\x00\x00\x00\x01\x00\x00\x00\x02\x00\x00\x00\x03\x00\x00\x00\x04\x00\x00\x00\x05\x00\x00\x00'
```

Narsil:
No, this is by design, everything is supposed to be little endian.  https://github.com/huggingface/safetensors#notes.

We should definitely raise an error if it's coming as big-endian, but I deferred that code because I want to know if there are still big endian users out there and if there's any legit use case for trying to keep both.

IMO supporting both (like Fortran order) causes pain because now every reader has to wonder about what endianness is under the hood and paying the cost of conversion eventually. 

The current assumption is that almost all machines relevant for machine learning are using little-endian, so we can happily discard big-endian altogether and force those machines to write as little-endian, and they only have to pay the cost.

Since it's a big decision and I don't claim I know every kind of hardware out there, I delayed taking that relatively firm stance for now, I am waiting to be proven wrong :) .

KOLANICH:
>The current assumption is that almost all machines relevant for machine learning are using little-endian

I know.

>so we can happily discard big-endian altogether and force those machines to write as little-endian, and they only have to pay the cost.

But currently it is not done. Instead it just dumps arrays from memory as they are.

>Since it's a big decision and I don't claim I know every kind of hardware out there, I delayed taking that relatively firm stance for now, I am waiting to be proven wrong :) .

[...]

PR #82, opened 2022-11-17
Preventing big endian files to be saved.

Narsil:
Fixes https://github.com/huggingface/safetensors/issues/75

### DIFF
--- a/bindings/python/py_src/safetensors/numpy.py
+++ b/bindings/python/py_src/safetensors/numpy.py
@@ -1,3 +1,4 @@
+import sys
 from typing import Dict, Optional
 
 import numpy as np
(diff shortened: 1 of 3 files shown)


mishig25:
lgtm, I got 2 questions:
1. Why the same enforcing rule is not implemented for tensorflow & flax load_files ?
2. Since we are enforcing it on serialization side, there is nothing else to do on deserialization side. Is it correct ?

Narsil:
> lgtm, I got 2 questions:
> 
>     1. Why the same enforcing rule is not implemented for tensorflow & flax load_files ?

Those rely on numpy itself so they are enforcing it as a byproduct. (pytorch used to rely on numpy too, but I had to break for bfloat16 support)

> 
>     2. Since we are enforcing it on serialization side, there is nothing else to do on deserialization side. Is it correct ?

Yes that's the purpose, serialized files will always be little endian.

mishig25 (approved):

PR #84, opened 2022-11-17
Add a set of sample files to test other implementations.

KOLANICH:


Narsil:
Not sure I understand the purpose of this PR.
I think this is the script you used to do some testing right ?

Maybe we should create real tests ?
Also I don't think we should commit actual tensor files. there are some deserialization examples with hardcoded values directly in Python to verify the capacity to deserialize some types.

KOLANICH:
These are just a few sample files. Ones who want to create an own impl of `safetensors` (I have written a Kaitai Struct-based impl for `npy` a few years ago, I guess it'd be easy to reuse its code to create an impl of `safetensors`, the hardest part there was the json header) can use them to test and debug an own impl. In order to do it he doesn't have to to build `safetensors` python package, he can just download the precreated files.

The script just generates the files. It also generates license files for `reuse` tool (but they are not commited into the repo, we use`reuse` tool in Kaitai Struct own repo of test files to track the copyright status of files, since not all the files are generated by us, but some can be fetched from other places).

Narsil:
Thank you for this work, but I don't think this belongs within the repository.

Just like the gists I post sometimes to showcase how to load without this library directly in PyTorch, they are helpful, but they do not belong here. The code here should only be about manipulating the files or helping testing this particular implementation works.

PR #86, opened 2022-11-18
Accepting `pathlib.Path` in `save_file/load_file/safe_open`.

Narsil:


bindings/python/src/lib.rs
```diff
@@ -467,8 +468,8 @@ struct safe_open {
 #[pymethods]
 impl safe_open {
     #[new]
-    fn new(filename: &str, framework: Framework, device: Option<Device>) -> PyResult<Self> {
-        let file = File::open(filename)?;
+    fn new(filename: PathBuf, framework: Framework, device: Option<Device>) -> PyResult<Self> {
+        let file = File::open(filename.clone())?;
```
thomasw21, line 472: Why do you need to clone? Can you not pass a ref?

thomasw21 (approved):
Cool! Thanks! I think you need to update docstring as well?

sgugger (approved):
Thanks for adding support!

PR #87, opened 2022-11-18
Trigger doc build on specific changes

mishig25:


.github/workflows/delete_doc_comment.yml
```diff
@@ -3,6 +3,9 @@ name: Delete dev documentation
 on:
   pull_request:
     types: [ closed ]
+    paths:
+      - "docs/**"
+      - "bindings/python/py_src/**"
```
mishig25, line 8: combination of `types` & `paths` should work right?

Narsil: It should yes. It's an OR here.

mishig25: oh, I was expecting AND

in this case https://github.com/huggingface/safetensors/pull/88

Narsil: Ohhh sorry, types and paths is and AND, however multiples paths is OR between each other.

PR #99, opened 2022-11-22
Fixing Windows overflow issue with `c_long`.

Narsil:
Workaround : https://github.com/PyO3/pyo3/issues/2768

### DIFF
--- a/bindings/python/src/lib.rs
+++ b/bindings/python/src/lib.rs
@@ -875,6 +875,7 @@ impl PySafeSlice {
     }
 }
 
+// TODO Remove this once https://github.com/PyO3/pyo3/issues/2768 is released
 fn pyslice_new(py: Python<'_>, start: isize, stop: isize, step: isize) -> &PySlice {
     let slice: &PySlice = unsafe {
         let ptr = pyo3::ffi::PySlice_New(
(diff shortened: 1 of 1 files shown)


bindings/python/src/lib.rs
```diff
@@ -875,6 +875,18 @@ impl PySafeSlice {
...
+        let ptr = pyo3::ffi::PySlice_New(
+            pyo3::ffi::PyLong_FromLongLong(start as std::ffi::c_longlong),
+            pyo3::ffi::PyLong_FromLongLong(stop as std::ffi::c_longlong),
+            pyo3::ffi::PyLong_FromLongLong(step as std::ffi::c_longlong),
+        );
+        py.from_owned_ptr(ptr)
+    };
+    slice
+}
+
```
mfuntowicz, line 889: Ah nice you managed to wrap it in a function!

Should we have an unittest for this?

Narsil: I hesitated. Currently I wanted the tests to be only in Python here, meaning it's not easy to actually trigger the bug (need a file with over 2Go file size which is large in a test suite).

I could add a pure Rust test, but I also need to create the CI for it.

And IMO this is a PyO3 bug, so I'd rather the test is added over there (if they confirm) and we don't add too many extra layers to test PyO3 itself (as long as this doesn't become regular :)

sgugger (approved):
Thanks!

PR #105, opened 2022-11-23
Fix `convert.py` to handle diffusers, and its transformers part

Narsil:
Turns out `diffusers` does rely on `transformers`:

`pytorch_model.safetensors` was not a recognized filename by
`transformers` which expects it to be named `model.safetensors`.

The convert_generic works by simply creating the new files and changing
the extension, which broke this.

Current proposed fix is to handle specially files named
`pytorch_model.bin` since they are more likely to be `transformers`
files.

This is however not necessarily true, we might want to protect with
checking against `model_info.library_name` though I feel like this is
better to just check against the name, since it's still likely to be a
transformers/diffusers/sentence-transformers thing.


sgugger (approved):
LGTM, thanks for fixing!

PR #119, opened 2022-11-30
Adding `cargo fuzz` and found sanitation opportunities.

Narsil:
Enter `cargo fuzz` 

Which tries to send wrong payloads and see how your code performs.
It spotted a few ways which could cause the deserialization to panic.

I also ran `miri` on tests https://github.com/rust-lang/miri and everything went through. But I need to check how to run it on the Python part to gain more confidence in the correctness of the current code.

### DIFF
--- a/safetensors/fuzz/.gitignore
+++ b/safetensors/fuzz/.gitignore
@@ -1,0 +1,4 @@
+target
+corpus
+artifacts
+coverage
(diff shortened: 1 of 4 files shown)


sgugger (commented):
@McPatate is probably better suited to comment than me on this one 😅

McPatate (approved):
Nice ! Maybe for the python part you'll need to use a python fuzzer, https://github.com/google/atheris ?
Or I wonder if by doing some pyo3-fu you could maybe get it to work, but it sounds easier to just use a python fuzzer

PR #131, opened 2022-12-13
Update README.md

thomasw21:


Narsil (approved):
LGTM.

The source of the README.md is actually rustdoc within `src/lib.rs`  and everything is recreated using 
` cargo readme > README.md` 
`cargo readme > ../README.md`

Which prevents forgetting one of them (since the doc lives in multiple places like https://docs.rs/safetensors/0.2.6/safetensors/

PR #139, opened 2022-12-21
Update docstrings & type-hints

alvarobartt:
This is just a minor PR to update some wrong imports in the docstrings and also to add some missing type-hints.


Narsil:
Thank you for this !

PR #140, opened 2022-12-23
Making SAFETENSORS_FAST_GPU=1 work on Windows again.

Narsil:
Turns out since we changed the loading of cudaMemcpy from looking
for cudart, to simply loading `_C.so` from Python, we broke Windows
fast path. That's because of this: https://github.com/nagisa/rust_libloading/issues/110
By default windows loads the DLL without being able to fetch nested DLLs.

There's a few choices to fix that:
- Use `load_with_flags` with a flag that would allow loading subdlls. 
  That's platform specific, and I got confused by all the options:
  https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-loadlibraryexw#parameters 
- Use `open_already_loaded`. This seems particularly interesting  because despite
  being OS specific, it will only load DLLs already loaded by the program which happens to be safe.
  So by using it we don't need `unsafe`, but we still need to manually look for cudart in the subdirectory (which is brittle).

Decided to go for Option 2.

Hopefully fixes: https://github.com/AUTOMATIC1111/stable-diffusion-webui/issues/5893 (This is not the only part of the fix)

### DIFF
--- a/bindings/python/src/lib.rs
+++ b/bindings/python/src/lib.rs
@@ -322,9 +322,7 @@ fn find_cudart(module: &PyModule) -> Option<Library> {
         let mut path = path.ok()?;
         path.pop();
         path.push("lib");
-        println!("Using path {path:?}");
         let paths = std::fs::read_dir(path).ok()?;
-        println!("Using paths {paths:?}");
         let cudart_paths: Vec<_> = paths
             .into_iter()
             .filter_map(|p| {
@@ -337,10 +335,8 @@ fn find_cudart(module: &PyModule) -> Option<Library> {
                 }
             })
             .collect();
-        println!("Using cudart paths {cudart_paths:?}");
         let cudart_path = cudart_paths.get(0)?;
-        println!("Using cudart_path {cudart_path:?}");
-        unsafe { libloading::os::windows::Library::open_already_loaded(cudart_path).ok()? }
+        Library::from(libloading::os::windows::Library::open_already_loaded(cudart_path).ok()?)
     };
 
     Some(lib)
--- a/bindings/python/src/lib.rs
+++ b/bindings/python/src/lib.rs
@@ -300,17 +300,49 @@ fn find_cudart(module: &PyModule) -> Option<Library> {
     if var != "1" {
         return None;
     }
-    let path: std::path::PathBuf = module
+
+    let path: PyResult<std::path::PathBuf> = module
         .getattr(intern!(module.py(), "_C"))
         .ok()?
         .getattr(intern!(module.py(), "__file__"))
         .ok()?
-        .extract()
-        .ok()?;
+        .extract();
+
     // SAFETY: This is unsafe because the library might run arbitrary code
     // So it's really important to make sure we are targeting the correct
     // library.
-    let lib = unsafe { Library::new(path).ok()? };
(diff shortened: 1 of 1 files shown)


sgugger (approved):
LGTM for what it's worth :-)
Thanks!

bindings/python/src/lib.rs
```diff
@@ -300,17 +300,45 @@ fn find_cudart(module: &PyModule) -> Option<Library> {
...
-    let lib = unsafe { Library::new(path).ok()? };
+        .extract();
+
+    #[cfg(not(target_os = "windows"))]
+    let lib = {
+        let path = path.ok()?;
+        // SAFETY: This is unsafe because the library might run arbitrary code
+        // So it's really important to make sure we are targeting the correct
+        // library.
+        unsafe { Library::new(path).ok()? }
```
McPatate, line 317: Is there a safe alternative to this ?

McPatate (approved):
Lgtm :)

bindings/python/src/lib.rs
```diff
@@ -300,17 +300,45 @@ fn find_cudart(module: &PyModule) -> Option<Library> {
...
+                let filename = p.file_name()?.to_string_lossy();
+                if filename.starts_with("cudart") {
+                    Some(p)
+                } else {
+                    None
+                }
+            })
+            .collect();
+        let cudart_path = cudart_paths.get(0)?;
+        Library::from(libloading::os::windows::Library::open_already_loaded(cudart_path).ok()?)
```
McPatate, line 339: As safe as the DLL named `cudart_path` is safe :smile:

bindings/python/src/lib.rs
Narsil: Not for Linux. 

https://docs.rs/libloading/latest/libloading/struct.Library.html

What makes it safe is that the library we're loading is already loaded by torch itself. On Windows, there's a way to load only already loaded librairies making the operation safe, but it seems the flag for it on Linux is not widely available:https://github.com/nagisa/rust_libloading/issues/110

bindings/python/src/lib.rs
Narsil: No, it won't load a DLL that isn't already loaded by the program ! So if someone puts a fake DLL it wouldn't get loaded (if I understand what I read, I didn't check it out myself)

McPatate: It does seem to limit the attack surface quite a bit, I'm unfamiliar with what "already loaded" mechanics mean though

Narsil: When you load a program that has linked dynamic libraries, the OS will go and resolve those DLLs at startup time. (It won't resolve individual symbols though so you might have a later error when accessing said functions).

Since `torch` uses `torch._C.so` which is a DLL which is linked against the aforementionned DLLs it will load `cudart-xxx.dll` too.

McPatate: Thanks for explaining !

NeedsMoar: torch loads kernel32.dll and maps LoadLibraryEx and a couple other functions into torch.kernel32.LoadLibraryEx() in its namespace and leaves it there for the duration of its run on Windows. 

I'm not too worried about whatever other DLL gets loaded in comparison.   👯

PR #143, opened 2022-12-28
Fixing torch version parsing.

### LINKED ISSUE #142: Version parsing fails on pytorch alpha
Pytorch may reportedly have a version string like `2.0.0a0+gitd1123c9` when compiled from source. That makes the safetensors Version parser fail with `Python patch version not an integer` https://github.com/huggingface/safetensors/blob/ba8de5ff52eee70e6bbc7e7c49a86c251da357d7/bindings/python/src/lib.rs#L463-L465

Reported by @lstein.

Narsil:
Fixes #142 

When compiling from source, torch version can be more complex and include `0a0` for alpha versions. so we need to parse this better

### DIFF
--- a/bindings/python/src/lib.rs
+++ b/bindings/python/src/lib.rs
@@ -452,14 +452,14 @@ impl Version {
         let major_str = parts.next().ok_or_else(err)?;
         let minor_str = parts.next().ok_or_else(err)?;
         let patch_str = parts.next().ok_or_else(err)?;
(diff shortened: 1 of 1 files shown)


JasonGiedymin:
I'm working on a torch version for rocm (AMD) and got bit by this. Versions used by that team look like `X.Y.Z+rocm5.2`. Even if following semver, patch should be lenient.

sgugger (approved):
Thanks for the fix!

Issue #44, opened 2022-10-28
Ability to remotely parse metadata over small HTTP requests


julien-c:
In this branch: https://github.com/huggingface/safetensors/compare/julien-c/js I pushed a proof-of-concept of how, given the simplicity of the format, one can fetch metadata about the weights over small (Range) HTTP requests.

The code is JS (can run in browsers or in Node) but it would be similar in any language.

Here's an example on how to fetch the header in a single file for instance:

```typescript
async function parseSingleFile(url: URL): Promise<FileHeader> {
	const bufLengthOfHeaderLE = await (
		await fetch(url, {
			headers: {
				Range: "bytes=0-7",
			},
		})
	).arrayBuffer();
	const lengthOfHeader = new DataView(bufLengthOfHeaderLE).getBigUint64(
		0,
		true
	);
	/// ^little-endian
	const header: FileHeader = await (
		await fetch(url, {
			headers: {
				Range: `bytes=8-${7 + Number(lengthOfHeader)}`,
			},
		})
	).json();
	/// no validation for now, we assume it's a valid FileHeader.
	return header;
}
```

where a `FileHeader` type is defined as:

```typescript
type TensorName = string;
type Dtype =
	| "F64"
	| "F32"
	| "F16"
	| "I64"
	| "I32"
	| "I16"
	| "I8"
	| "U8"
	| "BOOL";

interface TensorInfo {
	dtype: Dtype;
	shape: number[];
	data_offsets: [number, number];
}

type FileHeader = Record<TensorName, TensorInfo> & {
	__metadata__: Record<string, string>;
};
```

## Results

As a fun first experiment, I compute the number of params per dtype for all models currently with a safetensors version on the HuggingFace Hub.

Here's the results:

model | safetensors | params
--- | --- | ---
gpt2 | single-file | { 'F32' => 137022720 }
roberta-base | single-file | { 'F32' => 124697433, 'I64' => 514 }
Jean-Baptiste/camembert-ner | single-file | { 'F32' => 110035205, 'I64' => 514 }
roberta-large | single-file | { 'F32' => 355412057, 'I64' => 514 }
bigscience/bloom-560m | single-file | { 'F16' => 559214592 }
hf-internal-testing/tiny-random-bert-safetensors | single-file | { 'F32' => 127463, 'I64' => 512 }
hf-internal-testing/tiny-random-bert-sharded-safetensors | index-file | { 'F32' => 87929, 'I64' => 512 }
[...]

PR #148, opened 2023-01-04
[Major Change] Enforcing tensor alignment

Narsil:
- Now the header will automatically align itself to 8 bytes (f64) with
  appending extra spaces as necessary.
- This will allow extra fast memory mapping by reinterpreting bytes as
  f32/f64 etc.. Unaligned bytes do not allow for this. https://www.reddit.com/r/rust/comments/tanaxm/mutating_a_buffer_of_u8s_as_f32s_in_place/
- This does not change contiguousness of tensors
- This does not change the actual spec (we're just putting extra valid bytes
  in the header and using a different serialization ordering)
- Readers should still be able to read old files, they would just need
  to be copied before being cast as their final destination when using
  mmap
- This has no effect for GPU since copy is already necessary (*I think*,
  depends on the cuda API actually if it allows filling f32 addresses
  from raw unaligned bytes).

This change will only be interesting if things like https://github.com/Narsil/fast_gpt2
actually pick up. And even with the copy, load times are still vastly
faster than `pytorch`/`transformers`.

### DIFF
--- a/bindings/python/src/lib.rs
+++ b/bindings/python/src/lib.rs
@@ -11,7 +11,7 @@ use pyo3::types::{PyByteArray, PyBytes, PyDict, PyList};
 use pyo3::{intern, PyErr};
 use safetensors::slice::TensorIndexer;
 use safetensors::tensor::{Dtype, Metadata, SafeTensors, TensorInfo, TensorView};
-use std::collections::{BTreeMap, HashMap};
+use std::collections::HashMap;
 use std::fs::File;
 use std::iter::FromIterator;
 use std::ops::Bound;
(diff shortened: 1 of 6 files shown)


julien-c:
would this change require changes to #44? or would it be transparent?

> And even with the copy, load times are still vastly
> superior to `pytorch`/`transformers`.

superior is ambiguous, I would add "(i.e., lower then)"

Narsil:
> would this change require changes to #44? or would it be transparent?

Transparent, it's only making sure the header JSON is of the appropriate size by adding extra padding.
(Also tensor order is modified, which readers should be agnostic too anyway).

> 
> > And even with the copy, load times are still vastly
> > superior to `pytorch`/`transformers`.
> 
> superior is ambiguous, I would add "(i.e., lower then)"

True!

McPatate (approved):
lgtm, what prompted the change from `BTreeMap` to `HashMap` ?

Narsil:
> lgtm, what prompted the change from `BTreeMap` to `HashMap` ?

complexity. O(log (n) ) vs O(1). Shouldn't matter in practice really, but since I don't rely on the structure for ordering anymore, I don't need BTree anymore.

Issue #154, opened 2023-01-08
It rejects to serialize `uint8`


KOLANICH:
```python
safetensors.numpy.save({"8": np.array(range(-3, 4), dtype=np.uint8)})
```

```ValueError: Safetensor format only accepts little endian```

PR #155, opened 2023-01-08
Fix byte order detection for `np.uint8` and other types for which byte order is not defined.

KOLANICH:
Fixes: #154

### DIFF
--- a/bindings/python/py_src/safetensors/numpy.py
+++ b/bindings/python/py_src/safetensors/numpy.py
@@ -165,4 +165,6 @@ def _byte_order(tensor: np.ndarray) -> str:
             return "<"
         else:
             return ">"
+    elif byteorder == "|":
+        return "<"
     return byteorder


Narsil:
Hi @KOLANICH ,

Thanks for the fix.

Do you mind not changing the return type `|` seems like the correct return type for `_byteorder` so we would need to just do 

`if byteorder not in {'<', '|'|:`  to fix it instead.

In addition, I think adding a test is worth it.

I can take care of everything if you want.

KOLANICH:
I was thinking about a better solution - just remove that function and replace it with `def is_endian_ok(dtype) -> bool:`. We currently don't support any endian except le and it is unlikely that anything except LE will be implemented in any new hardware. TBH IMHO only le endian makes sense because ints of different sizes are compatible to each other (wrt zero padding in MSBs) and because it is easier to deal with in software-implemented arythmetics. It was strange that BE got used in any hardware at all given its drawbacks, such a simple idea as LE seems to be uncopyrightable and unpatentable.

KOLANICH:
and, BTW, feel free to do everything yourself.

Narsil:
> def is_endian_ok(dtype) -> bool:. We 

Even better idea, I did just that !

Narsil:
I had to create #160 since I can't commit to your branch unfortunately.

PR #165, opened 2023-01-17
Adding custom error type for users to properly catch.

### LINKED ISSUE #156: Expose `PanicException` in order to allow it be caught
```python
from safetensors.safetensors_rust import PanicException
```

Narsil:
Fixes #156

### DIFF
--- a/bindings/python/src/lib.rs
+++ b/bindings/python/src/lib.rs
@@ -2,7 +2,7 @@
 //! Dummy doc
 use libloading::{Library, Symbol};
 use memmap2::{Mmap, MmapOptions};
-use pyo3::exceptions;
+use pyo3::exceptions::PyException;
 use pyo3::once_cell::GILOnceCell;
 use pyo3::prelude::*;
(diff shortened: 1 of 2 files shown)


thomasw21 (approved):
Nice! Is there a way to re-use some of Python errors? Typically some of them should be ValueErrors

Narsil:
> Nice! Is there a way to re-use some of Python errors? Typically some of them should be ValueErrors

Regular Python choice. We could switch every exception to raw more "natural" exceptions (like `ValueError`) making them a bit more general, or we could use our custom type making them more catchable (and less natural).

I went the easy way and simply pushed everything into something catchable. If you want to override some to ValueError, you're welcome to it.

In python world, since there's no strong typing on exceptions and catching I tend to see much less value in error types than in rust for instance. Hence, I'm lowly opinionated about the end result and would favor the lowest common denominator in case of disagreement between different viewpoints.

sgugger (approved):
Thanks for working on this!

PR #166, opened 2023-01-19
Fixing the fact that the context manager wasn't properly cleaning up after itself.

### LINKED ISSUE #164: Handle to model file remains open after `safe_open` goes out of scope?
Hi, I want to read the metadata and tensors of a `.safetensors` file, update the metadata and write the file back (overwriting the original). I tried to do that with this code:

```python3
#!/usr/bin/env python

import sys
import os.path
import safetensors.torch

model_path = sys.argv[1]
if not model_path:
    print("Provide a model path.")
    exit(1)

if not os.path.splitext(model_path)[1] == '.safetensors':
    print("Must be .safetensors file.")
    exit(1)

tensors = {}
metadata = {}
with safetensors.torch.safe_open(model_path, framework="pt") as f:
  metadata = f.metadata()
  for k in f.keys():
    tensors[k] = f.get_tensor(k)

safetensors.torch.save_file(tensors, model_path, metadata)
```

But I get this error:

```
  File "C:\stable-diffusion-webui\extensions\sd-scripts\tset.py", line 24, in <module>
    safetensors.torch.save_file(tensors, model_path, metadata)
  File "C:\stable-diffusion-webui\extensions\sd-scripts\venv\lib\site-packages\safetensors\torch.py", line 71, in save_file
    serialize_file(_flatten(tensors), filename, metadata=metadata)
Exception: Error while serializing: IoError(Os { code: 1224, kind: Uncategorized, message: "The requested operation cannot be performed on a file with a user-mapped section open." })
```

I expect that the handle is closed when `safe_open` goes out of scope. I'm not using the "GPU fast load" option set by an envvar either

Also if I wait before the program exits and try to delete the file Explorer won't let me, I checked with LockHunter and it says no process is using the file

So if I use `safe_open` on several model files in a program they all get locked until the program exits

I am on Windows 10, using safetensors 0.2.8 from pip

Narsil:
Fixes #164 

What happened:
- Windows doesn't like to open to write memory mapped files (that's OK and would also prevent potential weird bugs on other platforms)
- Context manager in PyO3 do not drop themselves on `__exit__`, so resources are still held for the rest of the program.

So `safe_open` a file, even not doing anything, would open the file in memory mapped mode (to read the header) and would prevent writing over it afterwards.

Solution:

- Make a real Python wrapper context manager still named `safe_open` (so hopefully transparent).
- Make the new wrapper actually `del f` the rust class, which **does** call `Drop` and cleanup the ressources.

There are other way to solve this, but this seems quite simple. The biggest drawback is if users would actually be using `safetensors.safetensors_rust.safe_open` and not `safetensors.safe_open`. I think this is limited.
To prevent further potential misuse, this PR also renames the rust class to something private. `_safe_open`.

### DIFF
--- a/bindings/python/py_src/safetensors/__init__.py
+++ b/bindings/python/py_src/safetensors/__init__.py
@@ -1,7 +1,7 @@
 __version__ = "0.2.9"
 
 # Re-export this
-from .safetensors_rust import safe_open as rust_open  # noqa: F401
+from ._safetensors_rust import safe_open as rust_open, serialize, serialize_file, deserialize, SafetensorError
 
 
 class safe_open:
--- a/bindings/python/py_src/safetensors/__init__.py
+++ b/bindings/python/py_src/safetensors/__init__.py
@@ -1,4 +1,20 @@
 __version__ = "0.2.9"
 
 # Re-export this
-from .safetensors_rust import safe_open  # noqa: F401
+from .safetensors_rust import safe_open as rust_open  # noqa: F401
+
+
+class safe_open:
+    def __init__(self, *args, **kwargs):
+        self.args = args
+        self.kwargs = kwargs
+
+    def __getattr__(self, __name: str):
+        return getattr(self.f, __name)
+
+    def __enter__(self):
+        self.f = rust_open(*self.args, **self.kwargs)
+        return self
+
+    def __exit__(self, type, value, traceback):
+        del self.f
--- a/bindings/python/tests/test_simple.py
+++ b/bindings/python/tests/test_simple.py
@@ -67,7 +67,7 @@ class WindowsTestCase(unittest.TestCase):
             "b": torch.zeros((2, 3), dtype=torch.uint8),
         }
         save_file_pt(tensors, "./out.safetensors")
-        with safe_open("./out.safetensors", framework="pt"):
+        with safe_open("./out.safetensors", framework="pt") as f:
             pass
 
         with open("./out.safetensors", "w") as g:
--- a/bindings/python/tests/test_simple.py
+++ b/bindings/python/tests/test_simple.py
@@ -5,9 +5,8 @@ from pathlib import Path
 import numpy as np
 import torch
 
-from safetensors import safe_open
 from safetensors.numpy import load, load_file, save, save_file
-from safetensors.safetensors_rust import SafetensorError, serialize
+from safetensors import safe_open, SafetensorError, serialize
 from safetensors.torch import load_file as load_file_pt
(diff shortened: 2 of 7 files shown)


bindings/python/py_src/safetensors/__init__.py
```diff
@@ -1,4 +1,20 @@
...
+    def __init__(self, *args, **kwargs):
+        self.args = args
+        self.kwargs = kwargs
+
+    def __getattr__(self, __name: str):
+        return getattr(self.f, __name)
+
+    def __enter__(self):
+        self.f = rust_open(*self.args, **self.kwargs)
+        return self
```
thomasw21, line 17: Why not just return `self.f` and remove the `__getattr__`?

thomasw21 (commented):
Haven't tested since I don't access to a windows machine.

bindings/python/py_src/safetensors/__init__.py
McPatate: IIUC because you don't want to `del self` in the `__exit__` rather just the return value of `rust_open` given `self` references the `safe_open` ctx manager

bindings/python/tests/test_simple.py
```diff
@@ -59,6 +59,20 @@ def test_accept_path(self):
...
+        tensors = {
+            "a": torch.zeros((2, 2)),
+            "b": torch.zeros((2, 3), dtype=torch.uint8),
+        }
+        save_file_pt(tensors, "./out.safetensors")
+        with safe_open("./out.safetensors", framework="pt") as f:
+            pass
+
+        with open("./out.safetensors", "w") as g:
+            g.write("something")
```
McPatate, line 73: I'm not sure I understand how this tests if the file descriptor is correctly dropped.

Can you not have multiple open fds on a file?

McPatate (approved):

bindings/python/py_src/safetensors/__init__.py
thomasw21: I'm not sure I understand, it's fine to delete `self.f` in the `__exit__`, all I was suggesting was:
```suggestion
        return self.f
```

Narsil: Because then I wouldn't be able to `del` it. and it would keep things open still.

bindings/python/tests/test_simple.py
Narsil: This is the test that proves it fails before the actual modifications:

https://github.com/huggingface/safetensors/actions/runs/3957656229/jobs/6778277781

The answer is that no, at least not on Windows.

PR #176, opened 2023-02-07
Add PaddlePaddle backend

JunnYu:
Add PaddlePaddle backend.
image

### DIFF
--- a/bindings/python/benches/test_paddle.py
+++ b/bindings/python/benches/test_paddle.py
@@ -1,0 +1,53 @@
+import tempfile
+
+import numpy as np
+
+import paddle
+from safetensors.paddle import load_file, save_file
+
+
+def create_gpt2(n_layers: int):
+    tensors = {}
+    tensors["wte"] = paddle.zeros((50257, 768))
+    tensors["wpe"] = paddle.zeros((1024, 768))
+    for i in range(n_layers):
+        tensors[f"h.{i}.ln_1.weight"] = paddle.zeros((768,))
+        tensors[f"h.{i}.ln_1.bias"] = paddle.zeros((768,))
+        tensors[f"h.{i}.attn.bias"] = paddle.zeros((1, 1, 1024, 1024))
+        tensors[f"h.{i}.attn.c_attn.weight"] = paddle.zeros((768, 2304))
+        tensors[f"h.{i}.attn.c_attn.bias"] = paddle.zeros((2304,))
+        tensors[f"h.{i}.attn.c_proj.weight"] = paddle.zeros((768, 768))
+        tensors[f"h.{i}.attn.c_proj.bias"] = paddle.zeros((768,))
+        tensors[f"h.{i}.ln_2.weight"] = paddle.zeros((768,))
+        tensors[f"h.{i}.ln_2.bias"] = paddle.zeros((768,))
+        tensors[f"h.{i}.mlp.c_fc.weight"] = paddle.zeros((768, 3072))
+        tensors[f"h.{i}.mlp.c_fc.bias"] = paddle.zeros((3072,))
+        tensors[f"h.{i}.mlp.c_proj.weight"] = paddle.zeros((3072, 768))
+        tensors[f"h.{i}.mlp.c_proj.bias"] = paddle.zeros((768,))
+    tensors["ln_f.weight"] = paddle.zeros((768,))
+    tensors["ln_f.bias"] = paddle.zeros((768,))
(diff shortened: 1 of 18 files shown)


bindings/python/py_src/safetensors/paddle.py
```diff
@@ -0,0 +1,141 @@
+import os
+from typing import Dict, Optional, Union
+
+import numpy as np
+
+import paddle
+from safetensors import numpy
+
+
+np.bool = bool
```
Narsil: What is this for ?

bindings/python/setup.py
```diff
@@ -36,12 +37,19 @@ def deps_list(*pkgs):
 extras["numpy"] = deps_list("numpy")
 extras["tensorflow"] = deps_list("tensorflow")
 extras["jax"] = deps_list("jax", "flax")
+extras["paddlepaddle-gpu"] = deps_list("paddlepaddle-gpu")
```
Narsil: Why the `gpu` version? We don't require `tensorflow-gpu` so by pure symetric we shouldn't install the paddle-gpu either.

Cpu is fine for testing anyway (and if users have the gpu version, everything should work, correct ?)

bindings/python/benches/test_paddle.py
```diff
@@ -0,0 +1,53 @@
...
+    tensors["ln_f.weight"] = paddle.zeros((768,))
+    tensors["ln_f.bias"] = paddle.zeros((768,))
+    return tensors
+
+
+def test_paddle_paddle_load(benchmark):
+    # benchmark something
+    weights = create_gpt2(12)
+    with tempfile.NamedTemporaryFile() as f:
+        paddle.save(weights, f)
```
Narsil: This is wrong I think.
```suggestion
        paddle.save(weights, f.name)
```
At least I'm getting errors with this code.

Narsil (commented):
Thanks for the great PR.

I made some comments and took the liberty of modifying certain things (since I modified them, while looking at the PR, the work was done.)

Cheers.

bindings/python/src/lib.rs
```diff
@@ -35,6 +35,7 @@ type MemcpyFn = unsafe extern "C" fn(
 static TORCH_MODULE: GILOnceCell<Py<PyModule>> = GILOnceCell::new();
 static NUMPY_MODULE: GILOnceCell<Py<PyModule>> = GILOnceCell::new();
 static TENSORFLOW_MODULE: GILOnceCell<Py<PyModule>> = GILOnceCell::new();
+static PADDLE_MODULE: GILOnceCell<Py<PyModule>> = GILOnceCell::new();
```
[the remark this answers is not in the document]
JunnYu: i remove these  modifications and i got some error. I will remove `test_deserialization_safe_open` test
image

bindings/python/benches/test_paddle.py
JunnYu: oh, this is my mistake

bindings/python/setup.py
JunnYu: you are right, we donot need paddlepaddle-gpu

bindings/python/py_src/safetensors/paddle.py
JunnYu: i found this error. without  np.bool = bool,
image

JunnYu:
```python
from safetensors.paddle import load_file, save_file
import paddle
import numpy as np
# numpy version 1.24.2
# np.bool = bool
file_path = "model.safetensors"
tensors = {"embedding": paddle.randint(0, 2, (512, 1024)).cast("bool"), "attention": paddle.zeros((256, 256))}
save_file(tensors, file_path)
loaded2 = load_file(file_path)
```

bindings/python/src/lib.rs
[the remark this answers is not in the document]
ZHUI: > we need to remove those modifications. The rust codebase won't create those tensors directly, so just like TF and Flax, we don't open those either.

Hello,  since ` The rust codebase won't create those tensors directly`, it there another way to support `safe_open` for paddle backend ?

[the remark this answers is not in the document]
Narsil: If there is a *python* only way to read a safetensors file, we could use that in the rust code  (the binding can just call the equivalent Python code).

This is doable only if the tensor creation is purely lazy so it's still compatible with the slice loading.

PR #182, opened 2023-02-27
Deprecating loading files without specifying device (removing the default).

Narsil:
This is now only a deprecation notice.

The reason for this:
- Moving to `0.3` for the alignment modification allowing for more
  change.
- Not specifying the default has been a real performance hurt:
  https://github.com/AUTOMATIC1111/stable-diffusion-webui/issues/5893
  https://github.com/huggingface/diffusers/blob/e5810e686ea4ac499e325c2961808c8972dee039/src/diffusers/models/modeling_utils.py#L103

  This should only affect from disk -> CPU/GPU since this is where the
  location is modified. When loading from bytes, the location is already
  CPU so it's natural to use CPU (no alloc).

- Giving 2 "minor" versions before dropping support, this should allow
  users to have time to move.

### DIFF
--- a/bindings/python/py_src/safetensors/flax.py
+++ b/bindings/python/py_src/safetensors/flax.py
@@ -98,7 +98,7 @@ def load(data: bytes) -> Dict[str, jnp.DeviceArray]:
     return _np2jnp(flat)
 
 
-def load_file(filename: Union[str, os.PathLike]) -> Dict[str, jnp.DeviceArray]:
+def load_file(filename: Union[str, os.PathLike], device: str = None) -> Dict[str, jnp.DeviceArray]:
     """
     Loads a safetensors file into flax format.
 
@@ -108,6 +108,10 @@ def load_file(filename: Union[str, os.PathLike]) -> Dict[str, jnp.DeviceArray]:
         device (`Dict[str, any]`, *optional*, defaults to `cpu`):
             The device where the tensors need to be located after load.
             available options are all regular flax device locations
+
+            Note: In future versions, the default will be removed. The device
+            is required to be set to the real final device to get the best
+            load performance. Default will be removed in version >= 0.5.
 
     Returns:
         `Dict[str, jnp.DeviceArray]`: dictionary that contains name as key, value as `jnp.DeviceArray`
@@ -121,7 +125,7 @@ def load_file(filename: Union[str, os.PathLike]) -> Dict[str, jnp.DeviceArray]:
     loaded = load_file(file_path)
     ```
     """
-    flat = numpy.load_file(filename)
+    flat = numpy.load_file(filename, device=device)
     return _np2jnp(flat)
 
 
--- a/bindings/python/py_src/safetensors/paddle.py
+++ b/bindings/python/py_src/safetensors/paddle.py
@@ -71,13 +71,21 @@ def save_file(
     return numpy.save_file(np_tensors, filename, metadata=metadata)
 
 
-def load(data: bytes, device: str = "cpu") -> Dict[str, paddle.Tensor]:
+def load(data: bytes, device: str = None) -> Dict[str, paddle.Tensor]:
     """
     Loads a safetensors file into paddle format from pure bytes.
 
     Args:
         data (`bytes`):
             The content of a safetensors file
+        device (`Dict[str, any]`, *optional*, defaults to `cpu`):
+            The device where the tensors need to be located after load.
+            available options are all regular paddle device locations
+
+            Note: In future versions, the default will be removed. The device
+            is required to be set to the real final device to get the best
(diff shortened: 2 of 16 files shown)


OlivierDehaene (approved):

bindings/python/src/lib.rs
```diff
@@ -488,7 +488,19 @@ impl Open {
         let file = File::open(&filename).map_err(|_| {
             PyFileNotFoundError::new_err(format!("No such file or directory: {filename:?}"))
         })?;
-        let device = device.unwrap_or(Device::Cpu);
+        let device = device.unwrap_or_else(|| {
+            Python::with_gil(|py| -> PyResult<()>{
+                let warnings = PyModule::import(py, "warnings")?;
+                let builtins = PyModule::import(py, "builtins")?;
+                let warning = r#"Deprecated load without device specified. Please user `load_file(..., device="cpu")` or `safe_open(..., device="cpu")` to make sure you want to load on "cpu". Try to load on the device where the weights are needed directly as it could cause performance issues."#;
```
pcuenca, line 495:
```suggestion
                let warning = r#"Deprecated load without device specified. Please use `load_file(..., device="cpu")` or `safe_open(..., device="cpu")` to make sure you want to load on "cpu". Please, try to load weights on the device they are needed as doing otherwise could cause performance issues."#;
```

Narsil:
Needs further investigation: https://github.com/huggingface/diffusers/issues/2507

Current gut feeling, since we moved to TorchStorage, now it's skipping the CPU alloc too (which it wasn't before). 
But I need to check Windows too, as they seemed to exhibit very difference behaviors here).

bindings/python/py_src/safetensors/flax.py
```diff
@@ -98,7 +98,7 @@ def load(data: bytes) -> Dict[str, jnp.DeviceArray]:
     return _np2jnp(flat)
 
 
-def load_file(filename: Union[str, os.PathLike]) -> Dict[str, jnp.DeviceArray]:
+def load_file(filename: Union[str, os.PathLike], device: str = None) -> Dict[str, jnp.DeviceArray]:
```
sgugger, line 101: If putting a type annotation, this is an `Optional[str]` (here and everywhere below).

bindings/python/py_src/safetensors/paddle.py
```diff
@@ -71,13 +71,21 @@ def save_file(
...
 
-def load(data: bytes, device: str = "cpu") -> Dict[str, paddle.Tensor]:
+def load(data: bytes, device: str = None) -> Dict[str, paddle.Tensor]:
     """
     Loads a safetensors file into paddle format from pure bytes.
 
     Args:
         data (`bytes`):
             The content of a safetensors file
+        device (`Dict[str, any]`, *optional*, defaults to `cpu`):
```
sgugger, line 81:
```suggestion
        device (`Dict[str, any]`, *optional*, defaults to `"cpu"`):
```
Also the type is inconsistent with what is above.

bindings/python/py_src/safetensors/flax.py
```diff
@@ -109,6 +109,10 @@ def load_file(filename: Union[str, os.PathLike]) -> Dict[str, jnp.DeviceArray]:
             The device where the tensors need to be located after load.
             available options are all regular flax device locations
 
+            Note: In future versions, the default will be removed. The device
+            is required to be set to the real final device to get the best
+            load performance. Default will be removed in version >= 0.5.
```
sgugger, line 114:
```suggestion
            <Tip warning={true}>
            
            Explicitly setting the device is strongly recommended to get the best
            load performance. In version >= 0.5, not setting `device` will result in an error.
            
             </Tip> 
```
doc-builder syntax + better message.

bindings/python/src/lib.rs
sgugger, line 495:
```suggestion
                let warning = r#"Loading without specifying the device is deprecated and will error in version 0.5.0 and higher. Please use `load_file(..., device="cpu")` or `safe_open(..., device="cpu")` to make sure you want to load on "cpu". Load on the device where the weights are needed for optimal performance."#;
```

sgugger (commented):
Let some comments. As said privately, I still think we should leave the default in the API and just properly document that not setting the device hurts performance.

Narsil:
@sgugger Thanks for the review. I think I will postpone this until further notice.

https://github.com/huggingface/safetensors/pull/184
Showed that maybe loading on `cpu` first and moving to `cuda` after could be indeed faster.

I was taking opportunity of creating a new minor to put force this idea I thought was necessary at some point. But since it doesn't seem to be as straightforward conclusion as I thought, I will delay.

Narsil:
Closing until further notice.

PR #205, opened 2023-03-27
Disallowing zero-width tensor.

Narsil:
This is disabling tensors of shape `[2, 0]` which **are** valid in Pytorch and Numpy but seem pretty useless on disk (since it doesn't store any data).

It seems that such tensors are only used at runtime, either to represent empty slices, or growing tensor, but are not ever useful in actual model weights (to the best of my knowledge)

### DIFF
--- a/safetensors/src/tensor.rs
+++ b/safetensors/src/tensor.rs
@@ -911,7 +911,7 @@ mod tests {
 
     #[test]
     fn test_zero_sized_tensor() {
-        let serialized = b"<\x00\x00\x00\x00\x00\x00\x00{\"test\":{\"dtype\":\"I32\",\"shape\":[2,2],\"data_offsets\":[0, 0]}}\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00";
+        let serialized = b"<\x00\x00\x00\x00\x00\x00\x00{\"test\":{\"dtype\":\"I32\",\"shape\":[2,0],\"data_offsets\":[0, 0]}}\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00";
 
         match SafeTensors::deserialize(serialized) {
             Err(SafeTensorError::InvalidOffset(_)) => {
--- a/safetensors/src/tensor.rs
+++ b/safetensors/src/tensor.rs
@@ -458,7 +458,7 @@ impl Metadata {
         let mut start = 0;
         for (i, info) in self.tensors.iter().enumerate() {
             let (s, e) = info.data_offsets;
-            if s != start || e < s {
+            if s != start || e <= s {
                 let tensor_name = self
                     .index_map
                     .iter()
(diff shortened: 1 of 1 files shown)


sgugger:
Is it causing any security issue? Because I could imagine some users having such tensors has weights of their models (with the extra dim already there for easy broadcasting) and if I understand correctly, this change would make the deserialization of such model weights file fail.

Narsil:
> Because I could imagine some users having such tensors has weights of their models (with the extra dim already there for easy broadcasting) 

Could you expand a bit ? I thought that for models it was never really used, since zero-width contain no data, and are by definition, creating no-ops.

I'm happy to relax if there's a valid use cases for model saving (I know it's nice to have that capacity for past key values handling for instance)

sgugger:
Arf, nevermind I mixed things up. When adding a dimension for broadcasting the shape is [1, xxx] not [0, xxx]

ViktorooReps:
I use zero-width tensors to check what device my model is on. Probably not a great design choice, but works for me so far.

Anyway, had to overload `state_dict()` and `load_state_dict()` because safetensors do not support 0 dims. Is disallowing it really necessary?

ViktorooReps:
Well I use this tensor to infer device in `forward`, not during loading from disk.

I've installed safetensors from pip, so not running on main.

Here is the error I am getting:

```
tie_nerc_tests/lib/sequence_tagger/find_sequence_taggers.py:25 (TestFindSequenceTaggers.test_simple)
self = <tie_nerc_tests.lib.sequence_tagger.find_sequence_taggers.TestFindSequenceTaggers testMethod=test_simple>

    def test_simple(self):
        with TemporaryDirectory() as d:
            d = Path(d)
            for tagger_name, tagger in self.taggers.items():
                subdir = d.joinpath(tagger_name)
>               tagger.save(subdir)

find_sequence_taggers.py:31: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../../talisman-interfaces/tp_interfaces/abstract/model/model.py:67: in save
    return self.save_to_directory(path)
../../../tie_nerc/modules/sequence_tagger/abstract.py:53: in save_to_directory
    safetensors.torch.save_file(self.state_dict(), weights_path)
../../../../venv10/lib/python3.10/site-packages/safetensors/torch.py:72: in save_file
    serialize_file(_flatten(tensors), filename, metadata=metadata)
../../../../venv10/lib/python3.10/site-packages/safetensors/torch.py:237: in _flatten
    return {
../../../../venv10/lib/python3.10/site-packages/safetensors/torch.py:241: in <dictcomp>
    "data": _tobytes(v, k),
../../../../venv10/lib/python3.10/site-packages/safetensors/torch.py:209: in _tobytes
    data = np.ctypeslib.as_array(newptr, (total_bytes,))  # no internal copy
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

obj = <safetensors.torch.LP_c_ubyte object at 0x7fbc15753cc0>, shape = (0,)

    def as_array(obj, shape=None):
        """
        Create a numpy array from a ctypes array or POINTER.
    
        The numpy array shares the memory with the ctypes object.
    
        The shape parameter must be given if converting from a ctypes POINTER.
        The shape parameter is ignored if converting from a ctypes array
        """
        if isinstance(obj, ctypes._Pointer):
            # convert pointers to an array of the desired shape
            if shape is None:
                raise TypeError(
                    'as_array() requires a shape argument when called on a '
                    'pointer')
            p_arr_type = ctypes.POINTER(_ctype_ndarray(obj._type_, shape))
>           obj = ctypes.cast(obj, p_arr_type).contents
E           ValueError: NULL pointer access

../../../../venv10/lib/python3.10/site-packages/numpy/ctypeslib.py:524: ValueError
```

I know for sure it is the zero-width tensors since I dont get this error when I remove such tensors from state dict before saving.

ViktorooReps:
BTW I stole this questionable design decision from here https://stackoverflow.com/a/63477353. So I think I won't be the only safetensors user with zero-width tensors in their models.

Narsil:
If you're running on released version it means the issue is not linked to this modification, and looking at your traceback it seems the error is in the python bindings, where the buffer pointer is set to null (logical since there's no memory underneath).

I'll investigate a little to see what's going on.

It is definitely not something covered in the tests.

I'll add to the tests.

As for the stackoverflow answer it really is `next(model.parameters()).device` (Provided you know everything is on the same device).

Narsil:
Ok @ViktorooReps 

Thanks for your issue. I wasn't hard on removing this, but we've had a security audit questioning that option and we've thought disabling was a safe bet.
It seems it isn't (even though it wasn't possible to create those files on torch as you have seen)

Instead I'm going to officially support this use case.
https://github.com/huggingface/safetensors/pull/221

Despite me thinking it's not a great idea to save those, it doesn't seem have major implications so I'll go for it.

PR #208, opened 2023-03-27
Zero width shape are OK (0-rank tensors). They are simply a scalar.

Narsil:
Reasosn for support is that both torch and numpy support them, and
they flow naturally from the definition that an empty product has
length 1.


sgugger (approved):
Nice to add a test!

PR #210, opened 2023-03-28
More badges.

Narsil:


README.md
```diff
@@ -1,3 +1,15 @@
+Python
+[![Pypi](https://img.shields.io/pypi/v/safetensors.svg)](https://pypi.org/pypi/safetensors/)
+[![Documentation](https://img.shields.io/website/http/huggingface.co/docs/safetensors/index.svg?label=docs)](https://huggingface.co/docs/safetensors/index)
+[![Codecov](https://codecov.io/github/huggingface/safetensors/coverage.svg?branch=main)](https://codecov.io/gh/huggingface/safetensors)
+[![Downloads](https://static.pepy.tech/badge/safetensors/month)](https://pepy.tech/project/safetensors)
```
mishig25, line 5: shouldn't they be inside rust code that readme gets auto-generated ?

mishig25: I see `safetensors/README.tpl` file handles this issue

mishig25 (approved):
lgmt !

this is how it is looking: 
image

README.md
Narsil: There's a `README.tpl` which is used to add README specific content, because those wouldn't get rendered correctly in the rustdoc (and rustdoc already has a lot of included links)

PR #215, opened 2023-03-28
Overspecify a bit the format.

Narsil:
Adding a lot of precisions for the dark corners of the spec.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,8 +72,16 @@ with safe_open("model.safetensors", framework="pt", device="cpu") as f:
 
 Notes:
  - Duplicate keys are disallowed. Not all parsers may respect this.
+ - In general the subset of JSON is implicitly decided by `serde_json` for
+ this library. Anything obscure might be modified at a later time, that odd ways
+ to represent integer, newlines and escapes in utf-8 strings. This would only
+ be done for safety concerns
  - Tensor values are not checked against, in particular NaN and +/-Inf could
  be in the file
+ - Empty tensors (tensors with 1 dimension being 0) are not allowed.
+ They are not storing any data in the databuffer, yet retaining size in the header.
+ This might be relaxed at a later time if good use cases are found.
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -66,10 +66,23 @@ with safe_open("model.safetensors", framework="pt", device="cpu") as f:
...
+ - Duplicate keys are disallowed. Not all parsers may respect this.
+ - In general the subset of JSON is implicitly decided by `serde_json` for
+ this library. Anything obscure might be modified at a later time, that odd ways
+ to represent integer, newlines and escapes in utf-8 strings. This would only
+ be done for safety concerns
+ - Tensor values are not checked against, in particular NaN and +/-Inf could
+ be in the file
+ - Empty tensors (tensors with 1 dimension being 0) are not allowed.
+ They are not storing any data in the databuffer, yet retaining size in the header.
+ This might be relaxed at a later time if good use cases are found.
```
sgugger, line 83: As said on the other PR, I'm not sure that's a good thing.

sgugger (approved):
Thanks, LGTM!

README.md
Narsil: I'm not either. I find that disallowing earlier, and allowing when necessary is usually and easier path than disallowing later.

Storing no data is really weird IMO, and is never used for *models*.

Narsil: Let's keep discussion here: https://github.com/huggingface/safetensors/pull/206#pullrequestreview-1360999496

sgugger: It's true that in a perfectly coded model it wouldn't happen (not even sure it never happens in the whole Transformer code base TBH), just want to make sure we are not shooting ourselves in the foot by making it harder to use.

Narsil: Ok, I think it's worth disabling and allowing later if needed (I'll go with major version here anyway). Worst case we rollback reallowing, on only 1 version will be affected. (If the errors pops much later, it probably means it's a side issue)

PR #218, opened 2023-03-28
Fixing empty shape test.

Narsil:


safetensors/src/tensor.rs
```diff
@@ -914,7 +914,7 @@ mod tests {
 
     #[test]
     fn test_empty_shapes_allowed() {
-        let serialized = b"8\x00\x00\x00\x00\x00\x00\x00{\"test\":{\"dtype\":\"I32\",\"shape\":[],\"data_offsets\":[0,4]}}\x00\x00\x00\x00\x00\x00\x00\x00";
+        let serialized = b"8\x00\x00\x00\x00\x00\x00\x00{\"test\":{\"dtype\":\"I32\",\"shape\":[],\"data_offsets\":[0,4]}}\x00\x00\x00\x00";
```
mishig25, line 917: and nothing else to be changed in the rest of the test?

Narsil: No, I wrote this test before some other PR which invalidated this test, and because of the merge order it went through.

The check that made this wrong is that the buffer should necessarily match the expected size (here 4 bytes).

And I wrote it before with a buffer too long !

McPatate (approved):

PR #230, opened 2023-04-11
Fixing  the logic to determine if PR is up-to-date.

Narsil:


bindings/python/convert.py
```diff
@@ -203,14 +203,16 @@ def check_final_model(model_id: str, folder: str):
 
 
 def previous_pr(api: "HfApi", model_id: str, pr_title: str) -> Optional["Discussion"]:
+    main_commit = api.list_repo_commits(model_id)[0].commit_id
     try:
         discussions = api.get_repo_discussions(repo_id=model_id)
     except Exception:
         return None
```
Narsil, line 210:
```suggestion

    try:
        main_commit = api.list_repo_commits(model_id)[0].commit_id
        discussions = api.get_repo_discussions(repo_id=model_id)
    except Exception:
        return None
```

PR #232, opened 2023-04-11
Fixing shared detection logic.

Narsil:


bindings/python/tests/test_pt_comparison.py
```diff
@@ -161,6 +161,18 @@ def test_cannot_serialize_a_non_contiguous_tensor(self):
         with self.assertRaises(ValueError):
             save_file(data, "./tests/data/out.safetensors")
 
+    def test_cannot_serialize_shared(self):
+        A = torch.arange(6, dtype=torch.float32).reshape((2, 3))
+        B = A[:1]
+        data = {"A": A, "B": B}
+        with self.assertRaises(RuntimeError):
+            save_file(data, "./tests/data/out.safetensors")
```
mishig25, line 169: what kind of runtime error message gets generated? Basically, does the error message clearly convey the problem?

mishig25: Also, does this weights "shared" feature only exist on torch? Do you need to handle for tf, jax?

PR #233, opened 2023-04-12
Committing Cargo.lock.

### LINKED ISSUE #229: Use cargo.lock to make the project reproducible
Building the project at two different times of day may produce very different results, because the project is not storing a lock file, and it is in the `.gitignore`. It would be helpful for projects like my own which attempt to make incredibly complex generative AI software more reproducible (https://nixified.ai) if you did this.

Narsil:
Fix #229


.gitignore
```diff
@@ -1,5 +1,6 @@
 safetensors/target
-Cargo.lock
+safetensors/Cargo.lock
+safetensors/fuzz/Cargo.lock
```
mishig25, line 3: Could we just combine them into one?
```suggestion
safetensors/**/Cargo.lock
```

Narsil: Thanks !

mishig25 (approved):
lgtm!

PR #235, opened 2023-04-20
Fixing alignment

### LINKED ISSUE #231: buffer not aligned
I converted the pytorch checkpoint to safetensors. The buffer is not aligned

RWKV-4-Pile-430M-20220808-8066.pth is from https://huggingface.co/BlinkDL/rwkv-4-pile-430m
The convert script is here: https://github.com/iacore/rwkv-np/blob/main/convert.py

```
> xxd RWKV-4-Pile-430M-20220808-8066.safetensors | head -n 2
00000000: 66a7 0000 0000 0000 7b22 626c 6f63 6b73  f.......{"blocks
00000010: 2e30 2e61 7474 2e6b 6579 2e77 6569 6768  .0.att.key.weigh
```

The tensor data are all f32.

0xa766 % 4 == 2

Why not aligned: the offset count from after the metadata header (sized 0xa766).

Narsil:
Alignment was broken.

Simple logic failure (with pretty bad consequences):

- We're calculating how "unaligned" the header is. But we were adding the wrong padding to fix the alignment.
- Adding some test in the proptest to make sure it's aligned after reload
- Update the other tests

Fix #231

### DIFF
--- a/safetensors/src/tensor.rs
+++ b/safetensors/src/tensor.rs
@@ -745,6 +745,7 @@ mod tests {
             let before = SafeTensors { metadata, data: &data };
             let tensors = before.tensors();
             let bytes = serialize(tensors.iter().map(|(name, view)| (name.to_string(), view)), &None).unwrap();
+
             let after = SafeTensors::deserialize(&bytes).unwrap();
 
             // Check that the tensors are the same after deserialization.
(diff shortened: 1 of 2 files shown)


safetensors/src/tensor.rs
```diff
@@ -194,7 +194,9 @@ fn prepare<S: AsRef<str> + Ord + std::fmt::Display, V: View, I: IntoIterator<Ite
     let mut metadata_buf = serde_json::to_string(&metadata)?.into_bytes();
     // Force alignment
     let extra = metadata_buf.len() % 8;
-    metadata_buf.extend(vec![b' '; extra]);
```
iacore, line 197: :frowning:

Narsil: Yes, bad blunder :(

OlivierDehaene (approved):

PR #236, opened 2023-04-21
Adding `save_model` and `load_model` to help with shared tensors with PyTorch.

Narsil:
This PR adds new surface which is PyTorch specific.
The idea is that allowing shared tensors on files opens a lot of issues, 
but at the same time, they are indeed quite used within torch ecosystem and
not trivial to workaround when not familiar with torch internals (or understand
the implications).

This PR adds 2 new functions:

`load_model` and `save_model` which should really be as convenient to use
as `torch.save(model.state_dict(), filename)` and `model.load_state_dict(torch.load(filename))`.

It also adds some documentation explaining the situation and various choices.

Some caveats:

- The functions will fail if user is attempting to save buffer which are unrecoverable,
  meaning trying to save a slice of a large shared tensor, where no reference points to 
  the entire tensor.
- If you load shared tensors on **not** shared model, torch will copy the weights
  to satisfy the non shared property. We will drop the shared weights on save
  meaning the missing weights will yell on load (you absolutely need to save
  tensor copies on disk if you want to do the same)
- If you load **not** shared tensors on shared model, PyTorch will arbitrarily
  choose one version to put on your model without alarming you that some weights
  were obfuscated. These function will explain that some keys were not expected
  on the file.

### DIFF
--- a/bindings/python/py_src/safetensors/torch.py
+++ b/bindings/python/py_src/safetensors/torch.py
@@ -23,9 +23,9 @@ def _is_complete(tensor: torch.Tensor) -> bool:
     return tensor.data_ptr() == storage.data_ptr() and tensor.nelement() * _SIZE[tensor.dtype] == storage.nbytes()
 
 
-def _remove_duplicate_names(state_dict: Dict[str, torch.Tensor]) -> List[str]:
+def _remove_duplicate_names(state_dict: Dict[str, torch.Tensor]) -> Dict[str, List[str]]:
     shareds = _find_shared_tensors(state_dict)
-    to_remove = []
+    to_remove = defaultdict(list)
     for shared in shareds:
         complete_names = [name for name in shared if _is_complete(state_dict[name])]
         if not complete_names:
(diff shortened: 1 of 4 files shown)


bindings/python/py_src/safetensors/torch.py
```diff
@@ -1,13 +1,126 @@
...
+    if missing or unexpected:
+        missing_keys = ", ".join([f'"{k}"' for k in sorted(missing)])
+        unexpected_keys = ", ".join([f'"{k}"' for k in sorted(unexpected)])
+        error = f"Error(s) in loading state_dict for {model.__class__.__name__}:"
+        if missing:
+            error += f"\n    Missing key(s) in state_dict: {missing_keys}"
+        if unexpected:
+            error += f"\n    Unexpected key(s) in state_dict: {unexpected_keys}"
+
+        raise RuntimeError(error)
```
sgugger, line 121: Having a non-strict mode here could be useful if you are loading part of a checkpoint in a model (for instance if adding a new randomly initialized head to a pretrained model).

sgugger (approved):
Thanks for investigating this! I think it could unlock a lot of usage in the PyTorch community to at least support simple use cases of shared weights like this.
I completely agree with the choice of erroring when we don't have a full tensor covering all shared copies (which is way better than having surprise checkpoints 5 times bigger in PyTorch).

bindings/python/py_src/safetensors/torch.py
Narsil: Great suggestion ! I added it.

Narsil:
> Thanks for investigating this! I think it could unlock a lot of usage in the PyTorch community to at least support simple use cases of shared weights like this.

Thanks ! I think this approach is better than doing nothing obviously, but also supporting shared tensors (with all the drawbacks mentioned in the docs)

I need to think a bit more about contiguous tensors here too.

PR #246, opened 2023-05-17
There should be `.dev0` while on `main`

mishig25:


Narsil:
Yup let's do it, I'll try not to forget to update everything on next release.

Ignore the failing test it's on my end

PR #252, opened 2023-05-22
Added method `len` on `SafeTensors` 

mfuntowicz:
This avoid allocating a new tensor through `names()` or `tensors()`


Narsil (approved):
LGTM as they do skip allocs.

`is_empty` is recommended over `len() == 0` for perf reasons too.

PR #253, opened 2023-05-22
Fixing the spec with mention about the buffer which cannot be non-contiguous.

Narsil:


mishig25 (approved):
lgtm !

9999years:
Why was the `Cargo.lock` removed???

PR #259, opened 2023-05-29
Add CMake project to generated C API binding

mfuntowicz:
This PR continues the initial work around adding a lower-level, C-API binding to the safetensors project.

It builds upon the `binding/c` folder with the following additions:

- CMake based build to ease integration
- Add `include/safetensors/safetensors.h` include folder
- Add `catch2` as external third-party library managed through git submodule (requires `git clone --fetch-submodule`)
- Add unittests to ensure the API remains compatible through CTests + Catch2.

Potential improvements:
- Removed `TensorView` shape copy enforcement when creating the struct through `Cow`


bindings/c/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "c"
+version = "0.3.0"
```
Narsil, line 3:
```suggestion
version = "0.3.1"
```

bindings/c/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+[lib]
+name = "safetensors"
+crate-type = ["staticlib"]
```
Narsil, line 9: Any specific reason to build static libs instead of a shared one ? (Ideally I imagine user could switch easily if they wanted). I think shared is a saner default probably, no ?

Narsil:
As discussed internally (and some stuff I added afterwards):

- Removed the submodule usage and used Cmake versioning instead.
- Reverted the Cow (it's an optimization, which doesn't belong here, it could be a separate PR, I have personal doubts it makes any significant difference, for the API simplicity's cost `Vec` being more used than `Cow`, but it's indeed debatable). In any case it belongs in a different PR.
- Fixed a bug where the `shape` was passed as a pointer without being owned first, leading the `release` version to crash because c client could access uninitialized memory.
- Added `start` and `stop` to get the buffer offsets (which are necessary to get the buffer data from the file).
- Added an example to describe a given `model.safetensors` file.
- Forced Release mode (to prevent people using debug inadvertently).
- Moved to dynamic lib by default. *Tests on windows still somehow need static lib, we suppose it's a bug in Cache2 or corrosion in the linking process since everything works properly with proper paths.*

RobinQu:
Is C++ binding support terminated?

jingzhao-salient:
Exactly what I am looking for. Cannot wait to see this PR getting merged. 👍

bindings/c/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
...
+
+# Build up Rust part
+include(FetchContent)
+
+FetchContent_Declare(
+    Corrosion
+    GIT_REPOSITORY https://github.com/corrosion-rs/corrosion.git
+    GIT_TAG v0.4 # Optionally specify a commit hash, version tag or branch here
+)
+FetchContent_MakeAvailable(Corrosion)
```
SomeoneSerge, line 29: Seeing that Corrosion [installs find_package targets](https://github.com/corrosion-rs/corrosion/blob/123be1e3d8170c86e121392e8bffa4def7dc3447/CMakeLists.txt#L67-L73), it would be nice to add explicit `FIND_PACKAGE_ARGS` (helpful for distributions)

AlpinDale:
Any updates on this? This would be very nice to have.

PR #264, opened 2023-05-31
Make tensor identifier a bit more robust, and transformers compatible

thomasw21:
`transformers`: https://github.com/huggingface/transformers/pull/23871

I think `nbytes` is required otherwise you might be having multiple storage starting from the same offset but not the same length considered as duplicate, which I'm guessing is not something you support. Granted it's a weird case.

### DIFF
--- a/bindings/python/py_src/safetensors/torch.py
+++ b/bindings/python/py_src/safetensors/torch.py
@@ -38,7 +38,7 @@ def _find_shared_tensors(state_dict: Dict[str, torch.Tensor]) -> List[Set[str]]:
     for k, v in state_dict.items():
         if v.device != torch.device("meta"):
             # Need to add device as key because of multiple GPU.
-            tensors[(storage_ptr(v), v.device)].add(k)
+            tensors[(v.device, storage_ptr(v), storage_size(v))].add(k)
     tensors = list(sorted(tensors.values()))
     return tensors
 


Narsil:
> I think nbytes is required otherwise you might be having multiple storage starting from the same offset but not the same length considered as duplicate, which I'm guessing is not something you support. Granted it's a weird case.

Correct me if I'm wrong, but if both storage start at the same pointer, they are necessarily shared (it's the same address location).

If the storage are not of the same size, but point to the same location, it would be super odd, no ? Either my understanding of storage in PyTorch is erroneous, or sounds like a bug in PT.

If that's true it means the logic to find the maximal overlapping buffer might be not working as expected.
Any way to reproduce this ?

thomasw21:
> Correct me if I'm wrong, but if both storage start at the same pointer, they are necessarily shared (it's the same address location).

So essentially tensor storage is basically a 1d contiguous array, that are sufficient to express the tensor that are built on top. Since it's a 1d array then, the only way to correctly identify is by passing both start and size (or start and end).

> If the storage are not of the same size, but point to the same location, it would be super odd, no ? Either my understanding of storage in PyTorch is erroneous, or sounds like a bug in PT.

Nothing from PT verifies that two tensors don't have overlapping storage. In fact in a lot of cases that's quite nice since `view`/`slice` etc are pure metadata computation, while keeping the storage the same. Though in those cases they keep the entire storage, you can build custom operations to set the storage yourself.

> If that's true it means the logic to find the maximal overlapping buffer might be not working as expected.
Any way to reproduce this ?

Yeah so this is really hard. In the general case you need to compute the union of everyone. I added a test that creates such a weird system where `data_ptr` match but not the storage size. (so just saving the first value storage is not sufficient)

Narsil:
We discussed this extensively internally, just sharing here for posterity.

Storage faking like done in the test *will* break things in torch itself. As it since the storage are different for torch, it will not survive `torch.load(torch.save())`. Both storages will be stored in different places on disk (extra storage on disk), and be loaded onto different storages afterwards.

The current PR will resume that behavior (with the difference compared to torch if someone creates a fully overlapping storage we detect it as the same, PT doesn't).

This is a *very* edge case issue, because PT will just do weird things all the time when doing that. And since PT itself doesn't maintain a hierarchy of storages, it has no way of knowing of the tensors are actually shared or not.

The only way we're aware of detecting such "shared" storages, would be to actually look for all pointer + storage looking for overlap, looking for the maximal range storages ([0; 2] and [1; 3] do not admit a maximum storage which we could save, therefore we do not have a valid tensor over [0; 3] which could act as a proxy). This is a lot of work + runtime costs and oddly breaking torch behavior, therefore it's out of scope for now.

PR #266, opened 2023-06-05
Adding error checking, convertion can fail on schedulers.

Narsil:
Original issue:

https://huggingface.co/spaces/safetensors/convert/discussions/8#6478327dbb7681ad671719d3

Solution:

Just ignore the file and report the failure.

### DIFF
--- a/bindings/python/convert.py
+++ b/bindings/python/convert.py
@@ -5,7 +5,7 @@ import shutil
 from collections import defaultdict
 from inspect import signature
 from tempfile import TemporaryDirectory
-from typing import Dict, List, Optional, Set
+from typing import Dict, List, Optional, Set, Tuple
 
 import torch
 
@@ -32,6 +32,8 @@ If you find any issues: please report here: https://huggingface.co/spaces/safete
 
 Feel free to ignore this PR.
 """
+
+ConversionResult = Tuple[List["CommitOperationAdd"], List[Tuple[str, "Exception"]]]
 
 
 class AlreadyExists(Exception):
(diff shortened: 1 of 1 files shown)


bindings/python/convert.py
```diff
@@ -95,18 +97,20 @@ def convert_multi(model_id: str, folder: str) -> List["CommitOperationAdd"]:
     operations = [
         CommitOperationAdd(path_in_repo=local.split("/")[-1], path_or_fileobj=local) for local in local_filenames
     ]
+    errors: List[Tuple[str, "Exception"]] = []
```
sgugger, line 100: It's Python, no need for the type hint here, especially since the name is descriptive enough :-)

bindings/python/convert.py
```diff
@@ -95,18 +97,20 @@ def convert_multi(model_id: str, folder: str) -> List["CommitOperationAdd"]:
...
-def convert_single(model_id: str, folder: str) -> List["CommitOperationAdd"]:
+def convert_single(model_id: str, folder: str) -> ConversionResult:
     pt_filename = hf_hub_download(repo_id=model_id, filename="pytorch_model.bin")
 
     sf_name = "model.safetensors"
     sf_filename = os.path.join(folder, sf_name)
     convert_file(pt_filename, sf_filename)
     operations = [CommitOperationAdd(path_in_repo=sf_name, path_or_fileobj=sf_filename)]
-    return operations
+    errors: List[Tuple[str, "Exception"]] = []
```
sgugger, line 112: Same here

sgugger (approved):
Thanks for adding this!

bindings/python/convert.py
Narsil: Ahah, I have mypy setup here, and it's not smart enough to figure it out on its own (and does yell if type is not present).

bindings/python/convert.py
```diff
@@ -320,6 +328,6 @@ def convert(api: "HfApi", model_id: str, force: bool = False) -> Optional["Commi
             " Continue [Y/n] ?"
         )
     if txt.lower() in {"", "y"}:
-        convert(api, model_id, force=args.force)
+        _commit_info, _errors = convert(api, model_id, force=args.force)
```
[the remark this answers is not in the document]
Narsil: I added the same print as the spaces.

PR #284, opened 2023-06-29
Fix wrong docstring for SafeTensors' names method (#250)

### LINKED ISSUE #250: Small Typo in Rust Docs
See here which should explain the names() method: https://github.com/huggingface/safetensors/blob/main/safetensors/src/tensor.rs#L377

But the docs instead repeats the explanation for `tensor method`.

If you want,, I can provid quick PR.

dirac3000:
# What does this PR do?

Just correct the docstring of the SafeTensors' names method.

Fixes #250

### DIFF
--- a/safetensors/src/tensor.rs
+++ b/safetensors/src/tensor.rs
@@ -374,9 +374,9 @@ impl<'data> SafeTensors<'data> {
         }
     }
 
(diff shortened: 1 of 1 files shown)


Narsil (approved):
LGTM

PR #293, opened 2023-07-07
Autodoc for save_model and load_model

wouterzwerink:
# What does this PR do?

Adds load_model and save_model to the autodoc. These are useful functions that are currently only documented under the shared tensors page.
These functions are also useful outside of dealing with shared tensors and probably should be documented on the torch page.


Narsil:
Thank you !

PR #296, opened 2023-07-11
Support musicgen conversion.

Narsil:
# What does this PR do?

### DIFF
--- a/bindings/python/convert.py
+++ b/bindings/python/convert.py
@@ -161,8 +161,11 @@ def check_final_model(model_id: str, folder: str):
     shutil.copy(config, os.path.join(folder, "config.json"))
     config = AutoConfig.from_pretrained(folder)
 
-    _, (pt_model, pt_infos) = infer_framework_load_model(model_id, config, output_loading_info=True)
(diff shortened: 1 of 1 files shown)


bindings/python/convert.py
```diff
@@ -161,8 +161,11 @@ def check_final_model(model_id: str, folder: str):
     shutil.copy(config, os.path.join(folder, "config.json"))
     config = AutoConfig.from_pretrained(folder)
 
-    _, (pt_model, pt_infos) = infer_framework_load_model(model_id, config, output_loading_info=True)
-    _, (sf_model, sf_infos) = infer_framework_load_model(folder, config, output_loading_info=True)
+    import transformers
+
+    class_ = getattr(transformers, config.architectures[0])
+    (pt_model, pt_infos) = class_.from_pretrained(folder, output_loading_info=True)
+    (sf_model, sf_infos) = class_.from_pretrained(folder, output_loading_info=True)
```
Narsil, line 168: The old version doesn't work anymore.

This might fail on very old models (where config.architectures is not written), but I think it's ok now, they should be converted by now.

sgugger (approved):
Works for me, thanks!

PR #298, opened 2023-07-12
[doc] add missing metadata doc 

stas00:
`transformers` requires `metadata` - so update the example to show how it's added.


Narsil:
It is not required. `safetensors` does not depend on `transformers`.

`transformers` requiring it is a `transformers` thing (and totally valid at that).

Issue #291, opened 2023-07-05
Writing files > RAM


wouterzwerink:
### Feature request

A (documented) method for writing files with more tensor data than can be fit into RAM

### Motivation

The documented method for writing files is to pass a dict of tensors. This requires having all the tensors in RAM.
However, it should be possible to write safetensors files larger than RAM by either adding tensors to existing files (without reading the files into memory).

That will enable some code like (function names are just to explain, could be different):
```python
safetensors.extend_file(tensors, filename)

# Alternatively, writing in chunks
with safetensors.create_file(filename) as s_file:
    for i in range(len(indices)):
         # Only one tensor loaded in RAM at a time
         s_file.write(create_tensor(i))
```

### Your contribution

Won't be able to work on this myself for a while. If this doesn't get picked up, I'll try implement this in a few weeks

Narsil:
It's trickier than you think.

We need ALL metadata to be able to write the header. Since the data comes *after* the metadata, you cannot write the first tensors before having the entire header.

Making the API for such a thing relatively confusing (also lots of sanity checking because you need to write the proper tensors at every proper location, and not miss tensors for your final file to be correct.

Is there any reason you cannot just split the tensors over multiple files ? (This is what we do in distributed training for instance).
Fusing existing on disk files is much much simpler (just take all headers, add offsets whereever they might need to be, and copy the buffers next to each other in the new file (for which you can require as low ram as you want).

akx:
Well, you can in fact stream a safetensors file to disk so long as you have a good guesstimate of the length of your header and reserve space for it in advance. You'll just need to pad the JSON with whitespace at the end; `serde` chokes on null bytes, but not spaces...

```python
import json
import random
import struct

import safetensors.torch

header_size = 65536
data_start_offset = header_size + 8
header = {}
filename = "test.safetensors"
with open(filename, "wb+") as f:
    f.seek(data_start_offset)
    for x in range(3):
        data = random.randbytes(4 * 4)
        start = f.tell() - data_start_offset
        header[f"tensor-{x}"] = {"dtype": "U8", "shape": [4, 4], "data_offsets": [start, start + len(data)]}
        f.write(data)
    f.seek(0)
    f.write(struct.pack("<Q", header_size))
    header_bytes = json.dumps(header).encode()
    assert len(header_bytes) < header_size
    f.write(header_bytes.ljust(header_size))
print(safetensors.torch.load_file(filename))
```
happily prints out
```
[...]

PR #306, opened 2023-08-01
Clarify specification around endianness and header padding

akx:
# What does this PR do?

This PR clarifies the readme around the spec a little; the endianness of the header size was not specified.

Further, as discussed in https://github.com/huggingface/safetensors/issues/291#issuecomment-1660365431 `serde_json` (and in fact any spec-compliant JSON parser, I suppose) allows padding the JSON data with whitespace characters, so note that – and add a test for that, too.

**EDIT:** As discussed in the comments below, this also nails down the spec that the padding may only be trailing.

---

As an aside: it's not very clear for a contributor that you shouldn't modify `README.md` directly... 😁

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ with safe_open("model.safetensors", framework="pt", device="cpu") as f:
 
 ### Format
 
-- 8 bytes: `N`, a u64 int, containing the size of the header
+- 8 bytes: `N`, an unsigned little-endian 64-bit integer, containing the size of the header
 - N bytes: a JSON utf-8 string representing the header.
   - The header is a dict like `{"TENSOR_NAME": {"dtype": "F16", "shape": [1, 16, 256], "data_offsets": [BEGIN, END]}, "NEXT_TENSOR_NAME": {...}, ...}`, where offsets point to the tensor data relative to the beginning of the byte buffer, with `BEGIN` as the starting offset and `END` as the one-past offset (so total tensor byte size = `END - BEGIN`).
   - A special key `__metadata__` is allowed to contain free form string-to-string map. Arbitrary JSON is not allowed, all values must be strings.
--- a/README.md
+++ b/README.md
@@ -76,8 +76,12 @@ with safe_open("model.safetensors", framework="pt", device="cpu") as f:
 ### Format
 
 - 8 bytes: `N`, an unsigned little-endian 64-bit integer, containing the size of the header
-- N bytes: a JSON utf-8 string representing the header.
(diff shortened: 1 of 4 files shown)


Narsil:
Hi @akx ,

I'm very fine adding a test, and the endianness precision.
I don't feel easy adding the whitespace to the SPEC. the spec says VALID JSON which should imo be plenty.
The additional notes do mention that there is no one single valid JSON spec, so weird tricks like this are not really defined on purpose (so we can revoke clearly ill defined JSON that cause actual security issues).
The space padding seems ok in the long term but it doesn't really suit my notion of "obviously valid json".
So adding in the notes feels better instead.

Currently the space padding is used to get alignment and is a "trick" and implementation detail, not something that should ever be required.
Mentioning in the docs that this particular implementation uses that is fine in my book ! I would keep only right hand side padding maybe (despite many more options being available as you mention) just so that implementors try to stick with that.

Are you ok with the proposed change about padding ?

akx:
I'm totally okay with specifying that only trailing whitespace (or null, but that's presently not accepted by serde-json) padding is allowed, but that doesn't reflect the reality of this current canonical implementation (which evidently allows leading whitespace padding).

Whether it's okay that this implementation is more lenient than the spec is up to you - if we do hammer it down in the spec that only trailing spacing is allowed, I would suggest we add a check that the first byte of the header is `{` in this implementation too, to make it less probable that leading-padded safetensors files begin appearing in the wild.

Ps. I'm just boarding a plane so I'm not able to reply immediately 😁

Narsil:
>  null, but that's presently not accepted by serde-json
Educated guess but this prevents polyglot files because all C-strings are null terminated.

> we do hammer it down in the spec that only trailing spacing is allowed,
>  I would suggest we add a check that the first byte of the header is { i

That's actually a neat idea.

akx:
> Educated guess but this prevents polyglot files because all C-strings are null terminated.

Or maybe just that NUL isn't at all in the JSON spec whereas the documented ws characters are, and it's easier for a parser to just consume them when outside a string literal regardless of other parser state :)

akx:
@Narsil Okiedoke, I amended the second commit to

* write the spec more clearly
* add the constraint that the first character of the decoded UTF-8 string must be `{` (i.e. leading padding is disallowed)

WDYT?

Narsil (commented):
LGTM just the nit of removing other paddings in the spec.

akx:
@Narsil Rebased + ran `make`.

Narsil:
I'm making a release without this in. So we can have the deepspeed fix (DS shares tensors without actually shared tensors).

I will then go ahead and merge this so and advertise to other safetensors implementors see if what we're doing here is actually break free for them. (Which it should).

PR #309, opened 2023-08-05
feat: Add filter for shared storage non overlapping tensors

Narsil:
# What does this PR do?

Deepspeed does some tensor sharing but not actually overlapping
tensors. (It's not exactly clear why).

This PR fixes it by adding a global check for all shared tensors if the
actual offsets are overlapping or not.
It looks at all (start, stop) pointers of the tensors, sorts them by start pointer, and decides 2 tensors are disjoint if stop(previous) <= start(next)

Linked issues:

- https://github.com/huggingface/text-generation-inference/issues/742
- https://github.com/huggingface/text-generation-inference/issues/701
- https://github.com/huggingface/text-generation-inference/issues/622
- https://github.com/huggingface/diffusers/issues/4361

This will require a new release after merge to fix user's issues.

### DIFF
--- a/bindings/python/py_src/safetensors/torch.py
+++ b/bindings/python/py_src/safetensors/torch.py
@@ -19,6 +19,7 @@ def storage_ptr(tensor: torch.Tensor) -> int:
             # Fallback for meta storage
             return 0
 
+
 def _end_ptr(tensor: torch.Tensor) -> int:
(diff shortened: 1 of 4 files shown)


Narsil:
@thomasw21

sgugger (approved):
Thanks for adding!

PR #318, opened 2023-08-09
Add support to load model within torch._subclasses.FakeTensorMode

thiagocrepaldi:
# What does this PR do?

Currently, when a user tries to load a model under Pytorch's FakeTensorMode, `result[k] = f.get_tensor(k)` fails with a `ValueError` that says `could not determine the shape of object type 'torch.storage.UntypedStorage'`

This happens because `FakeTensor` does not have actual data to be loaded

This PR tries to detect a fake mode and when one is found, an empty tensor is created with the original shape and type - no data is actually loaded

Repro:

```python
from torch._subclasses.fake_tensor import FakeTensorMode
from transformers import AutoModel

model_name = "bigscience/bloom-560m"
fake_mode = FakeTensorMode()
with fake_mode:
    model = AutoModel.from_pretrained(model_name)
```

Error:

```bash
[2023-08-09 17:20:39,037] [INFO] [real_accelerator.py:110:get_accelerator] Setting ds_accelerator to cuda (auto detect)
  File "/opt/pytorch/repro_bloom.py", line 12, in <module>
    model = AutoModel.from_pretrained(model_name)
  File "/opt/conda/envs/ptca/lib/python3.8/site-packages/transformers/models/auto/auto_factory.py", line 493, in from_pretrained
    return model_class.from_pretrained(
  File "/opt/conda/envs/ptca/lib/python3.8/site-packages/transformers/modeling_utils.py", line 2629, in from_pretrained
    state_dict = load_state_dict(resolved_archive_file)
  File "/opt/conda/envs/ptca/lib/python3.8/site-packages/transformers/modeling_utils.py", line 458, in load_state_dict
    return safe_load_file(checkpoint_file)
  File "/opt/safetensors/bindings/python/py_src/safetensors/torch.py", line 319, in load_file
    result[k] = f.get_tensor(k)
ValueError: could not determine the shape of object type 'torch.storage.UntypedStorage'
```

### DIFF
--- a/bindings/python/py_src/safetensors/torch.py
+++ b/bindings/python/py_src/safetensors/torch.py
@@ -7,6 +7,13 @@ import torch
 
 from safetensors import deserialize, safe_open, serialize, serialize_file
 
+try:
+    from torch._guards import detect_fake_mode
+except Exception:
+    # Pytorch older than 2.x does not support FakeTensorMode
+    def detect_fake_mode():
+        return False
+
 
 def storage_ptr(tensor: torch.Tensor) -> int:
     try:
@@ -308,7 +315,13 @@ def load_file(filename: Union[str, os.PathLike], device="cpu") -> Dict[str, torc
     result = {}
     with safe_open(filename, framework="pt", device=device) as f:
         for k in f.keys():
-            result[k] = f.get_tensor(k)
+            fake_mode = detect_fake_mode()
+            if not fake_mode:
+                result[k] = f.get_tensor(k)
+            else:
+                empty_tensor = f.get_slice(k)
+                result[k] = torch.empty(tuple(empty_tensor.get_shape()),
+                                        dtype=_getdtype(empty_tensor.get_dtype()))
     return result
 
 


Narsil:
I'd rather not import anything from private locations in `torch`.

Also this seems like something which should be done in user code.
Under the hood `safetensors`  really just calls pure torch code. https://gist.github.com/Narsil/3edeec2669a5e94e4707aa0f901d2282

thiagocrepaldi:
> I'd rather not import anything from private locations in `torch`.
> 
> Also this seems like something which should be done in user code. Under the hood `safetensors` really just calls pure torch code. https://gist.github.com/Narsil/3edeec2669a5e94e4707aa0f901d2282

Thank you for the feedback. I am looking into a more elegant way as tracked by https://github.com/pytorch/pytorch/issues/106732. Maybe the Torch Function approach help us as ezyang suggested.

The issue with patching `load_file` in user code is that it can be called within different modules and we would need to patch different namespaces as we discover them by accident. For example, `safetensors.torch.load_file` is also aliased to `transformers.modeling_utils.safe_load_file` through `from safetensors.torch import load_file as safe_load_file` at `transformers.modeling_utils`

The private location issue is kind of by-design for `FakeTensorMode` (and most of PyTorch Dynamo actually). They are kind of new features for PyTorch 2.x series.  Only now that torch.export will become public, for instance (https://github.com/pytorch/pytorch/pull/106904)

Narsil:
Thanks, I'll track the Pytorch thread.

Not importing from private locations is because I don't like having to do mess with version parsing (which torch can modify at will perfectly understandable).

When I meant user code, I meant since FakeTensorMode still seems beta/private, all hacks to make it work should remain private/beta, no ?

PR #329, opened 2023-08-17
impl `View` for `TensorView`

coreylowman:
# What does this PR do?

The existing impl only exists for `&TensorView`, which means you cannot pass a `Vec<(String, TensorView)>` into `serialize_to_file`. With this new impl, you can.


Narsil:
LGTM

PR #332, opened 2023-08-20
Add NumPy for all backends dependencies

chainyo:
# What does this PR do?

I noticed that all backends require `numpy` as a dependency when you use the serialization and deserialization methods from `safetensors`.
I propose to add them as an extra dependency to `torch`, `tensorflow`, `paddlepaddle` and `flax`.

Wdyt? @Narsil


Narsil:
Thanks.

Issue #340, opened 2023-08-23
Windows 64-bit wheel is missing for 0.3.3


henk717:
### System Info

Python 3.8, Windows 64-bit, Safetensors 0.3.3

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Reproduction

Install safetensors using pip install safetensors in a new python environment, doing this you get this installation failure : error: can't find Rust compiler

### Expected behavior

The rust compiler is either being handled automatically or not needed because of prebuilt wheels for this platform.

Narsil:
It's fixed. I messed up the windows 64 bit when rewriting the action. 

Should be fixed now, can you confirm ?

henk717:
Can confirm, I see the wheels on pypi and was able to install it on a machine that did not have rust installed.
I have resumed the rollout to the KoboldAI users, thanks!

PR #341, opened 2023-08-23
Fix building Windows x64 wheels

jllllll:
This PR attempts to fix building Windows x64 wheels by changing the `python-release.yml` workflow.
Currently, x64 wheels are missing for Windows as the current workflow is only generating x86 wheels.

I am unfamiliar with your particular workflow needs, so I can only hope that this fixes the issue.
If it does not, I hope this can give you some ideas on how to fix it.

See this relevant issue: https://github.com/huggingface/safetensors/issues/340

### DIFF
--- a/.github/workflows/python-release.yml
+++ b/.github/workflows/python-release.yml
@@ -48,31 +48,24 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v2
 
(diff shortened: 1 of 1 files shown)


Narsil:
Thanks for the PR I've done it a bit differently.

PR #345, opened 2023-08-25
[docs] Convert weights

stevhliu:
From internal Slack [thread](https://huggingface.slack.com/archives/C027NLU6CE9/p1692804031143159), this PR adds a short guide for converting weights to `.safetensors`. Let me know if I'm missing anything!

### DIFF
--- a/docs/source/convert-weights.md
+++ b/docs/source/convert-weights.md
@@ -1,0 +1,23 @@
+# Convert weights to safetensors
+
+PyTorch model weights are commonly saved and stored as `.bin` files with Python's [`pickle`](https://docs.python.org/3/library/pickle.html) utility. To save and store your model weights in the more secure `safetensor` format, we recommend converting your weights to `.safetensors`.
+
+The easiest way to convert your model weights is to use the [Convert Space](https://huggingface.co/spaces/diffusers/convert), given your model weights are already stored on the Hub. The Convert Space downloads the pickled weights, converts them, and opens a Pull Request to upload the newly converted `.safetensors` file to your repository. Merge the Pull Request to upload the weights, but if you can't wait to try it out, you can also use the `.safetensors` immediately by specifying the reference to the Pull Request in the revision parameter:
+
+```py
+from transformers import AutoModel
+
+model = AutoModel.from_pretrained(
+    "my-safe-model", revision="refs/pr/1", use_safetensors=True
+)
+```
+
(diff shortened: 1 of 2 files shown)


docs/source/convert-weights.md
```diff
@@ -0,0 +1,23 @@
...
+
+Another way to convert your `.bin` files is to use the [`~safetensors.torch.save_model`] function:
+
+```py
+from transformers import AutoModel
+from safetensors.torch import save_model
+
+unsafe_model = AutoModel.from_pretrained("my-unsafe-model")
+save_model(unsafe_model, "model.safetensors")
+```
```
Narsil, line 23:
```suggestion
```

I'm not a big fan of suggesting either way here.

1. AutoModel will miss the model's head leading to bad models being saved and we cannot invent the correct head for users. Also it requires `transformers` which safetensors doesn't require.
2. This should work and is the preferred way for doing it when not using transformers, however, it might cause issues with transformers models because of weight naming (some weights share the same name, safetensors will discard any and everything will work fine, but that might be breaking the names transformers expects to see and clash with their own internals).

Using the space should be the go-to IMO.

Narsil: `bindings/python/convert.py` is basically the code run on the space too.

Narsil: I could remove the logit sanity check, it should work for more models. Still going to be dead slow..

If we could at least add progressbars that'd help but I'm not sure how to actually do that.

Narsil: @christophe-rannou @co42 Any way we could bump/replicate this space to have more replicas ?
It's chugging pretty much everything it can atm

stevhliu: > This should work and is the preferred way for doing it when not using transformers, however, it might cause issues with transformers models because of weight naming (some weights share the same name, safetensors will discard any and everything will work fine, but that might be breaking the names transformers expects to see and clash with their own internals).

Maybe we can add a warning here to let users know they might run into issues if they use the `save_model` method with `transformers` and they can try it on their own if they want?

Otherwise, we can just keep the Convert Space option and have a warning that says it may be slower for larger models + point them towards `bindings/python/convert.py` if they want to convert it locally.

stevhliu: Went with option #2 to only recommend the Convert Space; let me know if this is ok for you @Narsil! 🙂

Narsil: LGTM thanks for the ping.

PR #349, opened 2023-08-25
Fix id_tensor_storage for XLA tensors.

qihqi:
XLA tensors don't have storage and attempting to get it's storage will get

  RuntimeError: Attempted to access the data pointer on an
                invalid python storage.

Repro:
```python
from transformers import pytorch_utils
import torch_xla.core.xla_model as xm
device = xm.xla_device()
a = torch.ones((10,10)).to(device)
pytorch_utils.id_tensor_storage(a)  # raises RuntimeError
```
With this patch it would print out
(device(type='xla', index=0), 1, 400).

### DIFF
--- a/bindings/python/py_src/safetensors/torch.py
+++ b/bindings/python/py_src/safetensors/torch.py
@@ -12,6 +12,11 @@ def storage_ptr(tensor: torch.Tensor) -> int:
     try:
         return tensor.untyped_storage().data_ptr()
     except Exception:
+        if tensor.device.type == 'xla':
+            # this is a XLA tensor, it must be created using torch_xla's
+            # device. So the following import is safe:
+            import torch_xla
+            return torch_xla._XLAC._xla_get_tensor_id(tensor)
         # Fallback for torch==1.10
         try:
             return tensor.storage().data_ptr()


qihqi:
Hi @muellerzr this is Han from Pytorch/XLA. currently this is affecting the ability to save the model after training for on TPUs. 

for example:
```
+ python3 examples/pytorch/xla_spawn.py --num_cores=4 examples/pytorch/image-pretraining/run_mae.py --dataset_name=cifar10 --remove_unused_columns=False --label_names=pixel_values --mask_ratio=0.75 --norm_pix_loss=True --do_train=true --do_eval=true --base_learning_rate=1.5e-4 --lr_scheduler_type=cosine --weight_decay=0.05 --num_train_epochs=3 --warmup_ratio=0.05 '--per_device_train_batch_size=8 ' '--per_device_eval_batch_size=8 ' --logging_strategy=steps --logging_steps=30 --evaluation_strategy=epoch --save_strategy=epoch --load_best_model_at_end=True --save_total_limit=3 --seed=1337 --output_dir=MAE --overwrite_output_dir=true --logging_dir=./tensorboard-metrics --debug=tpu_metrics_debug
``` 
would raise Runtime error. 

Please take a look, thanks!

cc. @JackCaoG

Narsil:
Hey thanks for the patch.

I really relinquish relying on private torch parts (aside from backporting functionality to older torch versions).
Also tensor_id isn't a `data_ptr` iiuc, so it would probably break other logic (trying to figure out storage space and such).

Can we move the fix directly in `transformers` to keep as-is ?

muellerzr:
Agree with @Narsil, it'd be better to keep this isolated to the library, with either `transformers` or `accelerate` (or both if needed)

qihqi:
Sounds good, thanks @Narsil  @muellerzr !

PR #350, opened 2023-08-29
Reduce memory allocations

cospectrum:


Narsil:
Thanks for the PR !

Have you noticed any performance change ? That I could try to reproduce.

The code looks to be better anyway.

cospectrum:
I definitely noticed an improvement in `deserialization` performance with `cargo bench`, for me it is about 100 microseconds (4%). The `serialization` bench does not show stable changes for me. I didn't check python

PR #356, opened 2023-09-08
Update types for load_model filename

kevinhu:
# What does this PR do?

The type for `filename` in `load_model` should be the same as the `load_file` method that it calls. Also, fixes a couple of typos.


Narsil:
LGTM. Thanks for the PR !

PR #362, opened 2023-09-16
perf: make loading shallow weights up to 20X faster

### LINKED ISSUE #361: Loading shallow/augmenting weights (such as LoRAs) with a lot of layers are 10X slower than pure torch
### System Info

```
- `transformers` version: 4.31.0
- Platform: Linux-6.5.3-060503-generic-x86_64-with-glibc2.37
- Python version: 3.11.2
- Huggingface_hub version: 0.16.4
- Safetensors version: 0.3.1
- Accelerate version: 0.21.0
- Accelerate config:    not found
- PyTorch version (GPU?): 2.0.1+cu117 (False)
- Tensorflow version (GPU?): not installed (NA)
- Flax version (CPU?/GPU?/TPU?): not installed (NA)
- Jax version: not installed
- JaxLib version: not installed
```

### Information

- [X] My own modified scripts

### Reproduction

When using pickled/safetensor weights with lots of layers (think 2000+, which is very common with LoRAs) and very little actual data (32x640, 32x1280) safetensors is extremely slow compared to torch (and it gets exponentially worse when layers go from 100->1000->2500->5000, so it implies this is not a data loading problem).

```
import time
import torch
import secrets
import statistics
import safetensors.torch
from contextlib import contextmanager
from tempfile import NamedTemporaryFile


@contextmanager
def fake_lora(num_layers: int, format: str) -> str:
    state_dict = {}
    for _ in range(num_layers):
        state_dict[secrets.token_hex(8)] = torch.zeros((32, 32))

    with NamedTemporaryFile() as stream:
        if format == "torch":
            torch.save(state_dict, stream)
        elif format == "safetensors":
            buffer = safetensors.torch.save(state_dict)
            stream.write(buffer)
        stream.flush()
        yield stream.name


def load_lora(file_name: str, format: str) -> None:
    if format == "torch":
        torch.load(file_name)
    elif format == "safetensors":
        safetensors.torch.load_file(file_name, device="cpu")


for format in ["torch", "safetensors"]:
    print("Testing format: ", format)
    samples = []
    with fake_lora(2500, format=format) as lora_file:
        for _ in range(10):
            start = time.perf_counter()
            load_lora(lora_file, format=format)
            samples.append(time.perf_counter() - start)

    print("Mean: ", statistics.mean(samples))
    print("Median: ", statistics.median(samples))
```

### Expected behavior

```
Testing format:  torch
Mean:  0.028713150699968538
Median:  0.023241078999944875
Testing format:  safetensors
Mean:  0.33690932710005655
Median:  0.3342677030000232
```

I would expect safetensors to be at least as fast as torch, not an order of magnitude slower.

isidentical:
# What does this PR do?

Fixes #361. Local profiling shows majority of the time is spent on recurring accesses to the tensor metadata on every individual tensor fetching. This might be negligible when fetching a few tensors (and is actually better to have it lazily, so I understand the design intentions) but when reading whole weight files (such as a diffusion LoRA) it adds up quickly and causes exponential performance degradation.

With this PR, the example in #361 is now as fast as torch (and performs even better)! Median went from 0.33 seconds to 0.017 seconds. Torch is still at 0.023 seconds so safetensors beats it :) 

Testing a [real world LoRA](https://huggingface.co/nerijs/pixel-art-xl/blob/main/pixel-art-xl.safetensors) with results in similiar perf gains ([script.py](https://gist.github.com/isidentical/0739bfa3d0c5cb0d6946a7561d7c8171)) where median goes from ~0.4 seconds to under 0.018 seconds (more than 20X).

### DIFF
--- a/bindings/python/src/lib.rs
+++ b/bindings/python/src/lib.rs
@@ -413,50 +413,7 @@ impl Open {
         })
     }
 
-    /// Return the special non tensor information in the header
-    ///
-    /// Returns:
-    ///     (`Dict[str, str]`):
-    ///         The freeform metadata.
-    pub fn metadata(&self) -> Option<HashMap<String, String>> {
-        self.metadata.metadata().clone()
-    }
-
-    /// Returns the names of the tensors in the file.
-    ///
-    /// Returns:
-    ///     (`List[str]`):
-    ///         The name of the tensors contained in that file
-    pub fn keys(&self) -> PyResult<Vec<String>> {
-        let mut keys: Vec<String> = self.metadata.tensors().keys().cloned().collect();
-        keys.sort();
-        Ok(keys)
-    }
-
-    /// Returns a full tensor
-    ///
-    /// Args:
-    ///     name (`str`):
-    ///         The name of the tensor you want
-    ///
-    /// Returns:
-    ///     (`Tensor`):
-    ///         The tensor in the framework you opened the file for.
-    ///
-    /// Example:
-    /// ```python
-    /// from safetensors import safe_open
-    ///
-    /// with safe_open("model.safetensors", framework="pt", device=0) as f:
-    ///     tensor = f.get_tensor("embedding")
-    ///
-    /// ```
-    pub fn get_tensor(&self, name: &str) -> PyResult<PyObject> {
-        let tensors = self.metadata.tensors();
-        let info = tensors.get(name).ok_or_else(|| {
-            SafetensorError::new_err(format!("File does not contain tensor {name}",))
-        })?;
-
+    fn hydrate_tensor(&self, info: &TensorInfo) -> PyResult<PyObject> {
         match &self.storage.as_ref() {
             Storage::Mmap(mmap) => {
                 let data =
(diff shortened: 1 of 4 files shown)


isidentical:
Benchmarks:
```
---------------------------------------------------------------------------------------------------------------------------- benchmark: 4 tests ---------------------------------------------------------------------------------------------------------------------------
Name (time in ms)                                                                                   Min                   Max                  Mean             StdDev                Median                IQR            Outliers       OPS            Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_pt_sf_load_cpu[weight_creator0] (NOW)                                                       1.3438 (1.0)         33.3606 (12.41)        1.5071 (1.0)       1.9692 (111.51)       1.3747 (1.0)       0.0184 (1.0)          2;43  663.5463 (1.0)         511           1
test_pt_sf_load_cpu[weight_creator0] (.benchmarks/Linux-CPython-3.11-64bit/0003_baselin)         2.5957 (1.93)         2.6880 (1.0)          2.6348 (1.75)      0.0177 (1.0)          2.6320 (1.91)      0.0223 (1.21)        103;5  379.5325 (0.57)        329           1
test_pt_sf_load_cpu[weight_creator1] (NOW)                                                      34.9504 (26.01)       73.6812 (27.41)       55.2946 (36.69)    19.0215 (>1000.0)     70.4852 (51.27)    37.2772 (>1000.0)       4;0   18.0850 (0.03)         13           1
test_pt_sf_load_cpu[weight_creator1] (.benchmarks/Linux-CPython-3.11-64bit/0003_baselin)     1,493.7300 (>1000.0)  1,534.2338 (570.77)   1,505.5897 (999.03)   16.4876 (933.67)   1,499.5223 (>1000.0)  16.1243 (877.93)        1;0    0.6642 (0.00)          5           1
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

safetensors/benches/benchmark.rs
```diff
@@ -50,6 +61,30 @@ pub fn bench_deserialize(c: &mut Criterion) {
     });
 }
 
+pub fn bench_deserialize_shallow(c: &mut Criterion) {
```
isidentical, line 64: There is no problem when deserializing raw tensors, so added this to verify and keep it regression free.

bindings/python/src/lib.rs
```diff
@@ -544,6 +501,84 @@ impl Open {
...
+    /// ```
+    pub fn get_tensor(&self, name: &str) -> PyResult<PyObject> {
+        let tensors = self.metadata.tensors();
+        let info = tensors.get(name).ok_or_else(|| {
+            SafetensorError::new_err(format!("File does not contain tensor {name}",))
+        })?;
+        self.hydrate_tensor(info)
+    }
+
+    /// Returns a list of selected tensors
```
isidentical, line 550: Was going to implement it in a way that returns all tensors, but filtering also works perfectly fine. For cases where loading N tensors puts a lot of pressure to RAM (with the end goal of them shifting them to VRAM), one can do batching with `get_tensors()` that loads a batch of tensors, moves them to GPU, and then repeats.

FurkanGozukara:
amazing

Narsil:
Superseed by #363 

Thanks @isidentical for confirming the other PR works just as good and creating this !

PR #363, opened 2023-09-18
Stop recreating the hashmap all the time.

### LINKED ISSUE #361: Loading shallow/augmenting weights (such as LoRAs) with a lot of layers are 10X slower than pure torch

Narsil:
# What does this PR do?

This is a different approach from #362 but should fix the same issue in the same way.

This doesn't change any external python API, and makes every existing code faster (by just using the structurs smarter).
This is better IMO as it fixes every potential use for the `safe_open` method, whereas #362 introduces a new layer that has to be used in order to benefit (so only users of `safetensors.torch.load_file` would actually see the difference).

The crux of the problem was `Metadata::tensors` which is a method that creates a full `HashMap<String, String>` to access the tensors. The structure used to be exactly that and just give directly the structure to callers, therefore relatively free. But since, the structure was refactored into a `HashMap<String, usize>` and `Vec<TensorInfo>`.
The main reason for it, is that the `Vec` is intrinsically ordered, and we maintain the `data_offsets` ordered in order to get efficient comparison when checking for malformed files (with overlapping data_offsets).
It also helps in during the serialization where we already have those information sorted so we write them sorted into the JSON header.

Fixes #361
Potentially superseeds #362

### DIFF
--- a/bindings/python/benches/test_pt.py
+++ b/bindings/python/benches/test_pt.py
@@ -30,6 +30,14 @@ def create_gpt2(n_layers: int):
     return tensors
 
 
+def create_lora(n_layers: int):
+    tensors = {}
+    for i in range(n_layers):
+        tensors[f"lora.{i}.up.weight"] = torch.zeros((32, 32))
+        tensors[f"lora.{i}.down.weight"] = torch.zeros((32, 32))
+    return tensors
+
+
 def test_pt_pt_load_cpu(benchmark):
     # benchmark something
     weights = create_gpt2(12)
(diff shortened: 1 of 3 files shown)


FurkanGozukara:
do you have account on twitter? i would like to follow you

good job

Narsil:
https://twitter.com/narsilou
I'm not very active though :)

FurkanGozukara:
> https://twitter.com/narsilou I'm not very active though :)

Followed thanks

PR #379, opened 2023-11-06
Fix detection of duplicate torch tensors

fxmarty:
As per title. Two tensors with different data may share the same `data_ptr` in case they are views of an other tensor. For example:

```python
import torch

a = torch.rand(8, 10, 50)

b = a[2:4]
c = a[4:6]

print(a.untyped_storage().data_ptr())
print(b.untyped_storage().data_ptr())
print(c.untyped_storage().data_ptr())
```

prints

```
93888779151936
93888779151936
93888779151936
```

### DIFF
--- a/bindings/python/py_src/safetensors/torch.py
+++ b/bindings/python/py_src/safetensors/torch.py
@@ -41,6 +41,10 @@ def storage_size(tensor: torch.Tensor) -> int:
             return tensor.nelement() * _SIZE[tensor.dtype]
 
 
+def storage_offset(tensor: torch.Tensor) -> int:
+    return tensor.storage_offset()
+
+
 def _filter_shared_not_shared(tensors: List[Set[str]], state_dict: Dict[str, torch.Tensor]) -> List[Set[str]]:
     filtered_tensors = []
     for shared in tensors:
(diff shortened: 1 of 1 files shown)


bindings/python/py_src/safetensors/torch.py
```diff
@@ -71,7 +75,8 @@ def _find_shared_tensors(state_dict: Dict[str, torch.Tensor]) -> List[Set[str]]:
     for k, v in state_dict.items():
         if v.device != torch.device("meta") and storage_ptr(v) != 0 and storage_size(v) != 0:
             # Need to add device as key because of multiple GPU.
-            tensors[(v.device, storage_ptr(v), storage_size(v))].add(k)
+            # Need to add storage_offset as key because views may share the same data_ptr.
+            tensors[(v.device, storage_ptr(v), storage_size(v), storage_offset(v))].add(k)
```
Narsil, line 79: I think this line is correct.

It's trying to see which tensors have shared memory, which is indicated by storage.
`offset` will make you miss some shared tensors.

fxmarty: @Narsil @LysandreJik Here is a minimal repro of the issue:

```python
import torch
import torch.nn as nn
from safetensors.torch import _find_shared_tensors

class MyModel(nn.Module):
    def __init__(self):
        super().__init__()
        self.qkv = nn.Linear(20, 30)
        
        self.q = nn.Linear(20, 10)
        self.q.weight = torch.nn.Parameter(self.qkv.weight[:10])
    
        self.v = nn.Linear(20, 10)
        self.v.weight = torch.nn.Parameter(self.qkv.weight[10:20])

    def forward(self, x):
        return x

model = MyModel()
shared_params = _find_shared_tensors(model.state_dict())
print("shared_params", shared_params)
```

printing: `shared_params [{'v.weight', 'qkv.weight', 'q.weight'}, {'qkv.bias'}, {'q.bias'}, {'v.bias'}]`

fxmarty: Okay but I understand that you argue that `_find_shared_tensors` is meant to group tensors sharing memory, no matter whether some of them are views and the actual data is different. Is that the case?

Despite the fix https://github.com/huggingface/transformers/pull/27314, Transformers still calls `safetensors.torch.save_file`, and complains that tensors still have shared memory: https://github.com/fxmarty/safetensors/blob/f04d064884be5ede7b6f7d844ce22b793607d091/bindings/python/py_src/safetensors/torch.py#L474. Couldn't Transformers call `save_model` as the error suggests?

Narsil:
Fixed directly in transformers.

PR #386, opened 2023-11-17
use BTreeMap to preserve metadata dict order

wkpark:
# What does this PR do?

~~use `preserve_order` feature of serde_json to preserve the order of the metadata dict.~~
use ~~`IndexMap`~~ `BTreeMap` to preserve metadata dict order to get a reproducible hash of a given safetensors after saving.

(without metadata, the hash of the stored safetensors is always the same. but with metadata, the hash is not always reproducible.)

### DIFF
--- a/bindings/python/src/lib.rs
+++ b/bindings/python/src/lib.rs
@@ -16,6 +16,7 @@ use std::iter::FromIterator;
 use std::ops::Bound;
 use std::path::PathBuf;
 use std::sync::Arc;
+use std::collections::BTreeMap;
 
 static TORCH_MODULE: GILOnceCell<Py<PyModule>> = GILOnceCell::new();
 static NUMPY_MODULE: GILOnceCell<Py<PyModule>> = GILOnceCell::new();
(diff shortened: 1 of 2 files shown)


wkpark:
this is a test case
~~~python
import torch, unittest
import hashlib

class TestCase(unittest.TestCase):
    def test_hash_example(self):
        tensors = {
            "a": torch.zeros((2, 2)),
            "b": torch.zeros((100, 100), dtype=torch.float16),
            "test": torch.zeros((1024, 1024), dtype=torch.float32),
            "test2": torch.zeros((1024, 1024), dtype=torch.float32),
        }

        metadata = {
            "foo": "hello world",
            "bar": "metadata ordering test",
            "fo": "1",
            "bu": "{'foo':'1','boo':'2'}",
        }

        tensors2 = tensors.copy()

        save_file(tensors, "./out.safetensors", metadata)

        blksize = 1024 * 1024

        hash_sha256 = hashlib.sha256()
        with open("./out.safetensors", "rb") as file:
            for chunk in iter(lambda: file.read(blksize), b""):
                hash_sha256.update(chunk)

        hasha = hash_sha256.hexdigest()

        save_file(tensors, "./out1.safetensors", metadata)

        hash_sha256 = hashlib.sha256()
        with open("./out1.safetensors", "rb") as file:
            for chunk in iter(lambda: file.read(blksize), b""):
                hash_sha256.update(chunk)

        hashb = hash_sha256.hexdigest()

        self.assertTrue(hasha == hashb, f"hashes are different")
~~~

Narsil:
Why is it important to get stable SHA in the first place ?

If we were to do this, I'd rather use `BTreeMap` instead, that avoids creating a dependency in the core rust crate.

wkpark:
> Why is it important to get stable SHA in the first place ?
> 

a reproducible hash is important in some cases, for example:
- one can make sha256 in memory without saving `safetensors`. after saving `safetensors` we don't have to recalculate sha again.

> If we were to do this, I'd rather use `BTreeMap` instead, that avoids creating a dependency in the core rust crate.

Thank you for your suggestion. fixed now.

PR #388, opened 2023-11-18
Fix/errors for some models including Mistral-7b and Flan-ul2

lachlancahill:
* Use os.path.basename for all OS compatibility including Windows. 
 * Identify if the hub has downloaded to a snapshots subfolder and move to root directory before committing so that commits work.

### DIFF
--- a/bindings/python/convert.py
+++ b/bindings/python/convert.py
@@ -162,7 +162,7 @@ def convert_multi(
     local_filenames.append(index)
 
     operations = [
-        CommitOperationAdd(path_in_repo=local.split("/")[-1], path_or_fileobj=local) for local in local_filenames
+        CommitOperationAdd(path_in_repo=os.path.basename(local), path_or_fileobj=local) for local in local_filenames
     ]
     errors: List[Tuple[str, "Exception"]] = []
 
--- a/bindings/python/convert.py
+++ b/bindings/python/convert.py
@@ -145,7 +145,6 @@ def convert_multi(
         sf_filename = rename(pt_filename)
         sf_filename = os.path.join(folder, sf_filename)
         convert_file(pt_filename, sf_filename, discard_names=discard_names)
-
         local_filenames.append(sf_filename)
 
     index = os.path.join(folder, "model.safetensors.index.json")
--- a/bindings/python/convert.py
+++ b/bindings/python/convert.py
@@ -146,12 +146,7 @@ def convert_multi(
         sf_filename = os.path.join(folder, sf_filename)
         convert_file(pt_filename, sf_filename, discard_names=discard_names)
 
-        if 'snapshots' in sf_filename.split(os.sep):
-            new_sf_path = os.path.join(folder, os.path.basename(sf_filename))
-            shutil.move(sf_filename, new_sf_path)
-            local_filenames.append(new_sf_path)
-        else:
-            local_filenames.append(sf_filename)
+        local_filenames.append(sf_filename)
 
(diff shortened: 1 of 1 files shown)


Narsil:
Thanks a lot for your contribution @lachlancahill .

Happy to include the Windows fix (my bad on that path splitting).

For the snapshots thing I don't really like it as it is (anyone could have used any path there).
Are you ok, if we make it another PR and you explain a little bit more what's happening for you here ?
What's happening and what is failing ? So we can maybe find a workaround that is a little bit more robust.

bindings/python/convert.py
```diff
@@ -145,7 +145,13 @@ def convert_multi(
...
         sf_filename = os.path.join(folder, sf_filename)
         convert_file(pt_filename, sf_filename, discard_names=discard_names)
-        local_filenames.append(sf_filename)
+
+        if 'snapshots' in sf_filename.split(os.sep):
+            new_sf_path = os.path.join(folder, os.path.basename(sf_filename))
+            shutil.move(sf_filename, new_sf_path)
+            local_filenames.append(new_sf_path)
+        else:
+            local_filenames.append(sf_filename)
```
Narsil, line 154:
```suggestion
```

Narsil: Let's split that into another PR if that's ok. The current detection seems a bit brittle and I'm not sure I understand what's the underlying issue.

lachlancahill:
> Thanks a lot for your contribution @lachlancahill .
> 
> Happy to include the Windows fix (my bad on that path splitting).
> 
> For the snapshots thing I don't really like it as it is (anyone could have used any path there). Are you ok, if we make it another PR and you explain a little bit more what's happening for you here ? What's happening and what is failing ? So we can maybe find a workaround that is a little bit more robust.

My pleasure, thank you for your great feedback. 

I've removed the 'snapshots' change from the PR though worth keeping in mind that if we take out that whole block we need to add back the original `local_filenames.append(sf_filename)` line. I've actioned this in the latest commit.

To provide more detail on the snapshots issue, while investigating errors I noticed the folder structure for Mistral (similar for flan-ul2) was like the below in the temporary directory:

```
|-- modelss--mistralai--Mistral-7B-v0.1
    |-- .locks
        |-- models--mistralai--Mistral-7B-v0.1
    |-- model.safetensors.index.json
    |-- models--mistralai--Mistral-7B-v0.1
        |-- blobs
        |-- refs
            |-- main
        |-- snapshots
            |-- 5e9c98b96d071dce59368012254c55b0ec6f8658
                |-- model-00001-of-00002.safetensors
                |-- model-00002-of-00002.safetensors
                |-- pytorch_model-00001-of-00002.bin
                |-- pytorch_model-00002-of-00002.bin
                |-- pytorch_model.bin.index.json
```

I think the version I was working with had a step which checked the converted files, and this non-standard folder structure was causing issues (I think the checking function was looking in the root folder for the .safetensors files). In the latest version it looks like you have removed that check which I think makes this change unnecessary. Thanks for all your great work, this is such a valuable resource. 😄

bindings/python/convert.py
lachlancahill: Actioned the reversal in another commit (we still need `local_filenames.append(sf_filename)`)

Narsil:
Yes I removed the check because we're now pretty confident about the conversions as they exist, and removing it allows to convert models with much less RAM than the actual model requires.

I would keep it if I knew a given machine had enough RAM to run the check, but it seems too involved/unreliable atm.

PR #404, opened 2023-12-15
Implement PyTorch support for float8 types (F8_E5M2 and F8_E4M3)

zeux:
This PR completes support for float8 types by making them available when using safetensors from Python with PyTorch; float8 types are supported by PyTorch since July (https://github.com/pytorch/pytorch/pull/104242).

Note that PyTorch name for e4m3 type has an extra "fn" prefix to match MLIR, but the format should be the same ("fn" means "finite").

The added test checks that -0.5 roundtrips in both formats - both types are single-byte and have the same representation for zero, but different representations for -0.5.

### DIFF
--- a/bindings/python/py_src/safetensors/torch.py
+++ b/bindings/python/py_src/safetensors/torch.py
@@ -349,6 +349,8 @@ _SIZE = {
     torch.int8: 1,
     torch.bool: 1,
     torch.float64: 8,
+    torch.float8_e4m3fn: 1,
+    torch.float8_e5m2: 1,
 }
 
 _TYPES = {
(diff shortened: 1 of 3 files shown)


zeux:
This would need to increase the version of PyTorch required (I think it's currently listed as >=1.10, but float8 formats are only supported in 2.1. I am not sure what the versioning restrictions are here, but I can probably change the code to only use float8 et al when it is present / when requested? This will complicate the Python wrapper a little bit.

KohakuBlueleaf:
Really need this!

Narsil:
Thanks a lot for this PR, sorry I missed it when you published it.

PR #336, opened 2023-08-23
Temporary revert of the breaking change (keep it for 0.4.0).

Narsil:
# What does this PR do?

Need to do a patch release for big endian breakage. Not worth a breaking release for now.

Fixes # (issue) or description of the problem this PR solves.

PR #408, opened 2023-12-18
Reinstate brace checks from #306

akx:
# What does this PR do?

Reinstates the checks "temporarily reverted" in https://github.com/huggingface/safetensors/pull/336 (that doesn't describe what this mysterious "big endian breakage" is that was fixed by this change 🤷).

It's past 0.4.0 already, and we know there's nothing quite as permanent as something temporary. 😁


akx:
@Narsil s390x test seems to be failing on... Tailscale? 🤨

PR #411, opened 2023-12-27
use BTreeMap to preserve metadata dict order

wkpark:
# What does this PR do?

use `BTreeMap` to preserve metadata dict order to get a reproducible hash (or after in-memory hashing) of a given safetensors after saving.

(without metadata, the hash of the stored safetensors is always the same. but with metadata, the hash is not always reproducible.)

(rebased. reopened manually. PR #386


wkpark:
stale check

akx:
ping @Narsil

PR #419, opened 2024-01-04
Adding MLX backend.

Narsil:
# What does this PR do?

Adds MLX backend.
MLX already has native safetensors support, but this was trivial to add.

### DIFF
--- a/bindings/python/py_src/safetensors/mlx.py
+++ b/bindings/python/py_src/safetensors/mlx.py
@@ -9,7 +9,7 @@ from safetensors import numpy, safe_open
 
 def save(tensors: Dict[str, mx.array], metadata: Optional[Dict[str, str]] = None) -> bytes:
     """
-    Saves a dictionnary of tensors into raw bytes in safetensors format.
+    Saves a dictionary of tensors into raw bytes in safetensors format.
 
     Args:
         tensors (`Dict[str, mx.array]`):
@@ -42,7 +42,7 @@ def save_file(
     metadata: Optional[Dict[str, str]] = None,
 ) -> None:
     """
-    Saves a dictionnary of tensors into raw bytes in safetensors format.
+    Saves a dictionary of tensors into raw bytes in safetensors format.
 
     Args:
         tensors (`Dict[str, mx.array]`):
@@ -85,7 +85,7 @@ def load(data: bytes) -> Dict[str, mx.array]:
     Example:
 
     ```python
-    from safetensors.flax import load
+    from safetensors.mlx import load
 
     file_path = "./my_folder/bert.safetensors"
     with open(file_path, "rb") as f:
@@ -100,14 +100,11 @@ def load(data: bytes) -> Dict[str, mx.array]:
 
 def load_file(filename: Union[str, os.PathLike]) -> Dict[str, mx.array]:
     """
-    Loads a safetensors file into flax format.
+    Loads a safetensors file into MLX format.
 
     Args:
         filename (`str`, or `os.PathLike`)):
             The name of the file which contains the tensors
-        device (`Dict[str, any]`, *optional*, defaults to `cpu`):
-            The device where the tensors need to be located after load.
-            available options are all regular flax device locations
 
     Returns:
         `Dict[str, mx.array]`: dictionary that contains name as key, value as `mx.array`
--- a/bindings/python/py_src/safetensors/mlx.py
+++ b/bindings/python/py_src/safetensors/mlx.py
@@ -80,7 +80,7 @@ def load(data: bytes) -> Dict[str, mx.array]:
             The content of a safetensors file
 
     Returns:
-        `Dict[str, mx.array]`: dictionary that contains name as key, value as `mx.array` on cpu
+        `Dict[str, mx.array]`: dictionary that contains name as key, value as `mx.array`
 
     Example:
 
--- a/bindings/python/py_src/safetensors/mlx.py
+++ b/bindings/python/py_src/safetensors/mlx.py
@@ -73,7 +73,7 @@ def save_file(
 
 def load(data: bytes) -> Dict[str, mx.array]:
     """
-    Loads a safetensors file into flax format from pure bytes.
+    Loads a safetensors file into MLX format from pure bytes.
 
     Args:
         data (`bytes`):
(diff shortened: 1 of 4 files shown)


Narsil:
We can safely ignore the failure which is entirely unrelated to the PR (more recent clippy detects an issue with PyO3 itself, which will most likely be fixed, by simply upgrading PyO3)

Vaibhavs10 (commented):
Dope! Thanks for putting this up! 🚀 Conceptually, this looks good to me.

That said, I'm not an expert in MLX so would appreciate if Awni could do a deeper review.

One qq: Can we potentially also test for persisting and loading quantised models as well? we don't need to be exhaustive here, but just testing for 4-bit should be okay IMO.

bindings/python/py_src/safetensors/mlx.py
```diff
@@ -0,0 +1,141 @@
...
+    save_file(tensors, "model.safetensors")
+    ```
+    """
+    np_tensors = _mx2np(tensors)
+    return numpy.save_file(np_tensors, filename, metadata=metadata)
+
+
+def load(data: bytes) -> Dict[str, mx.array]:
+    """
+    Loads a safetensors file into flax format from pure bytes.
```
pcuenca, line 76:
```suggestion
    Loads a safetensors file into MLX format from pure bytes.
```

bindings/python/py_src/safetensors/mlx.py
```diff
@@ -0,0 +1,141 @@
...
+        data (`bytes`):
+            The content of a safetensors file
+
+    Returns:
+        `Dict[str, mx.array]`: dictionary that contains name as key, value as `mx.array` on cpu
+
+    Example:
+
+    ```python
+    from safetensors.flax import load
```
pcuenca, line 88:
```suggestion
    from safetensors.mlx import load
```

bindings/python/py_src/safetensors/mlx.py
```diff
@@ -0,0 +1,141 @@
...
+    loaded = load(data)
+    ```
+    """
+    flat = numpy.load(data)
+    return _np2mx(flat)
+
+
+def load_file(filename: Union[str, os.PathLike]) -> Dict[str, mx.array]:
+    """
+    Loads a safetensors file into flax format.
```
pcuenca, line 103:
```suggestion
    Loads a safetensors file into MLX format.
```

bindings/python/py_src/safetensors/mlx.py
```diff
@@ -0,0 +1,141 @@
...
+
+    Args:
+        filename (`str`, or `os.PathLike`)):
+            The name of the file which contains the tensors
+        device (`Dict[str, any]`, *optional*, defaults to `cpu`):
+            The device where the tensors need to be located after load.
+            available options are all regular flax device locations
```
pcuenca, line 110:
```suggestion
```

Narsil:
Quantized is u8, therefore it should work out of the box (not sure how mlx persists quantized information in npx/npz, i'm guessing it's not saved)

bindings/python/py_src/safetensors/mlx.py
```diff
@@ -0,0 +1,141 @@
...
+def load(data: bytes) -> Dict[str, mx.array]:
+    """
+    Loads a safetensors file into flax format from pure bytes.
+
+    Args:
+        data (`bytes`):
+            The content of a safetensors file
+
+    Returns:
+        `Dict[str, mx.array]`: dictionary that contains name as key, value as `mx.array` on cpu
```
pcuenca, line 83:
```suggestion
        `Dict[str, mx.array]`: dictionary that contains name as key, value as `mx.array`
```

bindings/python/py_src/safetensors/mlx.py
```diff
@@ -0,0 +1,141 @@
...
+    return result
+
+
+def _np2mx(numpy_dict: Dict[str, np.ndarray]) -> Dict[str, mx.array]:
+    for k, v in numpy_dict.items():
+        numpy_dict[k] = mx.array(v)
+    return numpy_dict
+
+
+def _mx2np(mx_dict: Dict[str, mx.array]) -> Dict[str, np.array]:
```
pcuenca, line 137: Does this handle `bfloat16`?

bindings/python/tests/test_mlx_comparison.py
```diff
@@ -0,0 +1,59 @@
...
+
+if platform.system() == "Darwin":
+    # This platform is not supported, we don't want to crash on import
+    # This test will be skipped anyway.
+    import mlx.core as mx
+    from safetensors import safe_open
+    from safetensors.mlx import load_file, save_file
+
+
+# Jax doesn't not exist on Windows
```
pcuenca, line 13:
```suggestion
# MLX only exists on Mac
```

bindings/python/py_src/safetensors/mlx.py
pcuenca: Oh, I see in the tests below that it does :)

pcuenca (approved):
Awesome! 🔥

bindings/python/py_src/safetensors/mlx.py
Narsil: Yup, using the same flax trick which is just a special named dtype.

awni (approved):
LGTM! Thanks for adding that!

I think using your built-in `numpy.save` is probably a good call from a maintenance standpoint.

But just a thought that this will involve a copy (on the load side as well). Currently we do a copy to get in and out of Numpy.

Narsil:
> But just a thought that this will involve a copy (on the load side as well). Currently we do a copy to get in and out of Numpy.


Good to know. The only thing really needed is `frombuffer` equivalent in order to get loading partial tensors working.
On the rust side, I use the slice information given by the user to create. a CPU buffer of the correct data within the tensor, and use `frombuffer`  on given framework in order to send the correct data where it belongs.

If there's a copy from numpy-> MLX, that means that currently there are 2 copies created.

File -> Slice (local buffer inside rust) -> numpy.from_buffer (should be zero copy) -> mlx (new copy).

Is that correct ?

awni:
> If there's a copy from numpy-> MLX, that means that currently there are 2 copies created.

Yes that's correct. We can't avoid the copy from NumPy because we have to use specifically allocated memory to make it available to both the CPU and GPU.

PR #424, opened 2024-01-08
[docs] Use `hfoptions` for metadata

mishig25:
Based on [this comment](https://github.com/huggingface/safetensors/pull/417/files#r1442176507), bringing back the example on how to use http range requests to get metadata. Also, using [hfoptions](https://github.com/huggingface/doc-builder?tab=readme-ov-file#options) syntax for better UI/UX

you can test the page here: https://moon-ci-docs.huggingface.co/docs/safetensors/pr_424/en/metadata_parsing#usage

https://github.com/huggingface/safetensors/assets/11827707/53edac7e-5bfe-43b6-a158-a026d7540ac3

### DIFF
--- a/docs/source/metadata_parsing.mdx
+++ b/docs/source/metadata_parsing.mdx
@@ -23,7 +23,7 @@ There can be many potential use cases. For instance, we use it on the HuggingFac
 <hfoptions id="metadata">
 
 <hfoption id="http">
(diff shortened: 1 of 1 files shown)


Wauplin (approved):
Thanks for taking care of it @mishig25! Nice use case for hfoptions.
Not sure about naming `http` the one where we show a python snippet to parse using http requests but I don't have a better one to suggest so keeping it should be good :+1:

mishig25:
> Not sure about naming http the one where we show a python snippet to parse using http requests but I don't have a better one to suggest so keeping it should be good 👍

Indeed. That is the question. Maybe others can chime in

julien-c (approved):
ah nice, thanks for adding this back 🙏 

I think `http` (or `raw` but it's weird) works

PR #434, opened 2024-01-27
Reinstate brace checks implemented in #306

akx:
# What does this PR do?

Reinstates the checks "temporarily reverted" in https://github.com/huggingface/safetensors/pull/336 (that doesn't describe what this mysterious "big endian breakage" is that was fixed by this change 🤷).

It's past 0.4.0 already, and we know there's nothing quite as permanent as something temporary. 😁

This is #408 reopened after it was closed by GitHub Actions without review.


akx:
ping @Narsil

akx:
Gentle un-stale-ing ping.

Issue #437, opened 2024-02-08
Loading a model with safetensors results in OOM error whereas torch.load() does not.


maxdebayser:
### System Info

- `transformers` version: 4.37.1
- Platform: Linux-4.18.0-477.21.1.el8_8.ppc64le-ppc64le-with-glibc2.28
- Python version: 3.11.2
- Huggingface_hub version: 0.20.3
- Safetensors version: 0.4.2
- Accelerate version: 0.24.1
- Accelerate config: 	not found
- PyTorch version (GPU?): 2.1.0a0+gitda9b770 (True)
- Tensorflow version (GPU?): not installed (NA)
- Flax version (CPU?/GPU?/TPU?): not installed (NA)
- Jax version: not installed
- JaxLib version: not installed
- Using GPU in script?: yes
- Using distributed or parallel set-up in script?: no

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Reproduction

I'm trying to load LLama 13B on a 32GB V100. With safetensors I get an OOM error while  `torch.load()` just works. I've observed that since the model is initialized in `from_pretrained` in the transformers library before everything else, it already occupies most of the memory. After that the transformers library load the tensors from disk and copies them into the ones in the model. During `torch.load` the GPU memory does not increase, while with safetensors it increases with every call to `get_tensor()` until the GPU memory is exhausted.

### Expected behavior

I would expect safetensors to have more or less the same behavior as `torch.load()` with respect to GPU memory usage.

PR #438, opened 2024-02-08
Pass device=cpu to torch.asarray() in get_tensor()

maxdebayser:
The code in get_tensor assumes that torch.asarray will always load the tensor in cpu, and then the tensor is copied to self.device in the "to" call at the end of the function. However, that assumption is not always correct. To make sure, device=cpu has to be passed to asarray.

Another option would be to set the final device since the beginning:
```
let device = match self.device {
    Device::Cpu     => format!("cpu"),
    Device::Cuda(n) => format!("cuda:{n}"),
    Device::Mps     => format!("mps"),
    Device::Npu(n)  => format!("npu:{n}"),
    Device::Xpu(n)  => format!("xpu:{n}"),
}.into_py(py);

let kwargs = [(intern!(py, "dtype"), torch_uint8),(intern!(py, "device"), device)].into_py_dict(py);
```
but I'm not sure what the implications of that would be.

Fixes issue #437

### DIFF
--- a/bindings/python/src/lib.rs
+++ b/bindings/python/src/lib.rs
@@ -512,7 +512,7 @@ impl Open {
                     let torch = get_module(py, &TORCH_MODULE)?;
                     let dtype: PyObject = get_pydtype(torch, info.dtype, false)?;
                     let torch_uint8: PyObject = get_pydtype(torch, Dtype::U8, false)?;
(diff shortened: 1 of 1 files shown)


maxdebayser:
@Narsil, I'm wondering of any of the maintainers could take a look at this. It's a simple issue, but it's blocking some of our users. Thanks!

dtrifiro:
Any updates on this?

PR #440, opened 2024-02-14
Adding support for integer indexing `[0, :2, -1]`.

### LINKED ISSUE #439: Slice loading errors with integer indices
### System Info

safetensors==0.4.2

### Information

- [X] The official example scripts
- [X] My own modified scripts

### Reproduction

```python

import torch

save_file({"test":torch.zeros([10,10])}, "test.st")

with safe_open("test.st", framework="pt", device="cpu") as f:
    print(f.get_slice("test")[0, :])
```

### Expected behavior

I expect this to return a slice of the tensor "test", instead I get the error:

`argument 'slices': failed to extract enum Slice ('Slice | Slices')`

As far as I can see there's no notes in the documentation or examples about integer slices not working?
https://huggingface.co/docs/safetensors/en/index

Narsil:
# What does this PR do?

Fixes #439 

Adds support for integer indexing.

### DIFF
--- a/bindings/python/src/lib.rs
+++ b/bindings/python/src/lib.rs
@@ -167,24 +167,41 @@ fn deserialize(py: Python, bytes: &[u8]) -> PyResult<Vec<(String, HashMap<String
     Ok(items)
 }
 
-fn slice_to_indexer(slice: &PySlice) -> Result<TensorIndexer, PyErr> {
-    let py_start = slice.getattr(intern!(slice.py(), "start"))?;
-    let start: Option<usize> = py_start.extract()?;
-    let start = if let Some(start) = start {
-        Bound::Included(start)
-    } else {
-        Bound::Unbounded
-    };
+fn slice_to_indexer(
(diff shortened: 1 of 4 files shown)


bindings/python/src/lib.rs
```diff
@@ -167,24 +167,41 @@ fn deserialize(py: Python, bytes: &[u8]) -> PyResult<Vec<(String, HashMap<String
...
+            };
+            Ok(TensorIndexer::Narrow(start, stop))
+        }
+        SliceIndex::Index(idx) => {
+            if idx < 0 {
+                let idx = dim
+                    .checked_add_signed(idx as isize)
+                    .ok_or(SafetensorError::new_err(format!(
+                        "Invalid index {idx} for dimension {dim_idx} of size {dim}"
+                    )))?;
```
ArthurZucker, line 198: let's make sure we test the panic

ArthurZucker (approved):

bindings/python/src/lib.rs
Narsil: This is checked, the other out of bounds wasn´t, I added it.

PR #452, opened 2024-03-14
Fix byteswap for BF16 tensor on big-endian machine

### LINKED ISSUE #448: Load incorrect element values in bfloat16 tensor on big-endian
### System Info

```
>>> safetensors.__version__
'0.4.2'
```

### Information

- [X] My own modified scripts

### Reproduction

When I executed transformer models in bfloat16 at HF, I got the incorrect result on s390x. I realized values in weights are different between x86 and s390x. The following is a small reproduction.

Execute the following program on x86
```
import torch

tensors = {
   "weight1": torch.ones((8, 8), dtype=torch.bfloat16),
}
save_file(tensors, "bf16.safetensors")

read_tensors = {}
with safe_open("bf16.safetensors", framework="pt", device="cpu") as f:
   for key in f.keys():
       read_tensors[key] = f.get_tensor(key)
print(read_tensors)
```

Copy `bf16.safetensors` into s390x machine. Then, execute the following program
```
import torch

read_tensors = {}
with safe_open("bf16.safetensors", framework="pt", device="cpu") as f:
   for key in f.keys():
       read_tensors[key] = f.get_tensor(key)
print(read_tensors)
```

The result is as follows:
```
{'weight1': tensor([[7.6294e-06, 7.6294e-06, 7.6294e-06, 7.6294e-06, 7.6294e-06, 7.6294e-06,
         7.6294e-06, 7.6294e-06],
        [7.6294e-06, 7.6294e-06, 7.6294e-06, 7.6294e-06, 7.6294e-06, 7.6294e-06,
         7.6294e-06, 7.6294e-06],
        [7.6294e-06, 7.6294e-06, 7.6294e-06, 7.6294e-06, 7.6294e-06, 7.6294e-06,
         7.6294e-06, 7.6294e-06],
        [7.6294e-06, 7.6294e-06, 7.6294e-06, 7.6294e-06, 7.6294e-06, 7.6294e-06,
         7.6294e-06, 7.6294e-06],
        [7.6294e-06, 7.6294e-06, 7.6294e-06, 7.6294e-06, 7.6294e-06, 7.6294e-06,
         7.6294e-06, 7.6294e-06],
        [7.6294e-06, 7.6294e-06, 7.6294e-06, 7.6294e-06, 7.6294e-06, 7.6294e-06,
         7.6294e-06, 7.6294e-06],
        [7.6294e-06, 7.6294e-06, 7.6294e-06, 7.6294e-06, 7.6294e-06, 7.6294e-06,
         7.6294e-06, 7.6294e-06],
        [7.6294e-06, 7.6294e-06, 7.6294e-06, 7.6294e-06, 7.6294e-06, 7.6294e-06,
         7.6294e-06, 7.6294e-06]], dtype=torch.bfloat16)}
```

### Expected behavior

The result on s390x should be as follows:
```
{'weight1': tensor([[1., 1., 1., 1., 1., 1., 1., 1.],
        [1., 1., 1., 1., 1., 1., 1., 1.],
        [1., 1., 1., 1., 1., 1., 1., 1.],
        [1., 1., 1., 1., 1., 1., 1., 1.],
        [1., 1., 1., 1., 1., 1., 1., 1.],
        [1., 1., 1., 1., 1., 1., 1., 1.],
        [1., 1., 1., 1., 1., 1., 1., 1.],
        [1., 1., 1., 1., 1., 1., 1., 1.]], dtype=torch.bfloat16)}
```

My colleague is curious whether [this code](https://github.com/huggingface/safetensors/blob/main/bindings/python/src/lib.rs#L560-L566) works well.

kiszk:
# What does this PR do?

This PR fixes an incorrect Tensor's data swap in BF16 on a big-endian machine. It uses `Storage.byteswap(datatype)` in PyTorch instead of `numpy.byteswap()`. This is because numpy does not support BF16. Unnecessary data conversions between BF16 and F16 caused incorrect results.

Fixes #448

### DIFF
--- a/bindings/python/src/lib.rs
+++ b/bindings/python/src/lib.rs
@@ -542,47 +542,40 @@ impl Open {
                         .get(py)
                         .ok_or_else(|| SafetensorError::new_err("Could not find storage"))?;
                     let storage: &PyBound<PyAny> = storage.bind(py);
-                    let storage_slice = storage
+                    let mut storage_slice = storage
                         .getattr(intern!(py, "__getitem__"))?
                         .call1((slice,))?;
 
                     let sys = PyModule::import_bound(py, intern!(py, "sys"))?;
                     let byteorder: String = sys.getattr(intern!(py, "byteorder"))?.extract()?;
 
+                    if byteorder == "big" {
+                        let module = PyModule::import(py, intern!(py, "torch"))?;
+                        let version: String = module.getattr(intern!(py, "__version__"))?.extract()?;
+                        let version = Version::from_string(&version).map_err(SafetensorError::new_err)?;
+
+                        if version >= Version::new(2, 1, 0) {
+                            // PyTorch 2.1 or later implements Storage.byteswap()
+                            storage_slice = storage_slice
+                                .getattr(intern!(py, "clone"))?
+                                .call0()?;
+                            let dtype: PyObject = get_pydtype(torch, info.dtype, false)?;
+                            storage_slice
+                                .getattr(intern!(py, "byteswap"))?
+                                .call1((dtype,))?;
+                            // storage_slice.clone().byteswap(dtype)
+                        } else {
+                            return Err(SafetensorError::new_err("PyTorch 2.1 or later is required for big-endian machine"));
+                        }
+                    }
+
                     let mut tensor = torch
                         .getattr(intern!(py, "asarray"))?
                         .call((storage_slice,), Some(&kwargs))?
                         .getattr(intern!(py, "view"))?
-                        .call((), Some(&view_kwargs))?;
-
-                    if byteorder == "big" {
-                        let inplace_kwargs =
-                            [(intern!(py, "inplace"), false.into_py(py))].into_py_dict_bound(py);
-                        if info.dtype == Dtype::BF16 {
-                            let torch_f16: PyObject = get_pydtype(torch, Dtype::F16, false)?;
-                            tensor = tensor.getattr(intern!(py, "to"))?.call(
-                                (),
-                                Some(&[(intern!(py, "dtype"), torch_f16)].into_py_dict_bound(py)),
-                            )?;
-                        }
-
-                        let numpy = tensor
-                            .getattr(intern!(py, "numpy"))?
-                            .call0()?
-                            .getattr("byteswap")?
-                            .call((), Some(&inplace_kwargs))?;
-                        tensor = torch.getattr(intern!(py, "from_numpy"))?.call1((numpy,))?;
-
-                        if info.dtype == Dtype::BF16 {
-                            let torch_bf16: PyObject = get_pydtype(torch, Dtype::BF16, false)?;
-                            tensor = tensor.getattr(intern!(py, "to"))?.call(
-                                (),
-                                Some(&[(intern!(py, "dtype"), torch_bf16)].into_py_dict_bound(py)),
-                            )?;
-                        }
-                    }
-
-                    tensor = tensor.getattr(intern!(py, "reshape"))?.call1((shape,))?;
+                        .call((), Some(view_kwargs))?
+                        .getattr(intern!(py, "reshape"))?
+                        .call1((shape,))?;
                     if self.device != Device::Cpu {
                         let device: PyObject = self.device.clone().into_py(py);
                         let kwargs = PyDict::new_bound(py);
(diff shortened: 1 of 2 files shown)


kiszk:
@Narsil Do you have time to look at this PR?

kiszk:
@Narsil Do you have time to look at this PR?

kiszk:
@Narsil Do you have time to look at this PR?

kiszk:
@Narsil Do you have time to look at this PR?

kiszk:
@Narsil Do you have time to look at this PR?

kiszk:
@Narsil Would it be possible to look at this PR?

abalib:
Can we kindly ask someone to review and merge this PR?

Narsil:
Hi @kiszk,

Sorry about the delay, I haven't been able to check every place I'm being pinged on (and far from it).

I understand the issue and the fix.
From what I understand, the fix introduced 2 issues imho:

- It limits byteswap to torch 2.1 which is not required (for all the other dtypes other than bf16).
- It introduces a `clone` operation, which is something I've vividly tried to avoid (given the size of the objects, cloning is super costly).

I'll try to take a look at how I can improve that solution if you're ok with my assumptions.
Also there is a test that was supposed to catch that: https://github.com/huggingface/safetensors/blob/7d29f617a33251a8b5f9e8228c004edc3a730f4a/bindings/python/tests/test_pt_comparison.py#L53

I am guessing it didn't because it used zeros instead of randn, and bfloat16 is specifically the same bytes for the zero representation :( Therefore we could just update that test.

Narsil:
Managed to trigger the issue https://github.com/huggingface/safetensors/actions/runs/10109667229/job/27958053499?pr=507 on old tests with random values.

kiszk:
@Narsil, thank you for your response while you were super busy. 

Your points are correct:
1. Since the prior to PyTorch 2.1 does not support Big Endian correctly, I chose the implementation to throw an exception. We can implement the same feature using Rust ([Python was too slow](https://github.com/pytorch/pytorch/pull/101925)) in safetensors since this swap function exists in 2.1 or later.
2. I chose the conservative implementation to avoid a destructive operation. It is fine if we can avoid the `clone` operation.

For testing point, I think that this test uses [`1.0`](https://github.com/huggingface/safetensors/blob/main/bindings/python/tests/test_pt_model.py#L22-L23) instead of zero. Anyway, I just realized [this code](https://github.com/huggingface/safetensors/pull/452/files#diff-39934f58bd5e2054da6e7d1700478540e54fa4686a75a7834050dd58c3d70379R177) does not update the original `model`. I agree that we have to update the test code.

I will work to update some of them if you are still super busy.

Narsil:
@kiszk Can you check the last implementation ?

I think numpy byteswap can still work, since bf16 byteswap and f16 byteswap are the same. The only thing necessary is those `.view(..)` which are just reinterpreting arrays.

No issue with torch version, no clone, faster byteswap.

Could you confirm the branch works for you (all the tests should show issues now).

kiszk:
Sure, I will check the latest main soon.

kiszk:
As you said, byteswap is the same between f16 and bf16.

IMHO, the problem may come from data conversion from [bf16 and f16](https://github.com/huggingface/safetensors/blob/main/bindings/python/src/lib.rs#L599-L603) / [from f16 to bf16](https://github.com/huggingface/safetensors/blob/main/bindings/python/src/lib.rs#L614-L618). We have to do byteswap without any data conversion.

In detail, on a big-endian platform, at the beginning, an element in bf16 tensor has an uninterpretable value since its format is bf16 in little-endian. **In the current implementation, an uninterpretable value is interpreted as bf16, and then it is converted from bf16 to f16**. The value in f16 is not meaningless. Then, it is byteswaped for big-endian in numpy. After byteswap, in numpy, the value is not good in bf16 on a big-endian platform.

kiszk:
@Narsil If I make mistakes, could you please let me know?

Narsil:
@kiszk I didn't say `main` branch, sorry I was referrring to  https://github.com/huggingface/safetensors/pull/507

Narsil:
Superseeded by https://github.com/huggingface/safetensors/pull/507

Basically the issue is that previous code was doing `(bf16) -> to -> (f16) -> byteswap -> to -> (bf16)`.

`to` is a cast operator which changes the underlying bits of the vectors which is not desired.
Changing to `view` (which is torch `reinterpret` operator) does not change the bits anymore, meaning the byteswap is valid.

Closing this assuming the bug is fixed. Let's reopen something if itś  not.

PR #472, opened 2024-04-23
Fixing empty serialization (no tensor) with some metadata.

Narsil:
# What does this PR do?


safetensors/src/tensor.rs
```diff
@@ -833,6 +838,34 @@ mod tests {
...
 
+    #[test]
+    fn test_empty() {
+        let tensors: HashMap<String, TensorView> = HashMap::new();
+
+        let out = serialize(&tensors, &None).unwrap();
+        assert_eq!(
+            out,
+            [8, 0, 0, 0, 0, 0, 0, 0, 123, 125, 32, 32, 32, 32, 32, 32]
+        );
```
McPatate, line 849: Why does calling `serialize` on an empty HashMap result in a non zero length array?

PR #488, opened 2024-06-17
Minor correction about Arrow in README.md

lhoestq:
arrow doesn't require decoding, it's an in-memory format


lhoestq:
taking the liberty to merge, lmk if I should revert

PR #497, opened 2024-07-10
Add conversion validation support for the float8 tensor dtype 

darthhexx:
# What does this PR do?

Torch does not support `equal_cpu` for the torch.float8_e4m3fn or torch.float8_e5m2 dtypes so the convert script fails with errors like:

```bash
RuntimeError: "equal_cpu" not implemented for 'Float8_e4m3fn'
```

This PR changes the output tensor validation to check that the dtypes are the same and the float32 value of each tensor is equal instead. This should always succeed if the values are, even for new dtypes in the future.

### DIFF
--- a/bindings/python/convert.py
+++ b/bindings/python/convert.py
@@ -205,7 +205,7 @@ def convert_file(
     for k in loaded:
         pt_tensor = loaded[k]
         sf_tensor = reloaded[k]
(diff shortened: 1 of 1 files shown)


Narsil:
Hi thanks for the contribution but no this cannot work.

Torch.equal should be implement for fp8, if it's not we should take it upstream to pytorch directly,

I'm closing this but feel free to reopen if you have a compelling case of how it's a big blocker (this convert script is a utility function, not a core member of the library, so you should be able to adapt easily,).

Issue #499, opened 2024-07-13
pytorch: safetensors library hardcodes using CUDA if only device index is provided


dvrogozh:
In relevance to:
* https://github.com/huggingface/transformers/issues/31941

safetensors library hardcodes returning CUDA device if only device index is provided. This causes runtime errors running huggingface models with `pipeline(device_map="auto")` as noted in https://github.com/huggingface/transformers/issues/31941 (see this issue for repro steps). Hardcoding is happening here:
https://github.com/huggingface/safetensors/blob/079781fd0dc455ba0fe851e2b4507c33d0c0d407/bindings/python/src/lib.rs#L296-L297

**A possible solution** might be to return the device returned by `torch.device(N)`. Note however that this will work for non-CUDA devices only after the following change in pytorch will be merged:
* https://github.com/pytorch/pytorch/pull/129119
This change modifies behavior for `torch.device(N)` to return current accelerator device instead of cuda device. This seems to be anticipated change by huggingface according to https://github.com/huggingface/accelerate/pull/2874#issuecomment-2204803710.

CC: @faaany @muellerzr @SunMarc @guangyey

dvrogozh:
FYI, https://github.com/pytorch/pytorch/pull/129119 got merged, so solution which I outlined should now be possible.

dvrogozh:
I have implemented a fix for this issue as I do see it. Please, help review https://github.com/huggingface/safetensors/pull/500.

PR #500, opened 2024-07-15
Query device name from pytorch if only device index is given

dvrogozh:
Fixes: #499
Fixes: https://github.com/huggingface/transformers/issues/31941

In some cases only device index is given on querying device. In this case both PyTorch and Safetensors were returning 'cuda:N' by default. This is causing runtime failures if user actually runs something on non-cuda device and does not have cuda at all. Recently this was addressed on PyTorch side by [1]: starting from PyTorch 2.5 calling 'torch.device(N)' will return current device instead of cuda device.

This commit is making similar change to Safetensors. If only device index is given, Safetensors will query and return device calling 'torch.device(N)'. This change is backward compatible since this call would return 'cuda:N' on PyTorch <=2.4 which aligns with previous Safetensors behavior.

[1]: https://github.com/pytorch/pytorch/pull/129119

CC: @guangyey @jgong5 @faaany @muellerzr @SunMarc @Narsil

### DIFF
--- a/bindings/python/src/lib.rs
+++ b/bindings/python/src/lib.rs
@@ -284,26 +284,40 @@ fn parse_device(name: &str) -> PyResult<usize> {
 
 impl<'source> FromPyObject<'source> for Device {
     fn extract_bound(ob: &PyBound<'source, PyAny>) -> PyResult<Self> {
-        if let Ok(name) = ob.extract::<String>() {
-            match &name[..] {
-                "cpu" => Ok(Device::Cpu),
-                "cuda" => Ok(Device::Cuda(0)),
-                "mps" => Ok(Device::Mps),
-                "npu" => Ok(Device::Npu(0)),
-                "xpu" => Ok(Device::Xpu(0)),
-                "xla" => Ok(Device::Xla(0)),
-                name if name.starts_with("cuda:") => parse_device(name).map(Device::Cuda),
-                name if name.starts_with("npu:") => parse_device(name).map(Device::Npu),
-                name if name.starts_with("xpu:") => parse_device(name).map(Device::Xpu),
-                name if name.starts_with("xla:") => parse_device(name).map(Device::Xla),
-                name => Err(SafetensorError::new_err(format!(
-                    "device {name} is invalid"
-                ))),
-            }
-        } else if let Ok(number) = ob.extract::<usize>() {
-            Ok(Device::Cuda(number))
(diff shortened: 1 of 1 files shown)


dvrogozh:
@Narsil : can you, please, help to review?

dvrogozh:
Rebased on top of latest main. @Narsil, @ArthurZucker, @SunMarc : can you, please, help to review?

dvrogozh:
Resolved conflict with https://github.com/huggingface/safetensors/commit/23319746e483012e6d3a7f2e4048cdbe7c458fa5. @Narsil @muellerzr @SunMarc : can this PR, please, be reviewed?

Narsil:
@dvrogozh Can you stop calling it a bug everywhere, since it's not a bug, it's breaking change you are proposing, that you introduced in torch==2.5.

The new behavior may be more user friendly on non cuda accelerators, it is nonetheless a breaking change and should be treated as such.

This PR introduces a dependency on torch itself since we would be dependent on torch to produce the correct string, therefore I cannot take the code as-is.

The raison d'être of this code is to provide simple validation so `safe_open(...., device="xxx")` is rejected with an appropriate error message (before doing work and sending the invalid values to torch.).

Narsil:
I made a cleaner implementation imho: https://github.com/huggingface/safetensors/pull/509

Can you check that it fixes your issue ? You're also more than welcome to steal the code from said PR so we can merge your PR instead of mine, so you get credit.

dvrogozh:
> I made a cleaner implementation imho: https://github.com/huggingface/safetensors/pull/509. Can you check that it fixes your issue ?

Unfortunately it does not. Fails with "RuntimeError: Invalid device string: '0'". See pytorch part of the stack at https://github.com/huggingface/safetensors/pull/509#pullrequestreview-2210418115

Narsil:
The logic you kept is the logic I want to get rid of. It's still wrong to depend on torch internals (here the string representation of resolved device) on that specific part.

dvrogozh:
> The logic you kept is the logic I want to get rid of. 

I am fine with this as soon as your change address the problem. I gave a try to modified version of #509, it works for me now. We can proceed with yours variant if you believe it's more aligned with the safetensors design.

Narsil:
Perfect done.

Thanks again for raising awareness about the upcoming new behavior !

Narsil:
Superseeded by #509

PR #504, opened 2024-07-25
Upgrading Pyo3 0.22 (with gil-refs)

Narsil:
# What does this PR do?

Upgrades to PyO3 0.22.
`gil-refs` still required because I couldn't find a way to get access to the `&[u8]` from the data pybuffer.
Without gil-refs, the data is created in `pydata` which is dropped at the end of the loop, therefore the reference doesn't live long enough.

I also tried changing everything around to use `FromPyObject` but it's not implemented for `&[u8]` as it would require `FromPyObjectBound` which is a sealed trait.

Additionally, the `SafetensorError` custom exception doesn't work with `gil-refs` because it doesn't implement the `HasPyGilRef`.
```
the trait `pyo3::PyNativeType` is not implemented for `SafetensorError`, which is required by `SafetensorError: HasPyGilRef`
```
Moving to just renaming `PyException` sort of seems to work without too much breaking change, although the exception is now the regular exception being effectively a catch all which is a breaking change.

### DIFF
--- a/bindings/python/src/lib.rs
+++ b/bindings/python/src/lib.rs
@@ -61,15 +61,6 @@ fn prepare(tensor_dict: HashMap<String, PyBound<PyDict>>) -> PyResult<HashMap<St
                 )));
             }
         };
-        // let shape = shape.ok_or_else(|| {
-        //     SafetensorError::new_err(format!("Missing `shape` in {tensor_desc:?}"))
-        // })?;
-        // let dtype = dtype.ok_or_else(|| {
-        //     SafetensorError::new_err(format!("Missing `dtype` in {tensor_desc:?}"))
-        // })?;
-        // let data = data.ok_or_else(|| {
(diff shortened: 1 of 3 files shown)


bindings/python/src/lib.rs
```diff
@@ -974,12 +973,12 @@ fn create_tensor<'a>(
         let dtype: PyObject = get_pydtype(module, dtype, is_numpy)?;
         let count: usize = shape.iter().product();
         let shape = shape.to_vec();
-        let shape: PyObject = shape.into_py(py);
         let tensor = if count == 0 {
             // Torch==1.10 does not allow frombuffer on empty buffers so we create
             // the tensor manually.
             // let zeros = module.getattr(intern!(py, "zeros"))?;
-            let args = (shape.clone(),);
+            let shape: PyObject = shape.clone().into_py(py);
```
Narsil, line 980: PyObject doesn'tseem to implement clone anymore therefore we're just cloning the raw Rust object.

PR #517, opened 2024-08-13
Use id rather than modelId 

osanseviero:
For context, modelId is getting deprecated https://github.com/huggingface/huggingface_hub/issues/2408


julien-c (approved):

julien-c (commented):
thx for merging

PR #518, opened 2024-08-19
fix lifetimes, add tensors iterator

gvilums:
# What does this PR do?

- Fixes the lifetimes of the `TensorView`s returned by the top-level `SafeTensors` object to depend on the underlying data instead of the `SafeTensors` object itself. This makes the API more flexible by allowing `TensorView`s to outlive the `SafeTensors` object that they were created from.
- Adds a method to `SafeTensors` to iterate over all contained `TensorView`s

### DIFF
--- a/safetensors/src/tensor.rs
+++ b/safetensors/src/tensor.rs
@@ -345,10 +345,10 @@ impl<'data> SafeTensors<'data> {
         Ok(Self { metadata, data })
     }
 
-    /// Allow the user to iterate over tensors within the SafeTensors.
+    /// Returns the tensors contained within the SafeTensors.
     /// The tensors returned are merely views and the data is not owned by this
     /// structure.
-    pub fn tensors(&self) -> Vec<(String, TensorView<'_>)> {
+    pub fn tensors(&self) -> Vec<(String, TensorView<'data>)> {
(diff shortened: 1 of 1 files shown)


galqiwi:
I just barely created the same pr, and found out that this one already exists :)
The lifetime part is very crucial for me -- to the extent that I'm planning to use your branch in my pet project.
Maintainers, please review this PR

Narsil:
Can you provide an actual example where this is impactful ?

I think adding a test showcasing the usefulness would help prevent future errors on this.

gvilums:
The basic pattern that this allows is the following:

```rust
        let serialized = b"...";

        let tensor = {
            let loaded = SafeTensors::deserialize(serialized).unwrap();
            loaded.tensor("test").unwrap()
        };
        
        do_something_with(tensor)
```

We want to get all the stored tensors without being tied to the lifetime of the `SafeTensors` object. This is of course a naive example - in our code we need this to be able to store our deserialized tensors without having to worry about keeping the `SafeTensors` object alive, which can get complicated because it itself borrows from the underlying data.

I added a test to check for this simple case.

Narsil (approved):
Thanks a lot for the tests/clarification. That makes it clearer and we can keep this for longer.

I'll release after this.

Issue #308, opened 2023-08-02
channels-last memory format


keturn:
### Feature request

Support `channels_last` memory format. (And probably any other pytorch dense [memory format](https://pytorch.org/docs/stable/tensor_attributes.html#torch.memory_format)), i.e. `channels_last_3d.`)

### Motivation

I was working with the [tiny-sd](https://huggingface.co/segmind/tiny-sd) model and it refused to save with the "You are trying to save a non contiguous tensor" error.

Upon closer examination, while these tensors did fail the `tensor.is_contiguous()` test, they _pass_ with `tensor.is_contiguous(memory_format=torch.channels_last)`.

### Your contribution

I might be able to help with this. But it smells like the sort of thing that requires some modification to how types are represented, and that's not something I want to blunder through without guidance.

Narsil:
Hmm `safetensors` doesn't have a concept of channels last or not. It's  a tensor of shape (N, C, H, W) or (N, H, W, C) but there are not semantics associated with it.

You can probably do `tensor.contiguous()` on your tensor.

`safetensors` is expecting to see C-format, little endian, formatted tensor. Anything else will be rejected.

I *think* `is_contiguous(memory_format=torch.channels_last) just means it's *not* contiguous, but it's ok because the shape (N, C, H, W) is a lie, and it's actually (N, H, W, C) and contiguous with that shape.
All pure guess, but you can probably  easily check.

We cannot do that in `safetensors` directly, because some non contiguous 4-rank tensors could exist and not be actual images.

keturn:
Yes, as far as being a storage for a bunch of bytes in-memory-order goes, the data file doesn't know about NCHW or NHWC. Same data type, same number of elements, etc.

But when it comes to `safetensors.torch.save_model` being an attractive replacement for the [standard `torch.save`](https://pytorch.org/docs/stable/generated/torch.save.html#torch.save), it would be helpful to support torch's memory-format semantics where we can.

You've thoroughly documented [reasons to not support shared tensors](https://huggingface.co/docs/safetensors/torch_shared_tensors#why-are-shared-tensors-not-saved-in-safetensors), but those reasons mostly don't apply to memory format. (With the probable exception of the "not all frameworks support this" caveat.)

This _could_ be done at the application layer, with the application sticking memory-format information in `metadata` and wrapping their load function to handle that -- but since weights are often trained and used in different applications, it would be best for consistency if it were handled at the library level, as `torch.save` does.

Narsil:
[...]

PR #530, opened 2024-09-25
fix(save_file): add force_contiguous(default=True) to save_file

dbyoung18:
# What does this PR do?

Refering the inferface of [`save_model`](https://github.com/huggingface/safetensors/blob/main/bindings/python/py_src/safetensors/torch.py#L131), add `force_contiguous`(default=True) to force the state_dict to be saved as contiguous tensors, and finally into raw bytes in safetensors format.

The fix could contribute such scenario, when running workloads w/ channels_last format in PyTorch, but need finally saved to .safetensor by calling [transformers::safe_save_file](https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_utils.py#L2824). Meanwhile, it could also align the semantics between `save_model` & `save_file`.

A related issue encountered such scenario: https://github.com/huggingface/safetensors/issues/308

### DIFF
--- a/bindings/python/py_src/safetensors/torch.py
+++ b/bindings/python/py_src/safetensors/torch.py
@@ -256,6 +256,7 @@ def save_file(
     tensors: Dict[str, torch.Tensor],
     filename: Union[str, os.PathLike],
     metadata: Optional[Dict[str, str]] = None,
+    force_contiguous: Optional[bool] = True,
 ):
     """
     Saves a dictionary of tensors into raw bytes in safetensors format.
@@ -269,6 +270,11 @@ def save_file(
             Optional text only metadata you might want to save in your header.
             For instance it can be useful to specify more about the underlying
             tensors. This is purely informative and does not affect tensor loading.
+        force_contiguous (`boolean`, *optional*, defaults to True):
+            Forcing the state_dict to be saved as contiguous tensors.
+            This has no effect on the correctness of the model, but it
+            could potentially change performance if the layout of the tensor
+            was chosen specifically for that reason.
 
     Returns:
         `None`
@@ -283,6 +289,8 @@ def save_file(
     save_file(tensors, "model.safetensors")
     ```
     """
+    if force_contiguous:
+        tensors = {k: v.contiguous() for k, v in tensors.items()}
     serialize_file(_flatten(tensors), filename, metadata=metadata)
 
 


dbyoung18:
@Narsil Could you help review the PR？

yao-matrix:
@ArthurZucker , hi, could you help take a look at this issue, thx very much.
The thing is channel_last format is a supported format in pytorch and adopted by intel xpu for better performance. So, when we run training/finetuning and save the checkpoint, transformers will call `safe_save_file` and finally to safetensors's `save_file` which not support `force_contiguous` yet(but `save_model` already). This will not impact the prior behavior, it just add non_contiguous format format.

Thx.

Narsil:
No, this option will not be added.
`safetensors` is very explicit in not allowing this. Users should do this on their own as making tensors contiguous should be an express intent by the user. `save_file` is extremely strict on purpose.

The reason is that there are valid use cases where the tensors *must not* be contiguous on the model, therefore they *must not* be casted to contiguous automatically during save(even with a choice on user's behalf). Doing so would break those models during reload (or the speed performance, or something else).
The use case is typically a channel location format. Safetensor doesn't hold semantic information about shapes, therefore it's impossible to recover it after being saved.

`save_model`/`load_model` follow different semantics because the layout and shape semantics are contained in the model, therefore those operations can occur be uncasted/recasted with those interfaces. Both functions exists purely because not everyone uses `transformers`.

If users want to force contiguous tensors, they must explicitly do so after getting their model's state.

It's up to `transformers` to hold those in `save_pretrained`/`from_pretrained` to follow similar patterns as `save_model`/`load_model` if it needs to be, but not this function.

dbyoung18:
> No, this option will not be added. `safetensors` is very explicit in not allowing this. Users should do this on their own as making tensors contiguous should be an express intent by the user. `save_file` is extremely strict on purpose.
> 
> The reason is that there are valid use cases where the tensors _must not_ be contiguous on the model, therefore they _must not_ be casted to contiguous automatically during save(even with a choice on user's behalf). Doing so would break those models during reload (or the speed performance, or something else). The use case is typically a channel location format. Safetensor doesn't hold semantic information about shapes, therefore it's impossible to recover it after being saved.
> 
> `save_model`/`load_model` follow different semantics because the layout and shape semantics are contained in the model, therefore those operations can occur be uncasted/recasted with those interfaces. Both functions exists purely because not everyone uses `transformers`.
> 
> If users want to force contiguous tensors, they must explicitly do so after getting their model's state.
> 
> It's up to `transformers` to hold those in `save_pretrained`/`from_pretrained` to follow similar patterns as `save_model`/`load_model` if it needs to be, but not this function.

Make sense to me. I will close the PR & take ur suggestion to fix the case on other place. THX for detailed explanation:)

PR #557, opened 2025-01-07
Fix wrong signature of `safe_open.__init__`  in stub file

SunghwanShim:
# What does this PR do?

First argument of `safe_open.__init__` (`self`) is omitted. So add it.


Narsil:
Thanks.

I don't write those files manually, they are generated via a script. Seems like PyO3 changed something in the generated Python that I overlooked. I fixed the script too now.

Narsil (approved):

Issue #544, opened 2024-12-02
Support no_std


ivila:
### Feature request

make safetensors crate support no_std feature.

### Motivation

Safetensors is used by some other crates, and some of them support no_std feature, for example, the [dfdx](https://github.com/coreylowman/dfdx.git), however, currently safetensors do not support no_std, so people cannot use dfdx to export model at no_std environment.

### Your contribution

Currently I make a patch in my own git repository, adding a no_std feature and:
1) when enable no_std, swtich from HashMap to hashbrown::HashMap
2) change some types or macros from std to alloc: 
    1. `String => alloc::string::String`
    2. `format! => alloc::format!`
    3. `Vec => alloc::vec::Vec`
    4. `vec! => alloc::vec!`
    5. `std::borrow::Cow => alloc::borrow::Cow`
3) change some traits or types from std to core: 
    1. `std::ops => core::ops`
    2. `std::fmt::Display => core::fmt::Display`
    3. `std::error::Error => core::error::Error`
    4. `std::str => core::str`
4) switch serde and serde_json to no_std
5) when enable, remove some dependencies of std:
    1. remove the IoError in std::io::Error
    2. remove methods that loads from or writes to file (`serialize_to_file`)

I would like to create a PR if you allow.

Narsil:
Sure that would be fine by me.

What's the target you're using where `std` isn't available ?
It's usually easier to motivate changes when there's a clear issue to fix.

ivila:
> Sure that would be fine by me.
> 
> What's the target you're using where `std` isn't available ? It's usually easier to motivate changes when there's a clear issue to fix.

I am trying machine learning in TrustZone (by using [dfdx](https://github.com/coreylowman/dfdx) and OP-TEE), and in OP-TEE there is no std (no file, no network, just alloc).

PR #563, opened 2025-01-24
Restore compatibility with Rust 1.74

tiran:
# What does this PR do?

PR #544 added support for `no_std` feature. The PR changed `std::error::Error` to `core::error::Error`. The `core::error` trait was stabilized in Rust 1.81, so the change bumped MSRV to 1.81. Before the Python package built with Rust 1.66 and the `safetensors` create with all features built with 1.74.

This commit restores compatibility with Rust 1.74 for `std` builds:

- `mixed_integer_ops` feature requires 1.66
- `half v2.4.1` requires 1.70
- `clap_lex v0.7.4` requires 1.74

I'm also adding `rust-version` to `Cargo.toml`, so cargo creates a backwards compatible `Cargo.lock`. By default, Cargo >= 1.83 creates a `v4` lock file, which is not compatible with Cargo < 1.78.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -4,11 +4,15 @@ on:
   pull_request:
 
 jobs:
-  build-stable:
+  build:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
+        toolchain: [stable]
+        include:
+          - os: ubuntu-latest
+            toolchain: "1.74"
     defaults:
       run:
         working-directory: ./safetensors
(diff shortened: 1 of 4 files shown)


Narsil:
Do you mind explaining a bit more the where and what of the issue (what distro, build system etc ?) ?

I'm curious why cargo new doesn't define `rust-version` as a default in Cargo.toml directly.
Thanks for the fix.

Narsil:
I updated the github action to unify both locations for testing.

tiran:
@Narsil Thanks for merging my PR!

Sure, I'm happy to tell you a bit more about our work. I'm a member of Red Hat's AI platform core components team. We are providing AI stacks for several AI accelerators (CUDA, Gaudi, ROCm, ...) for other teams in Red Hat. As part of our workflow, we are rebuilding all Python wheels from sources and with system packages. This is done for security and compliance reasons (e.g. SBOM), but also to build packages with special settings, optimizations, or to support more hardware.

Python wheels are built in our [Fromager](https://github.com/python-wheel-build/fromager) pipeline on RHEL 9. We have created Fromager to automate  re-building a dependency tree from sdists. The tool also does some extra steps to create self-contained source distributions. For Rust-based packages (maturin, setuptools-rust), it `cargo vendor`s all crates into an sdist, too.

We ran into problems with `Cargo.lock` v4 when Rust 1.83 made `v4` the default for projects without lower MSRV. For compatibility with hardware vendors, we have to support even RHEL y-stream releases with extended update support (e.g. 9.4) and cannot update to 9.5. RHEL 9.4 has with Rust 1.75, which does not understand `Cargo.lock` v4. I honestly don't know why `cargo new` doesn't add MSRV automatically...

Our build system has the capability to apply patches and we could just fix our build problem in a downstream patch. However we prefer to contribute our fixes to upstream, so upstream and other downstream rebuilders benefit from our work, too. Safetensors is the third project I fixed. The `blake3` maintainer has merged my MSRV change, too. My `hf_transfer` PR is waiting for review.

Feel free to ping me if you have more questions. I'm planning to write a longer explanation around downstream packaging and MSRV for Python packages with Rust extensions, too.

Issue #560, opened 2025-01-14
Thread panicked in slice.rs due to out-of-range index in safetensors


csaybar:
### System Info

transformers version: 4.42.3
Platform: Linux-6.8.0-51-generic-x86_64-with-glibc2.35
Python version: 3.10.12
Huggingface_hub version: 0.25.2
Safetensors version: 0.5.2
Accelerate version: 0.27.0
Accelerate config: not found
PyTorch version (GPU?): 2.3.1+cu121 (True)
Tensorflow version (GPU?): 2.15.0 (True)
Flax version (CPU?/GPU?/TPU?): not installed (NA)
Jax version: not installed
JaxLib version: not installed
Using distributed or parallel set-up in script?:
Using GPU in script?:
GPU type: NVIDIA GeForce RTX 3050 Laptop GPU

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### Reproduction

A panic occurred in `safetensors` while attempting to access a slice with an out-of-range index. 
The following error was raised:

> thread '<unnamed>' panicked at /home/runner/work/safetensors/safetensors/safetensors/src/slice.rs:354:31: range end index 458752000 out of range for slice of length 13762560 Traceback (most recent call last): File "<stdin>", line 2, in <module> pyo3_runtime.PanicException: range end index 458752000 out of range for slice of length 13762560

```python
import safetensors.numpy
import safetensors
toy_data = np.random.rand(3, 5, 7, 128, 128)
safetensors.numpy.save_file({"data": toy_data}, "model.safetensors")
with safetensors.safe_open("model.safetensors", "np") as tensor:
    tensor.get_slice("data")[0:128, 0:128]
```

### Expected behavior

The library should handle out-of-range indices gracefully without panicking.

oliness:
Can you see if [this](https://github.com/huggingface/safetensors/pull/565) pr fixes it ok?

Narsil:
Fixed in  https://github.com/huggingface/safetensors/pull/565

PR #565, opened 2025-01-29
Return error on out of range index

oliness:
Fix issue #560, return a SliceOutOfRange error if the stop value exceeds the available span.


Narsil:
Thanks for the PR.
I updated the fix, since the condition already existed, but was faulty.

PR #578, opened 2025-02-25
support hpu:0

IlyasMoutawwakil:
# What does this PR do?

Because some libs load tensors internally with `safe_load(name, device=str(torch_device))` and `str(torch_device)` will be `hpu:0` if `torch_device` is indexed


regisss (approved):
LGTM!

Narsil (approved):

PR #581, opened 2025-02-28
Add onnx-safetensors to the projects list in documentation

justinchuby:
# What does this PR do?

onnx-safetensors enables onnx files to use safetensors as external data for the ONNX model natively. This change adds onnx-safetensors to the projects list for users to discover support for using safetensors in onnx models.


Narsil (approved):
LGTM ! Thanks a lot !

PR #583, opened 2025-02-28
Always call .to(device) if it was passed as an argument

### LINKED ISSUE #582: load_file does apply to(device) when device = "cpu"
### System Info

latest

### Information

- [ ] My own modified scripts

### Reproduction

I think the [code here](https://github.com/huggingface/safetensors/blob/543243c3017e413584f27ebd4b99c844f62deb34/bindings/python/src/lib.rs#L553-L557) omits calling `tensor.to(device)` when the device is `"cpu"`. However, if you've called `torch.set_default_device("something other than cpu")` the tensor will be loaded on that device instead. I think the device argument should be optional, and remain optional throughout the code base and then always apply `.to(device)` when it is not `None`. 

### Expected behavior

tensor is loaded on cpu when calling `load_model(..., device="cpu")`

mickvangelderen:
This fixes https://github.com/huggingface/safetensors/issues/582 where if setting torch.set_default_device('not cpu'), the device for the tensor ends up being the default device rather than the passed device.

# What does this PR do?

Fixes #582 

NOTE: I could not find any development instructions in the repository after a brief search. This PR would be better with a test case that shows the problem and then another commit that fixes it but I am unfamiliar with maturin. Feel free to provide instructions or take my work and finish it.

### DIFF
--- a/bindings/python/src/lib.rs
+++ b/bindings/python/src/lib.rs
@@ -193,7 +193,7 @@ impl<'source> FromPyObject<'source> for Framework {
     }
 }
 
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 enum Device {
     Cpu,
     Cuda(usize),
@@ -326,7 +326,7 @@ struct Open {
     metadata: Metadata,
     offset: usize,
     framework: Framework,
-    device: Device,
+    device: Option<Device>,
     storage: Arc<Storage>,
 }
 
@@ -335,12 +335,13 @@ impl Open {
         let file = File::open(&filename).map_err(|_| {
             PyFileNotFoundError::new_err(format!("No such file or directory: {filename:?}"))
         })?;
-        let device = device.unwrap_or(Device::Cpu);
 
-        if device != Device::Cpu && framework != Framework::Pytorch {
-            return Err(SafetensorError::new_err(format!(
-                "Device {device:?} is not support for framework {framework:?}",
-            )));
+        if let Some(device) = device {
+            if framework != Framework::Pytorch {
+                return Err(SafetensorError::new_err(format!(
+                    "Device {device:?} is not support for framework {framework:?}",
+                )));
+            }
         }
 
         // SAFETY: Mmap is used to prevent allocating in Rust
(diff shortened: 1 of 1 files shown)


Narsil:
No, device is mandatory.

Again this is a design decision, location for where to load tensors is not something that should be ever implicit as it has implications about speed/tradeoffs and memory requirements.

If you really want to follow `torch.set_default_device()` (which I personally do not recommend) then just use https://pytorch.org/docs/stable/generated/torch.get_default_device.html during the load.

mickvangelderen:
Hey Narsil, thanks for the response. 

By "device is mandatory", where and how is it mandatory? It is not mandatory to provide the `device` argument when calling `safe_load` for example, where it defaults to `"cpu"`. 

I do not have full control over the code base that I am working with, and the default device is being set to "cuda". I need to load the model on the CPU and move only some of it to the GPU. It is confusing that the `safe_load` documentation claims to load the tensors on the CPU by default, but instead just omits calling `.to("cpu")`. This PR makes safetensors more predictable, in that it simply calls `.to(device)` if you provide the `device` argument.

Narsil:
I know and I agree with the principle (I followed that discussion in the issue instead).

The default to CPU that currently exists is really a python construct (in python it's very natural to leave defaults to arguments).
However it's being lowered as soon as possible (I think with recent PyO3 versions we could even have a real default argument directly).

My main issue lies in the silently moving tensors over and over that will likely happen with fixing this bug like in this PR.
I looked at the issue again in torch and couldn't really find a satistiable solution to the default (as torch doesn't seem to provide a consistent way to detect what's the current allocation device).

I've thought that over, and maybe the only escape hatch in the meantime would be to do what you are doing but add a warning when we're moving back something on CPU which was already allocated on some other device (because of some distant global). It seems something worthy of concern that should be adressed by users.

Just to double down, I've seen issues where users have started to rely on the current behavior to allocate on specific devices directly. So modifying the current behavior will be a breaking change for sure, so ideally we make a single breaking change with a clean solution afterwards (and with the silent casting, we will receive issues a few months down the line saying that the allocation is too slow because of the silent CPU casting).

CPU *is* special because of the memory mapping, the kernel will skip making any copies when giving access to those tensors making it extremely fast operation to load things.

Narsil:
Edit, discussed with maintainers, the device location inconsistency is apparently a bug: https://github.com/pytorch/pytorch/pull/148621

So maybe having a default following torch for torch (and only torch) is possible.

Narsil:
superseeded by https://github.com/huggingface/safetensors/pull/588

PR #586, opened 2025-03-05
Do not error with InvalidOffset if multiple tensors are tied to the exact same data

karl3wm:
This changes the conditions in which an `InvalidOffset` is thrown such that it is accepted for multiple tensor names in the header to refer to the same data.

This lets files constructed to represent tied weights load successfully.

I am new to Rust and encountered this error while writing my own safetensors encoder at https://huggingface.co/datasets/baffo32/llm_logits/blob/main/_safetensors.py . I added a hashing function to deduplicate tensors, but then learned that the official safetensors implementation throws an error for such files.

This is a very simple change that only changes what file structures are valid but does not otherwise add support for tied weights. It might be reasonable to issue a warning if tied weights are encountered in a file, as they would be duplicated if naively uploaded to a GPU separately.

### DIFF
--- a/safetensors/src/tensor.rs
+++ b/safetensors/src/tensor.rs
@@ -512,18 +512,18 @@ impl Metadata {
     }
 
     fn validate(&self) -> Result<usize, SafeTensorError> {
-        let mut start = 0;
+        let (mut prev, mut start) = (0, 0);
         for (i, info) in self.tensors.iter().enumerate() {
-            let (s, e) = info.data_offsets;
-            if s != start || e < s {
                 let tensor_name = self
                     .index_map
                     .iter()
                     .find_map(|(name, &index)| if index == i { Some(&name[..]) } else { None })
                     .unwrap_or("no_tensor");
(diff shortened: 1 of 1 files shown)


Narsil:
This is a won't fix. Safetensors can only allow single name to "own" tensor data location. Every tensor data location *must* have an owner, and a single owner.

This is by design because it will cause many issues otherwise. If weights are tied, then there are the same weights, just use the same name (`safetensors.torch.load_model` will handle this transparently for you too since it's a torch only issue.).

Narsil:
Closing as won't fix. Feel free to reopen if you think I missed something.

karl3wm:
So you know, I found significant reduction in storage usage and network transfer with no performance loss nor need for architectural or configuration changes when deduplicating weights using this change.

I haven't thought of any issues arising from this other than losing the benefits if used naively, and I'm curious what issues you're considering.

But I recognize this is a poor pull request that implies a potential design change without discussion nor changes to other code, docs, and testcases. It's appropriate to close it.

Thanks for your review.

PR #588, opened 2025-03-14
Pass device to torch.asarray in get_tensor

### LINKED ISSUE #582: load_file does apply to(device) when device = "cpu"

mickvangelderen:
# What does this PR do?

Fixes #582

### DIFF
--- a/bindings/python/src/lib.rs
+++ b/bindings/python/src/lib.rs
@@ -514,14 +514,6 @@ impl Open {
                     let sys = PyModule::import(py, intern!(py, "sys"))?;
                     let byteorder: String = sys.getattr(intern!(py, "byteorder"))?.extract()?;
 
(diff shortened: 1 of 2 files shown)


Narsil (approved):
LGTM

PR #594, opened 2025-03-20
fix typo in serialize_file doc-string

LVivona:
# What does this PR do?

function ``serialize_file`` within `binding/python/src/lib.rs` contains a doc-string typo 

### Now
```rust
/// Serializes raw data into file.
///
/// Args:
///     tensor_dict (`Dict[str, Dict[Any]]`):
///         The tensor dict is like:
///             {"tensor_name": {"dtype": "F32", "shape": [2, 3], "data": b"\0\0"}}
///     filename (`str`, or `os.PathLike`):
///         The name of the file to write into.
///     metadata (`Dict[str, str]`, *optional*):
///         The optional purely text annotations
///
/// Returns:
///     (`NoneType`):
///         On success return None
```

### Previous
```rust
/// Serializes raw data.
///
/// Args:
///     tensor_dict (`Dict[str, Dict[Any]]`):
///         The tensor dict is like:
///             {"tensor_name": {"dtype": "F32", "shape": [2, 3], "data": b"\0\0"}}
///     filename (`str`, or `os.PathLike`):
///         The name of the file to write into.
///     metadata (`Dict[str, str]`, *optional*):
///         The optional purely text annotations
///
/// Returns:
///     (`bytes`):
///         The serialized content.
```

### DIFF
--- a/bindings/python/py_src/safetensors/__init__.pyi
+++ b/bindings/python/py_src/safetensors/__init__.pyi
@@ -36,7 +36,7 @@ def serialize(tensor_dict, metadata=None):
 @staticmethod
 def serialize_file(tensor_dict, filename, metadata=None):
     """
-    Serializes raw data.
+    Serializes raw data into file.
 
     Args:
         tensor_dict (`Dict[str, Dict[Any]]`):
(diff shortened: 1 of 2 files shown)


daskol (changes requested):
Could you fix `tensor_dict` example too. It uses string `"F32"` dtype while `prepare()` routine actually infers dtype from fully-qualified string `"float32"`.

Narsil (approved):

PR #599, opened 2025-04-13
Ensure `ml_dtypes` imported before dtype resolution

### LINKED ISSUE #598: SafeTensors does not load JAX arrays of custom floats
### System Info

```
safetensors 0.5.3
Python 3.13.2
Linux hostname 6.14.2-arch1-1 #1 SMP PREEMPT_DYNAMIC Thu, 10 Apr 2025 18:43:59 +0000 x86_64 GNU/Linux
```

### Information

- [x] My own modified scripts

### Reproduction

The issue occurs with direct loading JAX/NumPy arrays from bfloat16 checkpoints (like [OLMOE][1]).

```python

st: safe_open
with safe_open(ns.model, 'jax', 'cpu') as st:
    arr = st.get_tensor('some-bfloat16-tensor')
# TypeError: data type 'bfloat16' not understood
```

[1]: https://huggingface.co/allenai/OLMoE-1B-7B-0924

### Expected behavior

Bfloat16 JAX array must be restored from checkpoint.

daskol:
Resolve #598.

### DIFF
--- a/bindings/python/pyproject.toml
+++ b/bindings/python/pyproject.toml
@@ -37,6 +37,7 @@ torch = [
     "torch>=1.10",
 ]
 tensorflow = [
+    "ml-dtypes",
     "safetensors[numpy]",
     "tensorflow>=2.11.0",
 ]
@@ -46,10 +47,10 @@ pinned-tf = [
     "tensorflow==2.18.0",
 ]
 jax = [
+    "ml-dtypes",
     "safetensors[numpy]",
     "flax>=0.6.3",
     "jax>=0.3.25",
-    "jaxlib>=0.3.25",
 ]
 mlx = [
     "mlx>=0.0.9",
--- a/bindings/python/src/lib.rs
+++ b/bindings/python/src/lib.rs
@@ -1069,6 +1069,18 @@ fn create_tensor<'a>(
                 true,
             ),
         };
+
+        // Ensure `ml_dtypes` package is imported and it annanounced custom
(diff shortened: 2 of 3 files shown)


Narsil:
Thanks for the issue + PR !

I went with another kind of fix which doesn't add the dependency (in case things get modified in the future in those packages).

https://github.com/huggingface/safetensors/pull/605
Importing the actual library seems enough (and is necessary anyway).

Also the unloading of the module doesn't seem to work, the tests pass regardless of the fix, I couldn't find any reliable way to reproduce in the test suite.

I added you as co-author too.

daskol:
Thank you for fixing the issue. However, it does something different in my use case. I'm working on weight quantization, so I basically need only access to weights in framework-free way (e.g. `jax.Array` and `numpy.ndarray` is basically enough). However, now I get `flax` loaded. 😅

> Importing the actual library seems enough (and is necessary anyway).

Importing `flax` is too much. It would be better to import `jax.numpy` instead of `flax` because `framework="flax"` is actually `jax`. Also, it is unclear what is framework in case of `nnx`?

> I went with another kind of fix which doesn't add the dependency (in case things get modified in the future in those packages).

This dependency on `ml-dtypes` does already present because of transitivity. The `ml-dtypes` library is designed and implemented as a standalone extension to NumPy. It is actually the best way to expose reduced precision dtypes for NumPy and TF/JAX.

As far as I understand, your concern is that the explicit use of transitive dependency does not ensure forward compatibility. At the same time, you do use explicitly transitive dependency `jax` of `flax`. Then what is the point?

PR #612, opened 2025-06-03
Make API more idiomatic, simplify code, improve style

H2CO3:
# What does this PR do?

There are a number of issues with the current code style, for example:

- The API sometimes forces sub-optimal choices (e.g. the use of `&Option<T>` forces the caller to have an owned `T`, whereas `Option<&T>` doesn't, and can be obtained from an `Option<T>` via `Option::<T>::as_ref()`)
- The implementation was sometimes overly verbose (e.g. heavily nested `if let` "triangles of hell", which can be replaced by `let … else` and early returns)
- There have been some notable omissions (e.g. the `source()` method on `impl Error for SafeTensorsError`)
- Some dependencies (notably, Criterion) were out-of-date and the code used deprecated functions

This PR fixes the aforementioned stylistic and API issues. Note that some of these _are_ technically breaking change, but since the library is still relatively young, and hasn't reached 1.0 yet, this should not be a problem.

### DIFF
--- a/safetensors/Cargo.toml
+++ b/safetensors/Cargo.toml
@@ -2,7 +2,7 @@
 name = "safetensors"
 version = "0.5.3-dev.0"
 edition = "2021"
-rust-version = "1.74"
+rust-version = "1.80"
 homepage = "https://github.com/huggingface/safetensors"
 repository = "https://github.com/huggingface/safetensors"
 documentation = "https://docs.rs/safetensors/"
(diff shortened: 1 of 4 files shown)


H2CO3:
@Narsil I believe this PR is now complete.

H2CO3:
> I see a few things slip which are not exactly obvious and making sure none slips review is hard.
> Keep
>
>    `write!(f, "{}...")`
>
> everywhere. It's just easier to read when code is uniform.
> §index.fmt(..)§-> Same comment.
> Those are the first ones I saw.

Oh, that's definitely intentional, because the formatting mechanism is slow, so it's unnecessarily wasteful to `write!()` when there's nothing to format. A great example is all-unit variant `enum`s such as `Framework`, which always write a constant string, so there's exactly one single call to `Formatter::write_str()` needed.

> Ah and another one, remove the Into<Vec<usize>>. I tried to avoid accepting traits anywhere not strictly necessary, I think it is cleaner to avoid the generic bloats everywhere.

Okay, I can do that.

H2CO3:
@Narsil I have split this into 4 separate PRs, split by the "theme" of the PR:

1. https://github.com/huggingface/safetensors/pull/614
2. https://github.com/huggingface/safetensors/pull/615
3. https://github.com/huggingface/safetensors/pull/616
4. https://github.com/huggingface/safetensors/pull/617

I hope this is now acceptable, and you can review and hopefully merge them.

PR #615, opened 2025-06-12
Simplify code and make it more robust

H2CO3:
# What does this PR do?

Make code more idiomatic, slightly more performant, less prone to some programming errors, and in other cases, simpler, using a couple of small changes:

- remove some unnecessary parentheses
- remove some unnecessary allocations (such as an identity vec-to-vec transform of the form `.into_iter().collect()`)
- replace manual impl of `next_multiple_of()` with the actual std function
- use `serde_json::to_vec()` instead of `to_string().into_bytes()`
- clean up some generic signatures
- extract the often-repeated magic number `8` into a named constant, `HEADER_SIZE_LEN`
- remove some redundant lifetime parameters and explicit annotations
- be explicit in comparing the size (and alignment) of data types when sorting tensors
- replace double negatives of the form `if negative_condition { Err(…) } else { Ok(…) }` with the condition un-negated and the branches flipped
- remove deep nesting and use early returns in some cases instead
- replace some instances of explicit size checking logic
- eta-reduce unnecessary closures: `|f| f.to_le_bytes()` -> `f32::to_le_bytes`
-

### DIFF
--- a/safetensors/src/tensor.rs
+++ b/safetensors/src/tensor.rs
@@ -2,10 +2,12 @@
 use crate::lib::{Cow, HashMap, String, ToString, Vec};
 use crate::slice::{InvalidSlice, SliceIterator, TensorIndexer};
 use serde::{ser::SerializeMap, Deserialize, Deserializer, Serialize, Serializer};
+use core::fmt::Display;
 #[cfg(feature = "std")]
 use std::io::Write;
 
 const MAX_HEADER_SIZE: usize = 100_000_000;
+const HEADER_SIZE_LEN: usize = size_of::<u64>();
 
 /// Possible errors that could occur while reading
 /// A Safetensor file.
@@ -168,22 +170,26 @@ pub trait View {
     fn data_len(&self) -> usize;
 }
 
-fn prepare<S: AsRef<str> + Ord + core::fmt::Display, V: View, I: IntoIterator<Item = (S, V)>>(
+fn prepare<S, V, I>(
     data: I,
     data_info: &Option<HashMap<String, String>>,
-    // ) -> Result<(Metadata, Vec<&'hash TensorView<'data>>, usize), SafeTensorError> {
-) -> Result<(PreparedData, Vec<V>), SafeTensorError> {
+) -> Result<(PreparedData, Vec<V>), SafeTensorError>
+where
+    S: AsRef<str> + Ord + Display,
+    V: View,
+    I: IntoIterator<Item = (S, V)>,
+{
     // Make sure we're sorting by descending dtype alignment
     // Then by name
     let mut data: Vec<_> = data.into_iter().collect();
     data.sort_by(|(lname, left), (rname, right)| {
-        right.dtype().cmp(&left.dtype()).then(lname.cmp(rname))
+        right.dtype().size().cmp(&left.dtype().size()).then(lname.cmp(rname))
     });
 
     let mut tensors: Vec<V> = Vec::with_capacity(data.len());
     let mut hmetadata = Vec::with_capacity(data.len());
     let mut offset = 0;
-    let data: Vec<_> = data.into_iter().collect();
+
     for (name, tensor) in data {
         let n = tensor.data_len();
         let tensor_info = TensorInfo {
@@ -197,16 +203,15 @@ fn prepare<S: AsRef<str> + Ord + core::fmt::Display, V: View, I: IntoIterator<It
     }
 
     let metadata: Metadata = Metadata::new(data_info.clone(), hmetadata)?;
-    let mut metadata_buf = serde_json::to_string(&metadata)?.into_bytes();
-    // Force alignment to 8 bytes.
-    let extra = (8 - metadata_buf.len() % 8) % 8;
-    metadata_buf.extend(vec![b' '; extra]);
+    let mut metadata_buf = serde_json::to_vec(&metadata)?;
 
-    let n: u64 = metadata_buf.len() as u64;
+    // Force alignment to 8 bytes.
+    let aligned_metadata_len = metadata_buf.len().next_multiple_of(HEADER_SIZE_LEN);
+    metadata_buf.resize(aligned_metadata_len, b' ');
 
     Ok((
         PreparedData {
-            n,
+            n: aligned_metadata_len as u64,
             header_bytes: metadata_buf,
             offset,
         },
@@ -231,13 +236,16 @@ pub fn serialize<
         },
         tensors,
     ) = prepare(data, data_info)?;
-    let expected_size = 8 + header_bytes.len() + offset;
+
+    let expected_size = HEADER_SIZE_LEN + header_bytes.len() + offset;
     let mut buffer: Vec<u8> = Vec::with_capacity(expected_size);
-    buffer.extend(&n.to_le_bytes().to_vec());
-    buffer.extend(&header_bytes);
+    buffer.extend(n.to_le_bytes());
+    buffer.extend(header_bytes);
+
     for tensor in tensors {
         buffer.extend(tensor.data().as_ref());
     }
+
     Ok(buffer)
 }
 
@@ -245,28 +253,33 @@ pub fn serialize<
 /// Writing directly to file reduces the need to allocate the whole amount to
 /// memory.
 #[cfg(feature = "std")]
-pub fn serialize_to_file<
-    S: AsRef<str> + Ord + core::fmt::Display,
-    V: View,
-    I: IntoIterator<Item = (S, V)>,
->(
+pub fn serialize_to_file<S, V, I>(
     data: I,
     data_info: &Option<HashMap<String, String>>,
     filename: &std::path::Path,
-) -> Result<(), SafeTensorError> {
+) -> Result<(), SafeTensorError>
+where
+    S: AsRef<str> + Ord + Display,
+    V: View,
+    I: IntoIterator<Item = (S, V)>,
+{
     let (
         PreparedData {
             n, header_bytes, ..
         },
         tensors,
     ) = prepare(data, data_info)?;
+
     let mut f = std::io::BufWriter::new(std::fs::File::create(filename)?);
     f.write_all(n.to_le_bytes().as_ref())?;
     f.write_all(&header_bytes)?;
+
     for tensor in tensors {
         f.write_all(tensor.data().as_ref())?;
     }
+
     f.flush()?;
+
     Ok(())
 }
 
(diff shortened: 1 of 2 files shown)


safetensors/src/tensor.rs
```diff
@@ -168,22 +170,26 @@ pub trait View {
...
+    S: AsRef<str> + Ord + Display,
+    V: View,
+    I: IntoIterator<Item = (S, V)>,
+{
     // Make sure we're sorting by descending dtype alignment
     // Then by name
     let mut data: Vec<_> = data.into_iter().collect();
     data.sort_by(|(lname, left), (rname, right)| {
-        right.dtype().cmp(&left.dtype()).then(lname.cmp(rname))
+        right.dtype().size().cmp(&left.dtype().size()).then(lname.cmp(rname))
```
Narsil: This one is wrong, we really want to sort by dtype order consistently, sorting by size only makes it unstable because many dtypes of have the same dtype right ?

safetensors/src/tensor.rs
```diff
@@ -2,10 +2,12 @@
 use crate::lib::{Cow, HashMap, String, ToString, Vec};
 use crate::slice::{InvalidSlice, SliceIterator, TensorIndexer};
 use serde::{ser::SerializeMap, Deserialize, Deserializer, Serialize, Serializer};
+use core::fmt::Display;
 #[cfg(feature = "std")]
 use std::io::Write;
 
 const MAX_HEADER_SIZE: usize = 100_000_000;
+const HEADER_SIZE_LEN: usize = size_of::<u64>();
```
Narsil: This is not really HEADER_SIZE_LEN, this is just 8 bytes, this is the size of the u64 binary value that contains the header size.

For now I'd call it `N_LEN` for lack of a better name, but anything HEADER is related to the JSON part, so we need something that doesn't use that word I think.

safetensors/src/tensor.rs
```diff
@@ -197,16 +203,15 @@ fn prepare<S: AsRef<str> + Ord + core::fmt::Display, V: View, I: IntoIterator<It
     }
 
     let metadata: Metadata = Metadata::new(data_info.clone(), hmetadata)?;
-    let mut metadata_buf = serde_json::to_string(&metadata)?.into_bytes();
-    // Force alignment to 8 bytes.
-    let extra = (8 - metadata_buf.len() % 8) % 8;
-    metadata_buf.extend(vec![b' '; extra]);
+    let mut metadata_buf = serde_json::to_vec(&metadata)?;
```
Narsil: The JSON should contain only valid utf-8 (in safetensors spec, that makes things so much simpler for other implementors), so this feels a bit wrong to me. JSON shouldn't contain binary data, but it's not necessarily utf-8. https://www.ietf.org/rfc/rfc4627.txt

safetensors/src/tensor.rs
```diff
@@ -281,47 +294,49 @@ pub struct SafeTensors<'data> {
...
+        let Some(header_size_bytes) = buffer.get(..HEADER_SIZE_LEN) else {
             return Err(SafeTensorError::HeaderTooSmall);
-        }
-        let arr: [u8; 8] = [
-            buffer[0], buffer[1], buffer[2], buffer[3], buffer[4], buffer[5], buffer[6], buffer[7],
-        ];
+        };
+        let arr: [u8; HEADER_SIZE_LEN] = header_size_bytes
+            .try_into()
+            .map_err(|_| SafeTensorError::InvalidHeaderLength)?;
```
Narsil: is there any benefit beyond style preference ?

Since we're checking already the buffer len, I don't see how this array creation can fail.

Is InvalidHeaderLength the most appropriate error here ?

safetensors/src/tensor.rs
```diff
@@ -281,47 +294,49 @@ pub struct SafeTensors<'data> {
...
             return Err(SafeTensorError::HeaderTooLarge);
         }
 
         let stop = n
-            .checked_add(8)
+            .checked_add(HEADER_SIZE_LEN)
             .ok_or(SafeTensorError::InvalidHeaderLength)?;
-        if stop > buffer_len {
+
+        let Some(header_bytes) = buffer.get(HEADER_SIZE_LEN..stop) else {
```
Narsil: The behavior of get on a slice out of bounds is not exactly clear imho, since purely slicing will return an incomplete slice instead. Can we at least keep a comment that this is indirectly checking that stop < buffer.len() ?

safetensors/src/tensor.rs
```diff
@@ -388,26 +400,26 @@ impl<'data> SafeTensors<'data> {
...
+
+        let Some(info) = &self.metadata.tensors.get(index) else {
+            return Err(SafeTensorError::TensorNotFound(tensor_name.to_string()));
+        };
+
+        Ok(TensorView {
+            dtype: info.dtype,
+            shape: info.shape.clone(),
+            data: &self.data[info.data_offsets.0..info.data_offsets.1],
+        })
```
Narsil: This is purely style, right ? I personally find the first version more readable, the control flow here is really confusing with the let else syntax. Preferences...

safetensors/src/tensor.rs
```diff
@@ -388,26 +400,26 @@ impl<'data> SafeTensors<'data> {
...
     }
 
     /// Return the names of the tensors within the SafeTensors.
     /// These are used as keys to access to the actual tensors, that can be
     /// retrieved using the tensor method.
-    pub fn names(&self) -> Vec<&'_ String> {
-        self.metadata.index_map.keys().collect()
+    pub fn names(&self) -> Vec<&'_ str> {
+        self.metadata.index_map.keys().map(String::as_str).collect()
```
Narsil, line 475: I'm not sure I see the benefit here, callers can also call it themselves if they really want `str` instead of  `&String` no ?

safetensors/src/tensor.rs
```diff
@@ -465,23 +477,21 @@ impl Serialize for Metadata {
...
+        let length = self.metadata.as_ref().map_or(0, HashMap::len);
+        let mut map = serializer.serialize_map(Some(self.tensors.len() + length))?;
+
         if let Some(metadata) = &self.metadata {
             map.serialize_entry("__metadata__", metadata)?;
         }
-        for (name, info) in tensors {
-            map.serialize_entry(&name, &info)?;
+
+        for (name, info) in names.iter().zip(&self.tensors) {
```
Narsil, line 553: This is much cleaner indeed.

safetensors/src/tensor.rs
```diff
@@ -507,7 +517,7 @@ impl Metadata {
             tensors,
             index_map,
         };
-        // metadata.validate()?;
+
```
Narsil: While working on a secondary crate, I realized there was no public API for Metadata, so I marked this as pub in another PR.

Since it's now public we really need to keep validate (so I uncommented it), to make sure the Metadata can only be valid. (Distinc names, ordered data_offsets vectors etc..)

safetensors/src/tensor.rs
```diff
@@ -849,7 +863,7 @@ mod tests {
     fn test_serialization() {
         let data: Vec<u8> = vec![0.0f32, 1.0, 2.0, 3.0, 4.0, 5.0]
             .into_iter()
-            .flat_map(|f| f.to_le_bytes())
+            .flat_map(f32::to_le_bytes)
```
Narsil: I prefer the earlier version, this one is quite important, as you can keep the loop ISO event with different dtypes.

safetensors/src/tensor.rs
```diff
@@ -989,9 +1003,10 @@ mod tests {
     }
 
     fn gpt2_like(n_heads: usize, model_id: &str) {
-        let mut tensors_desc = vec![];
-        tensors_desc.push(("wte".to_string(), vec![50257, 768]));
-        tensors_desc.push(("wpe".to_string(), vec![1024, 768]));
+        let mut tensors_desc = vec![
+            ("wte".to_string(), vec![50257, 768]),
+            ("wpe".to_string(), vec![1024, 768]),
+        ];
```
Narsil, line 1193: Thanks, this is cleaner indeed.

safetensors/src/tensor.rs
H2CO3: This doesn't remove the function itself, only the call. So are you saying the call itself is needed? (Otherwise, I would see no reason to keep around commented-out code.)

safetensors/src/tensor.rs
H2CO3: The point here is that `&String` is *strictly* worse than `&str` because:
1. it exposes the implementation detail and forces the implementation to use _exactly_ a `String` (and not, e.g. a `Cow<str>` or an `Arc<str>`), while;
2. not having any benefits to the caller, because `&String` can't do anything more than `&str`. As it's immutable, any functionality that would be useful for `String` manipulation – such as appending characters or increasing the capacity – is inaccessible on the returned reference, so it provides only the functionality of `&str`.
As such, it's preferable to return just an `&str`, because it doesn't take away any functionality from the caller, while allowing the implementation to swap out the backing type if needed.

safetensors/src/tensor.rs
H2CO3: > since purely slicing will return an incomplete slice instead.

No, that is incorrect. I think you are confusing this with Python's behavior. Rust does the correct and sensible thing and [returns `None` if either index is OOB](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=5ed6c35ae6f83e3dabe81f8cd13d4338).

Of course, feel free to add an explanatory comment anyway.

safetensors/src/tensor.rs
H2CO3: The array creation can't fail. The point is that this avoids the 8 individual bounds checks that the original code imposes, *and* it's much less repetitive, less noisy.

If you prefer not returning an error, the `.map_err()` can be changed to `.unwrap()` or `.expect("this can't happen")`.

safetensors/src/tensor.rs
H2CO3: I don't understand what you are referring to. `serde_json::to_vec()` [produces exactly the same output](https://docs.rs/crate/serde_json/latest/source/src/ser.rs#2237-2246) as `serde_json::to_string()`, it's just simpler than manually going through `to_string().into_bytes()` if you want a `Vec<u8>` anyway.

safetensors/src/tensor.rs
H2CO3: > This is not really HEADER_SIZE_LEN

Yes it is. I think you need to read and interpret the name more carefully. It _is_ the length of the header _size._

Note that I did **not** call it `HEADER_SIZE` or `HEADER_LEN`, because it's not the header size. It's the size of the number denoting the header length (kind of a double indirection going on). Yes, it's 8 bytes, but hard-coding that constant a dozen times into the code is really just asking for trouble.

I'd be more than happy to change the name to something that you find clearer; however I think `N_LEN` is non-informative and not helpful, as `N` is way too generic. Could it be e.g. `SIZEOF_HEADER_SIZE_NUMBER` or something like that?

safetensors/src/tensor.rs
H2CO3: Sure, if that was the intention then feel free to discard this change. 👍  However, it would then be beneficial to somehow indicate this more clearly, since the current comment specifically says that it's about the alignment (which coincides with the size).

safetensors/src/tensor.rs
H2CO3: I see, thanks. If you need to make this type-agnostic, the type can be left out and inference will fill it in, so `<_>::to_le_bytes` should work just as well and adapt to changes in the type, while still avoiding the unnecessary noise of a redundant closure wrapping a single call.

safetensors/src/tensor.rs
Narsil: Yes, it was commented out as it was only an internal function, and at some point it was easier to remove the call on specific pathways to allow reading "malformed"  data (Metadata can follow the spec, but still not respect Metadata layout, which would cause issues down the line). While it wasn't public, we made sure the data was valid in other ways, since it's public we must reject any invalid data directly.

safetensors/src/tensor.rs
Narsil: Indeed. I was thinking [..4] which does panic instead of returning None. for some reason I thought it was slicing inside.

safetensors/src/tensor.rs
Narsil: Won't the bound check get eliminated by the compiler given we are ensuring the length is bigger than 8 right above ?

But fair enough

safetensors/src/tensor.rs
H2CO3: Sure, then feel free to discard this change of course.

safetensors/src/tensor.rs
H2CO3: Yes, it can and probably will be eliminated indeed – this change is more about removing the repetitive code than raw performance.

safetensors/src/tensor.rs
H2CO3: Yes, this is purely style, or at least this change doesn't _intend_ to change semantics. The idea is to reduce the ever-growing nesting due to multiple conditions, and just return early if any criterion is violated. This is consistent with how all other error handling happens in Rust using the `?` operator.

In fact, this can be rewritten using the `?` operator, if you prefer to be more explicit about that:

```rust
        let &index = self.metadata.index_map.get(tensor_name).ok_or_else(
            || SafeTensorError::TensorNotFound(tensor_name.to_string())
        )?;

        let Some(info) = self.metadata.tensors.get(index).ok_or_else(
            || SafeTensorError::TensorNotFound(tensor_name.to_string())
        )?;

        Ok(TensorView {
            dtype: info.dtype,
            shape: info.shape.clone(),
            data: &self.data[info.data_offsets.0..info.data_offsets.1],
        })
```

If you find even that version less readable, feel free to undo this.

H2CO3:
@Narsil I believe I've addressed all of your comments; I basically undone the changes where you explained why the original was necessary in all places except for the early returning of errors in `SafeTensors::tensor()`.

safetensors/src/tensor.rs
Narsil: Let's go with N_LEN, it might be uninformative to you, but honestly keeping 8 everywhere hardcoded and called n when named is just fine to me. The whole point was to remove HEADER from the name.

> Yes, it's 8 bytes, but hard-coding that constant a dozen times into the code is really just asking for trouble.

Really ? I maintain quite a few librairies, and i don't think that it really matters. This constant will never change, and it's best descriptive name is "the first 8 bytes", so 8 isn't such a bad name.

> SIZEOF_HEADER_SIZE_NUMBER

I want the name to not contain HEADER, and let's keep it small, this is not java :)

safetensors/src/tensor.rs
Narsil: Because the fact that they emit no invalid utf-8 is an internal property, not an external one.

H2CO3: Alright, I don't think this particular change warrants arguing that much, so let's discard this change, then.

PR #629, opened 2025-07-03
Re-adding support for u16, u32, u64.

Narsil:
# What does this PR do?


McPatate (approved):

BenjaminBossan:
@Narsil What are the supported PyTorch versions? In v2.2.0, which is still supported by transformers, there is no `torch.uint64`, resulting in a failure to `import safetensors.torch`.

zhichao-aws:
> In v2.2.0, which is still supported by transformers, there is no torch.uint64, resulting in a failure to import safetensors.torch

Get the same exception when import PretrainedModel from transformers. I'm using transformers 4.44.1 and torch 2.0

PR #630, opened 2025-07-05
Support using paddle in safe_open

zeroRains:
# Support using paddle in `safe_open` API

This PR allows us to use `safe_open` API when set the `framework='pp'` or `framework='paddle'`. It just like the following code:
```python

tensors = {}
with safe_open(filename, framework="pp") as f:
    for k in f.keys():
        tensors[k] = f.get_tensor(k)
```

In this PR, I use the latest feature(MmapStorage) in paddle. It only support for paddle >= 3.1.0 or develop version.

We also can use the original way to load tensor if the paddle version is lower than 3.1.0.

> ps: MmapStorage still a PR https://github.com/PaddlePaddle/Paddle/pull/73820.  It will be merged in develop and 3.1.0 branch later.

### DIFF
--- a/bindings/python/src/lib.rs
+++ b/bindings/python/src/lib.rs
@@ -24,6 +24,7 @@ static NUMPY_MODULE: OnceLock<Py<PyModule>> = OnceLock::new();
 static TENSORFLOW_MODULE: OnceLock<Py<PyModule>> = OnceLock::new();
 static FLAX_MODULE: OnceLock<Py<PyModule>> = OnceLock::new();
 static MLX_MODULE: OnceLock<Py<PyModule>> = OnceLock::new();
+static PADDLE_MODULE: OnceLock<Py<PyModule>> = OnceLock::new();
 
 struct PyView<'a> {
     shape: Vec<usize>,
@@ -240,6 +241,7 @@ enum Framework {
     Tensorflow,
     Flax,
     Mlx,
+    Paddle,
 }
 
 impl fmt::Display for Framework {
@@ -250,6 +252,7 @@ impl fmt::Display for Framework {
             Framework::Tensorflow => "tensorflow",
             Framework::Flax => "flax",
             Framework::Mlx => "mlx",
+            Framework::Paddle => "paddle",
         })
     }
 }
@@ -271,6 +274,10 @@ impl<'source> FromPyObject<'source> for Framework {
             "jax" => Ok(Framework::Flax),
             "flax" => Ok(Framework::Flax),
             "mlx" => Ok(Framework::Mlx),
+
+            "paddle" => Ok(Framework::Paddle),
+            "paddlepaddle" => Ok(Framework::Paddle),
+            "pp" => Ok(Framework::Paddle),
             name => Err(SafetensorError::new_err(format!(
                 "framework {name} is invalid"
             ))),
@@ -282,6 +289,7 @@ impl<'source> FromPyObject<'source> for Framework {
 enum Device {
     Cpu,
     Cuda(usize),
+    Gpu(usize),
     Mps,
     Npu(usize),
     Xpu(usize),
@@ -299,6 +307,7 @@ impl fmt::Display for Device {
             Device::Cpu => write!(f, "cpu"),
             Device::Mps => write!(f, "mps"),
             Device::Cuda(index) => write!(f, "cuda:{index}"),
+            Device::Gpu(index) => write!(f, "gpu:{index}"),
             Device::Npu(index) => write!(f, "npu:{index}"),
             Device::Xpu(index) => write!(f, "xpu:{index}"),
             Device::Xla(index) => write!(f, "xla:{index}"),
@@ -327,6 +336,7 @@ impl<'source> FromPyObject<'source> for Device {
             match name.as_str() {
                 "cpu" => Ok(Device::Cpu),
                 "cuda" => Ok(Device::Cuda(0)),
+                "gpu" => Ok(Device::Gpu(0)),
                 "mps" => Ok(Device::Mps),
                 "npu" => Ok(Device::Npu(0)),
                 "xpu" => Ok(Device::Xpu(0)),
@@ -334,6 +344,7 @@ impl<'source> FromPyObject<'source> for Device {
                 "mlu" => Ok(Device::Mlu(0)),
                 "hpu" => Ok(Device::Hpu(0)),
                 name if name.starts_with("cuda:") => parse_device(name).map(Device::Cuda),
+                name if name.starts_with("gpu:") => parse_device(name).map(Device::Gpu),
                 name if name.starts_with("npu:") => parse_device(name).map(Device::Npu),
                 name if name.starts_with("xpu:") => parse_device(name).map(Device::Xpu),
                 name if name.starts_with("xla:") => parse_device(name).map(Device::Xla),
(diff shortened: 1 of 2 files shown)


bindings/python/src/lib.rs
```diff
@@ -271,6 +274,10 @@ impl<'source> FromPyObject<'source> for Framework {
             "jax" => Ok(Framework::Flax),
             "flax" => Ok(Framework::Flax),
             "mlx" => Ok(Framework::Mlx),
+
+            "paddle" => Ok(Framework::Paddle),
+            "paddlepaddle" => Ok(Framework::Paddle),
+            "pp" => Ok(Framework::Paddle),
```
Narsil: Can we remove paddlepaddle (I don't see any reference anywhere in the tutorials).

Same for `pp` unless I missed it as something being regularly used.

bindings/python/src/lib.rs
```diff
@@ -472,6 +490,43 @@ impl Open {
         })?;
 
         let storage = match &framework {
+            Framework::Paddle => Python::with_gil(|py| -> PyResult<Storage> {
+                let paddle = get_module(py, &PADDLE_MODULE)?;
+                let version: String = paddle.getattr(intern!(py, "__version__"))?.extract()?;
+                let version = Version::from_string(&version).map_err(SafetensorError::new_err)?;
+
+                // todo: version check, only paddle 3.1 or develop
+                if version >= Version::new(3, 1, 0) || version >= Version::new(0, 0, 0) {
```
Narsil: Can you remove version 0.0.0 ?

bindings/python/src/lib.rs
```diff
@@ -606,6 +661,88 @@ impl Open {
...
                 )
             }
+            Storage::PaddleStorage(storage) => {
+                Python::with_gil(|py| -> PyResult<PyObject> {
+                    let paddle = get_module(py, &PADDLE_MODULE)?;
+                    let mut cur_type = info.dtype;
+                    if cur_type == Dtype::U16 {
+                        // paddle set bf16 as u16
+                        cur_type = Dtype::BF16;
+                    }
```
Narsil: info is the source in the safetensors file, I think this is reversed logic.

Try to use non mutable variables. let cur_type = if info.dtype == xxx{ something } else{ something else};

Make it much easier to be sure no one if modifying this later on.

bindings/python/tests/test_paddle_comparison.py
```diff
@@ -1,7 +1,10 @@
+import threading
```
Narsil: Remove this import.

bindings/python/tests/test_paddle_comparison.py
```diff
@@ -45,3 +48,67 @@ def test_deserialization_safe(self):
...
+
+@unittest.skipIf(not HAS_PADDLE, "Paddle is not available")
+class WithOpenCase(unittest.TestCase):
+    def test_paddle_tensor(self):
+        A = paddle.randn((10, 5))
+        tensors = {
+            "a": A,
+        }
+        ident = threading.get_ident()
+        save_file(tensors, f"./tensor_{ident}.safetensors")
```
Narsil: Just use a unique name, no need to use ident.

bindings/python/src/lib.rs
```diff
@@ -282,6 +289,7 @@ impl<'source> FromPyObject<'source> for Framework {
 enum Device {
     Cpu,
     Cuda(usize),
+    Gpu(usize),
```
Narsil: No, it's called CUDA, let's not add something new. Let's reuse cuda(usize). If paddle uses a different name, let's just make sure the conversion happen correctly when they are called (so in paddle only code).

bindings/python/src/lib.rs
```diff
@@ -378,6 +390,7 @@ enum Storage {
     /// so Pytorch can handle the whole lifecycle.
     /// https://pytorch.org/docs/stable/storage.html#torch.TypedStorage.from_file.
     TorchStorage(OnceLock<PyObject>),
+    PaddleStorage(OnceLock<PyObject>),
```
Narsil, line 390: Documentation for where this is documented and how it behaves. Importantly, is there a security notice alongside it ?

Narsil (commented):
Looking good left a few comments.

bindings/python/src/lib.rs
zeroRains: We can remove paddlepaddle.  But we should save pp. pp in paddle just like pt in pytorch and np in numpy.

I also use pp in test case file. Do you mean that the abbreviation is not required?

bindings/python/src/lib.rs
zeroRains: ok, I will remove it.

bindings/python/src/lib.rs
zeroRains: Version 0.0.0 means  user uses the develop version. The latest develop version have supported this function.  Maybe I should use == but not >=.

bindings/python/src/lib.rs
zeroRains: Ok, I will modify it.

bindings/python/src/lib.rs
zeroRains: OK, I will add a document link.  https://www.paddlepaddle.org.cn/documentation/docs/en/develop/api/paddle/MmapStorage_en.html

bindings/python/tests/test_paddle_comparison.py
zeroRains: Ok

bindings/python/tests/test_paddle_comparison.py
zeroRains: Ok

zeroRains:
I have modified the code follow the comments. Could you review it again? @Narsil

bindings/python/src/lib.rs
Narsil: I meant, can you point me to places where the abbreviation is used "naturally". Something like tutorials, examples docs from the official paddlepaddle project (or some big user of it). 

For numpy (np): https://numpy.org/doc/stable/user/absolute_beginners.html
For pytorch: https://huggingface.co/docs/transformers/main_classes/tokenizer#transformers.PreTrainedTokenizerFast.__call__.return_tensors It's a convention used in a few key places in HF's ecosystem hence why I went with it (despite my preference for having only 1 clear name for things).

bindings/python/src/lib.rs
Narsil: Then yes == is preferred, otherwise afaik users for versions > 0.0.0 < 3.1.0 will encounter issues, right ?

bindings/python/src/lib.rs
Narsil: Re-reading this I have question about GPU semantics with paddle, how does it deal with non cuda GPUs ?

bindings/python/src/lib.rs
Narsil: Great.

bindings/python/src/lib.rs
zeroRains: Yes, this feature only support for paddle>=3.1.1 and develop version(0.0.0).  It should use the Mmap of safetensors for versions > 0.0.0 < 3.1.1.

bindings/python/src/lib.rs
zeroRains: Right,  it seems no need to save pp, I will remove it later.

bindings/python/src/lib.rs
zeroRains: Sorry, I misunderstood your question. The `gpu` in paddle refers to the cuda GPU, and other non cuda GPUs are represented by xxx_gpu(like intel_gpu/metax_gpu). This repo will show more ways to access hardware. https://github.com/PaddlePaddle/PaddleCustomDevice 

@Narsil

zeroRains:
> Since https://github.com/huggingface/safetensors/pull/646/commits is open, maybe we should close this ? It seems the overlap is significant here, right ?

No, this PR https://github.com/huggingface/safetensors/pull/646 still has some content that needs to be modified, but the `safe_open` function is relatively complete. We hope to use the `safe_open` function as soon as possible, so we hope to merge the current PR first, and then synchronize  PR https://github.com/huggingface/safetensors/pull/646 from the main branch.

Narsil:
Well my comments from here still apply. This PR is breaking paddle <3.1.1

zeroRains:
> Well my comments from here still apply. This PR is breaking paddle <3.1.1

OK, I am install the paddlepaddle == 3.0.0. I run the test file `test_paddle_comparison.py`.  I got this error.

image

I should handle the device when I use `cuda:x`, now it has solved.

Is there any other testing case which will trigger the breaking when paddle < 3.1.1 ?

PR #637, opened 2025-08-05
Add a multithreaded stress test

ngoldbaum:
# What does this PR do?

Adds a test based on the script I used to find https://github.com/pytorch/pytorch/issues/158071.

The test serializes a dict of tensors in many threads simultaneously. Currently this is enough to segfault pytorch on the free-threaded build - no mutation necessary.

IMO having a multithreaded stress test like this is worth doing, especially looking torward adding support for the free-threaded build.

### DIFF
--- a/bindings/python/tests/test_multithreaded.py
+++ b/bindings/python/tests/test_multithreaded.py
@@ -2,8 +2,6 @@ import sys
 import tempfile
 import threading
 
-import jax
-import jax.numpy as jnp
 import numpy as np
 import pytest
 import torch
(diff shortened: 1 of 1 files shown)


ngoldbaum:
Much simpler now, thanks for the code review.

Narsil:
Great, it's working and I can reproduce the segfault. Let's probably wait for it to be fixed in torch before merging this, no ?

Narsil:
I don't like skipping a test because it fails. The point of tests is to let them fail when something is wrong. If we simply skip them when something is working, then nothing will ever check that finally they are working again and we should switch them.

That's why I suggest we simply postpone this. The fact that the PR is open, means it will be revisited at some point.

McPatate:
> @danieldk, maybe look at https://github.com/huggingface/safetensors/pull/699 first though.

As said in #699, we're sticking with 3.13 for the moment. This PR looks like it can be merged independently! Lmk if there was a particular reason to merge #699 before.

PR #640, opened 2025-08-06
Add a direct Python test dependency on fsspect in the testing and testingfree extras

musicinmybrain:
# What does this PR do?

Adds `fsspec` to the Python testing dependencies in the `testing` and `testingfree` extras.

Fixes `ReadmeTestCase.test_fsspec` failing due to missing `fsspec` when testing without PyTorch available. Even when `torch` does bring in `fsspec` as an indirect dependency, it’s better to have a direct dependency on things we import directly.


McPatate:
closing in favour of #733

PR #646, opened 2025-08-11
Support paddle save/load/save_file/load_file without coverting to numpy

changeyoung98:
# What does this PR do?

Based on #630 .
This PR refactors the logic of the `save`, `load`, `save_file`, and `load_file` operations in `paddle.py` to eliminate the need for NumPy conversion. As a result, Paddle can now utilize safetensors to save and load additional data types—such as bfloat16 (bf16)—that were previously unsupported.

### DIFF
--- a/bindings/python/py_src/safetensors/paddle.py
+++ b/bindings/python/py_src/safetensors/paddle.py
@@ -7,7 +7,9 @@ import paddle
 from safetensors import numpy
 
 
-def save(tensors: Dict[str, paddle.Tensor], metadata: Optional[Dict[str, str]] = None) -> bytes:
+def save(
+    tensors: Dict[str, paddle.Tensor], metadata: Optional[Dict[str, str]] = None
+) -> bytes:
     """
     Saves a dictionary of tensors into raw bytes in safetensors format.
 
@@ -98,7 +100,9 @@ def load(data: bytes, device: str = "cpu") -> Dict[str, paddle.Tensor]:
     return _np2paddle(flat, device)
 
 
-def load_file(filename: Union[str, os.PathLike], device="cpu") -> Dict[str, paddle.Tensor]:
+def load_file(
+    filename: Union[str, os.PathLike], device="cpu"
+) -> Dict[str, paddle.Tensor]:
     """
     Loads a safetensors file into paddle format.
 
@@ -126,7 +130,9 @@ def load_file(filename: Union[str, os.PathLike], device="cpu") -> Dict[str, padd
     return output
 
 
-def _np2paddle(numpy_dict: Dict[str, np.ndarray], device: str = "cpu") -> Dict[str, paddle.Tensor]:
+def _np2paddle(
+    numpy_dict: Dict[str, np.ndarray], device: str = "cpu"
+) -> Dict[str, paddle.Tensor]:
     for k, v in numpy_dict.items():
         numpy_dict[k] = paddle.to_tensor(v, place=device)
     return numpy_dict
--- a/bindings/python/py_src/safetensors/paddle.py
+++ b/bindings/python/py_src/safetensors/paddle.py
@@ -1,10 +1,11 @@
 import os
-from typing import Dict, Optional, Union
+import sys
+from typing import Any, Dict, Optional, Union
 
 import numpy as np
-
 import paddle
-from safetensors import numpy
+
+from safetensors import deserialize, safe_open, serialize, serialize_file
 
 
 def save(
@@ -34,8 +35,9 @@ def save(
     byte_data = save(tensors)
     ```
     """
-    np_tensors = _paddle2np(tensors)
-    return numpy.save(np_tensors, metadata=metadata)
+    serialized = serialize(_flatten(tensors), metadata=metadata)
+    result = bytes(serialized)
+    return result
 
 
 def save_file(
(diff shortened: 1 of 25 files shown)


changeyoung98:
Could you please review this PR? @Narsil Thank you.

bindings/python/py_src/safetensors/paddle.py
```diff
@@ -1,10 +1,11 @@
 import os
-from typing import Dict, Optional, Union
+import sys
```
Narsil, line 2: This modification is breaking paddle < 3.1.1 which isn't great.

bindings/python/tests/test_paddle_comparison.py
```diff
@@ -45,3 +46,192 @@ def test_deserialization_safe(self):
...
+            assert tensor.place.is_gpu_place()
+
+            tensor = slice_[2:, -1]
+            self.assertEqual(list(tensor.shape), [8])
+            assert paddle.allclose(tensor, A[2:, -1]).item()
+            assert tensor.place.is_gpu_place()
+
+            tensor = slice_[list()]
+            self.assertEqual(list(tensor.shape), [0, 5])
+            assert paddle.allclose(tensor, A[list()]).item()
```
Narsil: This test never passes because allclose on empty set returns False.

Narsil (commented):
Thanks for this.

The biggest blocker is that this code is breaking paddle < 3.1. 

For bflioat16 we could skip the test if < 3.1 doesn't support them.

As a side node, it seems the benchmarks don't show that big of an improvement which seems suprising.

```
======================================================================================================================= warnings summary =======================================================================================================================
../../.venv/lib/python3.12/site-packages/paddle/utils/cpp_extension/extension_utils.py:717
  /home/nicolas/src/safetensors/.venv/lib/python3.12/site-packages/paddle/utils/cpp_extension/extension_utils.py:717: UserWarning: No ccache found. Please be aware that recompiling all source files may be required. You can download and install ccache from: https://github.com/ccache/ccache/blob/master/doc/INSTALL.md
    warnings.warn(warning_message)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html

-------------------------------------------------------------------------------------- benchmark: 2 tests -------------------------------------------------------------------------------------
Name (time in ms)                Min                 Max                Mean            StdDev              Median                IQR            Outliers     OPS            Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_paddle_paddle_load     371.5229 (1.0)      393.8909 (1.0)      380.8332 (1.0)      8.7418 (1.19)     381.2460 (1.0)      12.2720 (1.07)          2;0  2.6258 (1.0)           5           1
test_paddle_sf_load         474.4823 (1.28)     491.2912 (1.25)     480.5302 (1.26)     7.3160 (1.0)      476.6819 (1.25)     11.4378 (1.0)           1;0  2.0810 (0.79)          5           1
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

Legend:
  Outliers: 1 Standard Deviation from Mean; 1.5 IQR (InterQuartile Range) from 1st Quartile and 3rd Quartile.
  OPS: Operations Per Second, computed as 1 / Mean
================================================================================================================ 2 passed, 1 warning in 19.98s
```

bindings/python/tests/test_paddle_comparison.py
zeroRains: image

No, allclose on empty set returns True in paddle.

Narsil: I assure you there's one build that failed on this test returning false. Not sure which version it was since I scratched that environment. But if it was a bug in paddle, fine by me.

bindings/python/py_src/safetensors/paddle.py
changeyoung98: I’ve made several improvements to the code：
- Added version-dispatch logic for different Paddle versions.
- Refactored `load` using the new `frombuffer` function in Paddle, which is based on MmapStorage for improved performance. The new function is now available in develop, and will be released in 3.2.0.

Could you please review the updated implementation again? Thanks a lot! @Narsil

bindings/python/py_src/safetensors/paddle.py
```diff
@@ -125,9 +130,15 @@ def load_file(
     loaded = load_file(file_path)
     ```
     """
-    flat = numpy.load_file(filename)
-    output = _np2paddle(flat, device)
-    return output
+    result = {}
+    if paddle.__version__ == "0.0.0" or paddle.__version__ >= "3.1.1":
```
Narsil: Why are you using 2 different versions between deserialize and here ?

changeyoung98: "0.0.0" is the develop_version for Paddle, we just hope that developers using the develop Paddle can also use this function.

changeyoung98: > Why are you using 2 different versions between deserialize and here ?

Hi @Narsil , we have refined the version of Paddle in the develop branch, now it is no more "0.0.0" and can easily use one single statement to do the version control. As for the distinction of '3.1.1' and '3.2.0', that's because we recently added the `frombuffer` function after '3.1.1', so we should use a newer version '3.2.0' . Could you please take another look and consider merging this PR into the main branch? We're eager to release an updated version of safetensors that includes these enhancements.

Narsil: Can we have 3.2.0 everywhere? Adding version checks is a maintenance burden, having as few as possible is easier to maintain (and remember).

changeyoung98: Alright~ Now all versions are 3.2.0 now.

Narsil (approved):
LGTM thanks !

PR #651, opened 2025-09-05
Fix no-std feature configuration to auto-fallback from std to alloc

### LINKED ISSUE #650: Cannot automatically select between `std` and `alloc` based on default features
### Problem Description

The `safetensors` crate requires exactly one of two features to be enabled: either `std` OR `alloc`, but not both. This creates a challenge for downstream crates that want to support both `std` and `no_std` environments.

### Current Behavior

The crate enforces this requirement with compile-time checks:
```rust
#[cfg(all(feature = "std", feature = "alloc"))]
compile_error!("must choose either the `std` or `alloc` feature, but not both.");

#[cfg(all(not(feature = "std"), not(feature = "alloc")))]
compile_error!("must choose either the `std` or `alloc` feature");
```

### The Challenge

When building a crate that depends on `safetensors`:

1. **For `std` builds**: We need to enable `safetensors/std`
2. **For `no_std` builds**: We need to enable `safetensors/alloc`

However, Cargo's feature system is additive, making it impossible to conditionally enable different features based on whether default features are disabled. We cannot use:
- Target-specific dependencies (`[target.'cfg(not(feature = "std"))'.dependencies]`) because they work with platform features, not crate features
- A single dependency declaration that automatically selects the right feature

### Current Workaround

pass `None` for metadata for no_std scenario 

### Expected Behavior

Ideally, when a crate is built with `--no-default-features`, it should automatically use `no_std` mode. The common pattern in the Rust ecosystem is:

```toml
[features]
default = ["std"]
std = ["dep:some_dependency", ...]
```

When built with `--no-default-features`, the crate automatically becomes `no_std` compatible.

### Suggested Solution

1. Use `default = ["std"]` 
2. When `std` is not enabled, automatically fall back to `alloc`
3. Remove the explicit `alloc` feature flag

This would align with how most `no_std` compatible crates in the ecosystem work, where:
- Building with default features gives you `std`
- Building with `--no-default-features` gives you `no_std` (with `alloc`)

### Impact

This issue affects any downstream crate that wants to support both `std` and `no_std` environments while depending on `safetensors`. It makes it particularly challenging for libraries that want to be flexible about their runtime environment.

### Environment

- safetensors version: 0.6.2
- Rust version: 1.88+
- Cargo version: Latest

### Related Context

We're working on the Burn deep learning framework which supports both `std` and `no_std` environments for embedded and WASM targets. The current feature configuration of `safetensors` makes it difficult to cleanly support both modes without complex workarounds.

antimora:
This PR addresses issue #650 by following the convention of the feature configuration for no_std support.

### Changes

- Removed mutual exclusivity between std and alloc features: Previously, the crate required exactly one of std or alloc
to be enabled, causing compilation errors when both or neither were specified
- Automatic alloc fallback: When std is not enabled, the crate now automatically uses alloc + hashbrown::HashMap
- Backward compatibility: The alloc feature is retained as a no-op for compatibility with existing downstream crates

### How it works

- Default build (cargo build): Uses std with std::collections::HashMap
- No-std build (cargo build --no-default-features): Automatically uses alloc with hashbrown::HashMap
- Explicit alloc (cargo build --no-default-features --features alloc): Still works for backward compatibility

### Testing

All tests pass in both configurations:
```
cargo test                          # std mode ✓
cargo test --no-default-features    # no-std mode ✓
cargo test --no-default-features --features alloc  # backward compat ✓
```
This aligns with common Rust ecosystem patterns where `default = ["std"]` provides standard library support, and
`--no-default-features` automatically provides `no_std` support with `alloc`.

Fixes #650

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -46,7 +46,7 @@ jobs:
         run: cargo test --verbose
 
       - name: Run No-STD Tests
-        run: cargo test --no-default-features --features alloc --verbose
+        run: cargo test --no-default-features --verbose
 
       - name: Run Audit
         # RUSTSEC-2021-0145 is criterion so only within benchmarks
--- a/safetensors/Cargo.toml
+++ b/safetensors/Cargo.toml
@@ -16,15 +16,18 @@ The format is 8 bytes which is an unsized int, being the size of a JSON header,
 the JSON header refers the `dtype` the `shape` and `data_offsets` which are the offsets
 for the values in the rest of the file.
 """
-exclude = [ "rust-toolchain", "target/*", "Cargo.lock"]
+exclude = ["rust-toolchain", "target/*", "Cargo.lock"]
 
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-hashbrown = { version = "0.15.4", features = ["serde"], optional = true }
-serde = { version = "1.0", default-features = false, features = ["derive"] }
-serde_json = { version = "1.0", default-features = false }
+serde = { version = "1.0", default-features = false, features = [
+    "derive",
+    "alloc",
+] }
+serde_json = { version = "1.0", default-features = false, features = ["alloc"] }
+hashbrown = { version = "0.15.4", features = ["serde"] }
 
 [dev-dependencies]
 criterion = "0.6"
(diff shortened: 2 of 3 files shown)


antimora:
Hi @Narsil, this is to support both std/no_std. The feature flag patten follows "std" flag convention. This is needed for Burn project to support no_std when "std" is disabled by no-default-features. I tested locally and it works.

Narsil (approved):
LGTM. I'm not sure why we wrote it being exclusive features in the first place.

Thanks.

PR #652, opened 2025-09-23
Make converter more friendly to Transformers

Cyrilvallez:
# What does this PR do?

As per the title. This upstream the change made to the space converter https://huggingface.co/spaces/safetensors/convert/discussions/41.
Related to https://github.com/huggingface/transformers/issues/41038 and https://github.com/huggingface/transformers/issues/38870 for the full picture

### DIFF
--- a/bindings/python/convert.py
+++ b/bindings/python/convert.py
@@ -218,7 +218,7 @@ def convert_file(
     sf_filename: str,
     discard_names: List[str],
 ):
-    loaded = torch.load(pt_filename, map_location="cpu", weights_only=True)
+    loaded = torch.load(pt_filename, map_location="cpu")
     if "state_dict" in loaded:
         loaded = loaded["state_dict"]
     to_removes = _remove_duplicate_names(loaded, discard_names=discard_names)


bindings/python/convert.py
```diff
@@ -218,7 +218,7 @@ def convert_file(
     sf_filename: str,
     discard_names: List[str],
 ):
-    loaded = torch.load(pt_filename, map_location="cpu")
+    loaded = torch.load(pt_filename, map_location="cpu", weights_only=True)
```
Narsil: This will break pytorch lightning saved weights (which use a full dictionnary with "state_dict" for the actual state dict, and iirc (and unless they changed) they also save some custom lightning objects which are not weights only.

Narsil:
> Could you move convert.py into the safetensors package? It will make convertation to safetensors with safetensors more user-friendly.

Please no. This script is way too complex for most cases as it tries to solve transformers specific cases. And simple `torch.load` `safetensors.torch.save_file` should be enough in most cases (with save_model enabling almost every other use case).

`convert.py` is in the source code so it's readable and can provide help for power users, but shouldn't be distributed in the distributed code imho because it has *knowingly* some gaps and specific choices which don't work in general.

McPatate:
@Cyrilvallez script seems to have been updated again, could you upstream the new changes as well please 🙏🏻 ?

McPatate:
closing in favour of #746

PR #656, opened 2025-10-02
Use BTreeMap for metadata to eliminate hashbrown version conflicts

antimora:
### Problem

The serialize() function accepted Option<HashMap<String, String>> in its public API, which exposed hashbrown as a dependency and locked users to a
 specific version (0.15.5). This caused type mismatch errors in downstream crates like burn-store that use different hashbrown versions:

```
error[E0308]: mismatched types
  --> crates/burn-store/src/safetensors/store.rs:539:65
   |
   | let data = safetensors::serialize(tensors, Some(metadata))?;
   |                                            ^^^^ expected HashMap<String, String>, 
        found a different HashMap<String, String>
   |
note: two different versions of crate hashbrown are being used
```

### Solution

Changed the public API to use BTreeMap<String, String> instead of HashMap:

```rust
pub fn serialize<S, V, I>(
    data: I,
    data_info: Option<BTreeMap<String, String>>,  //  Now BTreeMap
) -> Result<Vec<u8>, SafeTensorError>
```

### Why BTreeMap?

1. No third-party dependency - BTreeMap is part of Rust's standard library (alloc::collections), available in both std and no_std
2. No version conflicts - Not tied to any external crate version
3. Still supports no_std - Available in alloc::collections::BTreeMap
4. Maintains performance - Internal index_map still uses HashMap for O(1) lookups
5. Added bonus the order is preserved

### Changes

- Public API: All metadata parameters now use BTreeMap
- Internal implementation: HashMap still used for index_map (performance)
- Imports: Unified approach using alloc for both std and no_std
- Removed facade: Simplified from mod lib to direct imports

### Compatibility

-  Backward compatible: serialize(tensors, None) still works without type annotations
-  All existing tests pass
-  No_std builds work (cargo build --lib --no-default-features)
-  No clippy warnings

---
This change decouples safetensors from hashbrown versions while maintaining full functionality and no_std support.

### DIFF
--- a/safetensors/src/lib.rs
+++ b/safetensors/src/lib.rs
@@ -1,38 +1,17 @@
+#![cfg_attr(not(feature = "std"), no_std)]
 #![deny(missing_docs)]
 #![doc = include_str!("../README.md")]
-#![cfg_attr(not(feature = "std"), no_std)]
-pub mod slice;
-pub mod tensor;
-/// serialize_to_file only valid in std
-#[cfg(feature = "std")]
-pub use tensor::serialize_to_file;
-pub use tensor::{serialize, Dtype, SafeTensorError, SafeTensors, View};
 
-#[cfg(not(feature = "std"))]
+// Always use alloc for consistency between std and no_std
+
 #[macro_use]
 extern crate alloc;
 
-/// A facade around all the types we need from the `std`, `core`, and `alloc`
-/// crates. This avoids elaborate import wrangling having to happen in every
-/// module.
-mod lib {
-    #[cfg(not(feature = "std"))]
-    mod no_stds {
-        pub use alloc::borrow::Cow;
-        pub use alloc::string::{String, ToString};
-        pub use alloc::vec::Vec;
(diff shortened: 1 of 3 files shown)


Narsil:
Wouldn't vendoring safetensors' hashbrown solve this issue ?

```rust
#[no_std]
pub use hashbrown;
```

Normally, rust doesn't break if multiple versions of a given crate exist in tree...

McPatate:
> I will check if your suggested solution works and will close this PR if does. However, it would have a been a good opportunity to decouple from 3rd party dependency for the API, but it's your call.

Any updates on this front? Will close the PR if not!

antimora:
> > I will check if your suggested solution works and will close this PR if does. However, it would have a been a good opportunity to decouple from 3rd party dependency for the API, but it's your call.
> 
> Any updates on this front? Will close the PR if not!

yes, it worked; we can close it (although, I still recommend clean up the API and not depend on 3rd party lib interfaces).

PR #660, opened 2025-10-17
Support the complex64 data type

danieldk:
**Warning:** not ready for review yet, I just want a CI run :).

# What does this PR do?

### DIFF
--- a/bindings/python/py_src/safetensors/numpy.py
+++ b/bindings/python/py_src/safetensors/numpy.py
@@ -154,7 +154,7 @@ _TYPES = {
     "I8": np.int8,
     "U8": np.uint8,
     "BOOL": bool,
-    "Complex64": np.complex64,
+    "C64": np.complex64,
 }
 
 
--- a/bindings/python/py_src/safetensors/torch.py
+++ b/bindings/python/py_src/safetensors/torch.py
@@ -411,7 +411,7 @@ _TYPES = {
     "BOOL": torch.bool,
     "F8_E4M3": _float8_e4m3fn,
     "F8_E5M2": _float8_e5m2,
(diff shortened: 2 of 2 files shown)


safetensors/src/tensor.rs
```diff
@@ -806,6 +806,8 @@ pub enum Dtype {
     U32,
     /// Floating point (32-bit)
     F32,
+    /// Complex (32-bit parts)
+    Complex64,
```
McPatate: 32-bit parts for complex64?

danieldk: Yeah:

```
>>> x = torch.zeros(1, dtype=torch.complex64)
>>> x.real.dtype
torch.float32
>>> x.imag.dtype
torch.float32
```

McPatate: somehow didn't parse `parts` (with an `s`), which was pretty clear, mb 🙃

safetensors/src/tensor.rs
```diff
@@ -871,6 +874,7 @@ impl Display for Dtype {
             Dtype::BF16 => "BF16",
             Dtype::F32 => "F32",
             Dtype::F64 => "F64",
+            Dtype::Complex64 => "Complex64",
```
Narsil: Any way we could keep things in line with what's before `"C64"` just for consistency's sake ?

Hi guys !

danieldk: Hey hey! Hope you are doing great!

Changed to `C64`, now one can only hope for a type `A500` in the future 😁.

McPatate: Feels a bit confusing to have it be `C64` in one place but `Complex64` everywhere else imo

Issue #668, opened 2025-10-28
SafetensorError: Error while deserializing header: header too large


KaneGreen:
### System Info

```
safetensors=0.6.2
python=3.13.9
```
All packages are installed from `conda-forge`

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### Reproduction

I'm using safetensors to save my own dataset, but when I try to read the data I'm getting this error.
```
  File "D:\xxxxxx\xxxxxxx.py", line 80, in __init__
    with safe_open(data_file, framework="pt") as f:
         ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
safetensors_rust.SafetensorError: Error while deserializing header: header too large
```

I knew my dataset was quite large: this safetensors file was about 7 GB and contained approximately 1,290,000 tensors, no metadata.

According to the [documentation](https://huggingface.co/docs/safetensors/index#format), I checked the first 8 bytes of the safetensors file and found them to be `40 9D DF 07 00 00 00 00`, so the header should be 132,095,296 bytes, or about 126MB.

This is certainly a bit large, but it shouldn't be unparseable.

This issue seems to be related to this hardcoded limitation: https://github.com/huggingface/safetensors/blob/v0.6.2/safetensors/src/tensor.rs#L10

### Expected behavior

This dataset should be able to load.

PR #669, opened 2025-10-29
feat: add header size check at serialization

McPatate:
#668 prompted the need to check the header size at serialization, to avoid producing files that will error when deser.

### DIFF
--- a/safetensors/src/tensor.rs
+++ b/safetensors/src/tensor.rs
@@ -1510,7 +1510,7 @@ mod tests {
 
         // a char is 1 byte in utf-8, so we can just repeat 'a' to get large metadata
         let very_large_metadata = "a".repeat(MAX_HEADER_SIZE);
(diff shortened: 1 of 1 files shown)


safetensors/src/tensor.rs
```diff
@@ -1494,4 +1502,21 @@ mod tests {
             _ => panic!("This should not be able to be deserialized"),
         }
     }
+
+    #[test]
+    fn test_invalid_header_size_serialization() {
```
danieldk, line 1507:
```suggestion
    fn test_invalid_header_size_deserialization() {
```

Maybe it's worth adding the counterpart for serialization, serializing a bunch of zero-sized tensors where the metadata is going to be > 100_000?

McPatate: Sorry my test was garbage I went a bit too quick. The new version should be better!

danieldk (approved):

PR #676, opened 2025-11-22
nvidia gds support

### LINKED ISSUE #299: Any plan to support Nvidia GPUDirect Storage?
### Feature request

Nvidia GPUDirect Storage has better performance to load model from NVMe disk or supported distributed storage. It will do the real `zero copy`.

### Motivation

It will get better performance with Nvidia GDS.

### Your contribution

Not sure.

maifeeulasad:
# What does this PR do?

Fixes #299

This PR adds support for GDS (GPU direct access) for pytorch for now. If I get positive feedback, definetly I am willing to integrate in other modules as well.

Open for making changes based on review.

### DIFF
--- a/bindings/python/build.rs
+++ b/bindings/python/build.rs
@@ -1,6 +1,6 @@
 fn main() {
     // Try linking necessary CUDA libraries on Linux for GDS
-    #[cfg(target_os = "linux")]
+    #[cfg(all(target_os = "linux", feature = "cuda-gds"))]
     {
         // Add CUDA library paths for linking
         println!("cargo:rustc-link-search=/usr/local/cuda/lib64");
--- a/bindings/python/src/gds/mod.rs
+++ b/bindings/python/src/gds/mod.rs
@@ -3,24 +3,24 @@
 //! This module provides zero-copy direct storage access between NVMe SSDs
 //! and GPU memory, bypassing the CPU for utilizing high-throughput storage.
 
-#[cfg(target_os = "linux")]
+#[cfg(all(target_os = "linux", feature = "cuda-gds"))]
 pub mod bindings;
-#[cfg(target_os = "linux")]
+#[cfg(all(target_os = "linux", feature = "cuda-gds"))]
 pub mod driver;
-#[cfg(target_os = "linux")]
+#[cfg(all(target_os = "linux", feature = "cuda-gds"))]
 pub mod error;
-#[cfg(target_os = "linux")]
+#[cfg(all(target_os = "linux", feature = "cuda-gds"))]
 pub mod handle;
-#[cfg(target_os = "linux")]
+#[cfg(all(target_os = "linux", feature = "cuda-gds"))]
 pub mod storage;
 
-#[cfg(target_os = "linux")]
+#[cfg(all(target_os = "linux", feature = "cuda-gds"))]
 pub use driver::GdsDriver;
-#[cfg(target_os = "linux")]
+#[cfg(all(target_os = "linux", feature = "cuda-gds"))]
 pub use error::GdsError;
-#[cfg(target_os = "linux")]
+#[cfg(all(target_os = "linux", feature = "cuda-gds"))]
 pub use handle::GdsHandle;
-#[cfg(target_os = "linux")]
+#[cfg(all(target_os = "linux", feature = "cuda-gds"))]
 pub use storage::GdsStorage;
 
 // Suppress unused warnings for conditional compilation
--- a/bindings/python/src/lib.rs
+++ b/bindings/python/src/lib.rs
@@ -19,7 +19,7 @@ use std::path::PathBuf;
 use std::sync::Arc;
 use std::sync::OnceLock;
 
-#[cfg(target_os = "linux")]
+#[cfg(all(target_os = "linux", feature = "cuda-gds"))]
 mod gds;
 
 static TORCH_MODULE: OnceLock<Py<PyModule>> = OnceLock::new();
@@ -394,7 +394,7 @@ enum Storage {
     Paddle(OnceLock<PyObject>),
     /// GPU Direct Storage for CUDA devices
     /// Provides zero-copy loading directly from NVMe to GPU memory
-    #[cfg(target_os = "linux")]
+    #[cfg(all(target_os = "linux", feature = "cuda-gds"))]
     CudaGds(gds::GdsStorage),
 }
 
@@ -460,7 +460,7 @@ impl Open {
         
         // Validate GDS usage
         if use_gds {
-            #[cfg(target_os = "linux")]
+            #[cfg(all(target_os = "linux", feature = "cuda-gds"))]
             {
                 if !matches!(device, Device::Cuda(_)) {
                     return Err(SafetensorError::new_err(
@@ -473,10 +473,11 @@ impl Open {
                     ));
                 }
             }
-            #[cfg(not(target_os = "linux"))]
+            #[cfg(not(all(target_os = "linux", feature = "cuda-gds")))]
             {
                 return Err(SafetensorError::new_err(
-                    "GPU Direct Storage is only available on Linux"
+                    "GPU Direct Storage is not available in this build. \
+                     To enable GDS support, rebuild with: maturin build --features cuda-gds"
                 ));
             }
         }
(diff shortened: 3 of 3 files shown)


bindings/python/py_src/safetensors/__init__.py
```diff
@@ -8,3 +8,6 @@
     serialize,
     serialize_file,
 )
+
+# Export GDS support status for programmatic checking
+from .torch import _is_gds_available  # noqa: F401
```
Narsil: You cannot import torch here, this file cannot depend on torch.

bindings/python/py_src/safetensors/torch.py
```diff
@@ -8,6 +9,14 @@
 
 from safetensors import deserialize, safe_open, serialize, serialize_file
 
+# GDS (GPU Direct Storage) support
+try:
+    import cupy as cp
```
Narsil: That's yet another library, we can't have another dependency here. Even if it was a core dependency of torch we wouldn't access it directly.

maifeeulasad:
The benchmarking and testing notebook can be found here, just run all: https://gist.github.com/maifeeulasad/741df3cf99a9970ce7a4897d5583b223

Kaggle uses loopback FS or tmpfs or overlay FS or container-mounted ephemeral disk.  But gds requires the following:
 - XFS and EXT4 filesystem in ordered mode on NVMe/NVMeOF/ScaleFlux CSD devices.
 - NFS over RDMA with MOFED 5.1 and above
 - RDMA capable distributed filesystems like DDN Exascaler, WekaFS, and VAST.
 - ScaleFlux Computational storage
 
So I wasn't able to test in Kaggle. Haven't tested on Colab yet, but I think the situation would be pretty much same. And most of the data centers still uses HDDs.

Based on what I have implemented so far, and tested so far, 
 - GDS performs really well when the tensor size is big
 - It also performs really well when block the RAM with another script. 
 - It also performs really well if we put lots of weight of CPU

So I see usage of GDS in real life scenario. Where loads are quite noisy, specially when we are dealing with lots of models and application.

I will try to share more information soon. **Open for discussion.**

ref:
 - https://github.com/NVIDIA/gds-nvidia-fs?tab=readme-ov-file#overview

bindings/python/py_src/safetensors/__init__.py
maifeeulasad: Understood, that's why moved GDS related code into rust

bindings/python/py_src/safetensors/torch.py
maifeeulasad: Understood the architecture, moved GDS related code in rust code with FFI calling with `libc`

frauttauteffasu:
This adds cufile as a hard dependency, failing to build on any platform without it: 
```
         Compiling safetensors v0.7.0-dev.0 (/tmp/safetensors/safetensors)
      warning: unused import: `driver::GdsDriver`
        --> src/gds/mod.rs:18:9
         |
      18 | pub use driver::GdsDriver;
         |         ^^^^^^^^^^^^^^^^^
         |
         = note: `#[warn(unused_imports)]` on by default

      warning: unused import: `error::GdsError`
        --> src/gds/mod.rs:20:9
         |
      20 | pub use error::GdsError;
         |         ^^^^^^^^^^^^^^^

      warning: unused import: `handle::GdsHandle`
        --> src/gds/mod.rs:22:9
         |
      22 | pub use handle::GdsHandle;
         |         ^^^^^^^^^^^^^^^^^

      warning: unused variable: `result`
        --> src/gds/handle.rs:78:21
         |
      78 |                 let result = cuFileHandleDeregister(self.handle);
         |                     ^^^^^^ help: if this is intentional, prefix it
      with an underscore: `_result`
         |
         = note: `#[warn(unused_variables)]` on by default

      warning: method `is_initialized` is never used
        --> src/gds/driver.rs:64:12
         |
      27 | impl GdsDriver {
         | -------------- method in this implementation
      ...
      64 |     pub fn is_initialized(&self) -> bool {
         |            ^^^^^^^^^^^^^^
         |
         = note: `#[warn(dead_code)]` on by default

      warning: method `fd` is never used
        --> src/gds/handle.rs:62:12
         |
      21 | impl GdsHandle {
         | -------------- method in this implementation
      ...
      62 |     pub fn fd(&self) -> i32 {
         |            ^^

      warning: field `path` is never read
        --> src/gds/storage.rs:11:5
         |
      9  | pub struct GdsStorage {
         |            ---------- field in this struct
      10 |     handle: GdsHandle,
      11 |     path: PathBuf,
         |     ^^^^

      warning: methods `path` and `size` are never used
        --> src/gds/storage.rs:36:12
         |
      15 | impl GdsStorage {
         | --------------- methods in this implementation
      ...
      36 |     pub fn path(&self) -> &PathBuf {
         |            ^^^^
      ...
      41 |     pub fn size(&self) -> usize {
         |            ^^^^

      warning: creating a shared reference to mutable static
        --> src/gds/driver.rs:44:13
         |
      44 | /             DRIVER_INSTANCE
      45 | |                 .clone()
         | |________________________^ shared reference to mutable static
         |
         = note: for more information, see
      <https://doc.rust-lang.org/nightly/edition-guide/rust-2024/static-mut-references.html>
         = note: shared references to mutable statics are dangerous; it's
      undefined behavior if the static is mutated or if a mutable reference is
      created for it while the shared reference lives
         = note: `#[warn(static_mut_refs)]` on by default

      error: linking with `cc` failed: exit status: 1
        |
        = note:  "cc" "-Wl,--version-script=/tmp/rustcO36qMS/list"
      "-Wl,--no-undefined-version" "-m64" "/tmp/rustcO36qMS/symbols.o"
      "<17 object files omitted>" "-Wl,--as-needed"
      "-Wl,-Bdynamic" "-lcufile" "-Wl,-Bstatic"
      "/tmp/safetensors/bindings/python/target/release/deps/{libsafetensors-6d7dcccfb74f4666.rlib,libserde_json-f8b6f585bf5df12b.rlib,libmemchr-792e9d77b6c57362.rlib,libitoa-eb11a3c454c526bf.rlib,libryu-e987ae88e9011779.rlib,libserde-737e8e7c0e87e24c.rlib,libserde_core-31036d4ca4b1a638.rlib,libpyo3-656dbc47a26c5839.rlib,libmemoffset-7d7bab1c5fa168b5.rlib,libonce_cell-1a293e670803d6dd.rlib,libpyo3_ffi-116dfab57aef71bd.rlib,libunindent-4f30b6354b8a919b.rlib,libmemmap2-846ee8a68c9859d8.rlib,liblibc-994a2b65212047a6.rlib}.rlib"
      "<sysroot>/lib/rustlib/x86_64-unknown-linux-gnu/lib/{libstd-*,libpanic_unwind-*,libobject-*,libmemchr-*,libaddr2line-*,libgimli-*,librustc_demangle-*,libstd_detect-*,libhashbrown-*,librustc_std_workspace_alloc-*,libminiz_oxide-*,libadler2-*,libunwind-*,libcfg_if-*,liblibc-*,librustc_std_workspace_core-*,liballoc-*,libcore-*,libcompiler_builtins-*}.rlib"
      "-Wl,-Bdynamic" "-lgcc_s" "-lutil" "-lrt" "-lpthread" "-lm" "-ldl"
      "-lc" "-L" "/tmp/rustcO36qMS/raw-dylibs" "-Wl,--eh-frame-hdr"
      "-Wl,-z,noexecstack" "-L" "<sysroot>/local/cuda/lib64" "-L"
      "<sysroot>/local/cuda-12/lib64" "-L" "<sysroot>/local/cuda-13.0/lib64"
      "-L" "<sysroot>/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-o"
      "/tmp/safetensors/bindings/python/target/release/deps/libsafetensors_rust.so"
      "-Wl,--gc-sections" "-shared" "-Wl,-z,relro,-z,now" "-Wl,-O1"
      "-Wl,--strip-debug" "-nodefaultlibs"
        = note: some arguments are omitted. use `--verbose` to show all linker
      arguments
        = note: /usr/bin/ld: cannot find -lcufile: No such file or directory
                collect2: error: ld returned 1 exit status
      

      warning: `safetensors-python` (lib) generated 9 warnings
      error: could not compile `safetensors-python` (lib) due to 1 previous
      error; 9 warnings emitted
      💥 maturin failed
```

maifeeulasad:
Excellent edge case, now I have added a feature as rail guard to prevent leaking GDS code into regular build. Please check. So GDS features need to build explicitly, which is even better, as per my understanding.

Updated notebook can be found in the previous url, here goes the url for you convinience: https://gist.github.com/maifeeulasad/741df3cf99a9970ce7a4897d5583b223#file-safetensors-benchmarking-withand-without-gds-ipynb

Open for further review, discussion, exploration.

frauttauteffasu:
Please explain why cupy is required. What limitations are there in the pytorch API that makes it impossible to implement without adding a dependency on cupy?

bindings/python/Cargo.toml
```diff
@@ -9,6 +9,10 @@ rust-version = "1.74"
 name = "safetensors_rust"
 crate-type = ["cdylib"]
 
+[features]
+default = []
+cuda-gds = []
```
frauttauteffasu, line 14: As this is not exposed in the python binding, it forces the user to manually invoke maturin which contradicts the README.

McPatate:
Hi, we won't move along with this PR, we're in the midst of building an optimised loading backend for safetensors which will eventually support GDS directly. Closing for now.

PR #678, opened 2025-11-25
Add arm64 windows support

### LINKED ISSUE #677: Add windows arm64 support
### Feature request

Now that github actions has added [windows arm64 runner support](https://github.blog/changelog/2025-04-14-windows-arm64-hosted-runners-now-available-in-public-preview/) for public repos (`windows-11-arm`) and both numpy and pytorch have releases for arm64 windows, would there be any appetite for supporting windows arm64 wheels?

### Motivation

There is currently no native wheel available for safetensors on arm64 windows.

### Your contribution

I'd be happy to submit a PR 😃

finnagin:
# What does this PR do?

Fixes #677 

This would add tests and builds for arm64 windows wheels.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -2,7 +2,6 @@ name: Python
 
 on:
   pull_request:
-  push:
 
 jobs:
   build_and_test:


.github/workflows/python.yml
```diff
@@ -125,12 +138,19 @@ jobs:
...
           cargo test
           pip install ".[testing]"
           pytest -sv tests/
 
+      - name: Run tests (Windows arm64)
+        if: matrix.os == 'windows-11-arm'
+        run: |
+          cargo test
+          pip install ".[testing]"
+          pytest -sv tests/ --ignore=tests/test_tf_comparison.py
```
finnagin, line 152: I'm skipping the `test_tf_comparison.py` tests here as tensorflow is not available for arm64 windows, and will fail when it tries to import tensor flow. Should I also skip `test_flax_comparison.py`, `test_mlx_comparison.py`, and `test_paddle_comparison.py` here because they will just be all skipped anyways? Or maybe specify just the simple and pt tests like in the free-threaded tests step?

danieldk: I think it's fine to let pytest determine the applicable tests.

PR #679, opened 2025-11-26
add gil free serialize_file(serialize_file_threadable)

TomQunChao:
# What does this PR do?

> Problem Statement: safetensors.torch.serialize_file function will occupy GIL during whole save_file process, even it's writing file, which is time consuming, but no other python threads can execute.

This pr add a function safetensors.torch.serialize_file_threadable which does following things:
1. **Release GIL when writing files**, when it is guaranteed that the tensors to be saved will not be modified during the checkpoint saving process, this function can be used with python threads ^_^
2. **Python only passes a data pointer to the Rust layer**, eliminating the need for complex type conversion and type checking when writing bytes. The benchmark shows that when saving ~100 MB of data, serialize_file takes ~1400 ms, while serialize_file_threadable only needs ~30 ms.

benchmark result of safetensors.torch.save_file vs save_file_threadable

| Name (time in ms)           | Min (ms) | Max (ms) | Mean (ms) | StdDev (ms) | Median (ms) | IQR (ms) | Outliers | OPS   | Rounds | Iterations |
|-----------------------------|----------|----------|-----------|-------------|-------------|----------|----------|--------|--------|------------|
| test_sf_threadable_save_cpu | 53.4114  | 239.5388 | 114.8421  | 75.5853     | 100.2545    | 95.5060  | 1;0      | 8.7076 | 5      | 1          |
| test_sf_save_cpu            | 2213.5468| 3129.8307| 2538.6077 | 354.9364    | 2437.3711   | 396.3095 | 1;0      | 0.3939 | 5      | 1          |

### DIFF
--- a/bindings/python/benches/test_pt.py
+++ b/bindings/python/benches/test_pt.py
@@ -160,6 +160,9 @@ def test_pt_sf_save_cpu(benchmark):
 
     filename = "tmp.safetensors"
 
+    # XXX: On some platforms (tested on Linux x86_64 ext4), writing to an already existing file is slower than creating a new one.
+    # On others, such as MacOS (APFS), it's the opposite. To have more consistent benchmarks,
+    # we ensure the file does not exist before each write, which is also closer to real world usage.
     def setup():
         try:
             os.unlink(filename)
--- a/bindings/python/py_src/safetensors/numpy.py
+++ b/bindings/python/py_src/safetensors/numpy.py
@@ -1,11 +1,28 @@
 import os
 import sys
-from typing import Dict, Optional, Union
+from typing import Dict, List, Optional, Union
 
 import numpy as np
 
 from safetensors import deserialize, safe_open, serialize, serialize_file
-from torch import t
+
+
+def _flatten(
+    tensor_dict: Dict[str, np.ndarray], keep_alive_buffer: List
+) -> Dict[str, Dict]:
+    flattened = {}
+    for k, v in tensor_dict.items():
+        tensor = v
+        if not _is_little_endian(tensor):
+            tensor = tensor.byteswap(inplace=False)
+            keep_alive_buffer.append(tensor)
+        flattened[k] = {
+            "dtype": tensor.dtype.name,
+            "shape": tensor.shape,
+            "data_ptr": tensor.ctypes.data,
+            "data_len": tensor.nbytes,
+        }
+    return flattened
 
 
 def save(
@@ -35,20 +52,8 @@ def save(
     byte_data = save(tensors)
     ```
     """
-    flattened = {}
-    keep_alive_buffer = []
-    for k, v in tensor_dict.items():
-        tensor = v
-        if not _is_little_endian(tensor):
-            tensor = tensor.byteswap(inplace=False)
-            keep_alive_buffer.append(tensor)
-        flattened[k] = {
-            "dtype": tensor.dtype.name,
-            "shape": tensor.shape,
-            "data_ptr": tensor.ctypes.data,
-            "data_len": tensor.nbytes,
-        }
-    serialized = serialize(flattened, metadata=metadata)
+    keep_alive_buffer = []  # to keep byteswapped tensors alive
+    serialized = serialize(_flatten(tensor_dict, keep_alive_buffer), metadata=metadata)
     result = bytes(serialized)
     return result
 
@@ -84,20 +89,10 @@ def save_file(
     save_file(tensors, "model.safetensors")
     ```
     """
-    flattened = {}
     keep_alive_buffer = []  # to keep byteswapped tensors alive
-    for k, v in tensor_dict.items():
-        tensor = v
-        if not _is_little_endian(tensor):
-            tensor = tensor.byteswap(inplace=False)
-            keep_alive_buffer.append(tensor)
-        flattened[k] = {
-            "dtype": tensor.dtype.name,
-            "shape": tensor.shape,
-            "data_ptr": tensor.ctypes.data,
-            "data_len": tensor.nbytes,
-        }
-    serialize_file(flattened, filename, metadata=metadata)
+    serialize_file(
+        _flatten(tensor_dict, keep_alive_buffer), filename, metadata=metadata
+    )
 
 
 def load(data: bytes) -> Dict[str, np.ndarray]:
--- a/bindings/python/src/lib.rs
+++ b/bindings/python/src/lib.rs
@@ -37,6 +37,7 @@ struct TensorRawDataView {
     dtype: Dtype,
     tensor_data_ptr: TensorDataPointer,
 }
+
 impl View for &TensorRawDataView {
     fn dtype(&self) -> Dtype {
         self.dtype
@@ -65,6 +66,7 @@ fn prepare_shape(tensor_desc: &PyBound<PyDict>) -> PyResult<Vec<usize>> {
         .ok_or_else(|| SafetensorError::new_err(format!("Missing `shape` in {tensor_desc}")))?
         .extract()
 }
+
 fn prepare_dtype(tensor_desc: &PyBound<PyDict>) -> PyResult<Dtype> {
     let pydtype = tensor_desc
         .get_item("dtype")?
@@ -97,6 +99,7 @@ fn prepare_dtype(tensor_desc: &PyBound<PyDict>) -> PyResult<Dtype> {
     };
     Ok(dtype)
 }
+
 fn prepare_tensor_raw_data_view(
     tensor_dict: HashMap<String, PyBound<PyDict>>,
 ) -> PyResult<HashMap<String, TensorRawDataView>> {
@@ -136,10 +139,17 @@ fn prepare_tensor_raw_data_view(
 
 /// Serializes raw data.
 ///
+/// NOTE: the caller is required to ensure any pointer passed via `data_ptr` is valid and will live
+/// long enough for the duration of the serialization.
+/// We will remove the need for the caller to hold references themselves when we drop support for
+/// python versions prior to 3.11 where the `PyBuffer` API is available.
+/// Creating a `PyBuffer` will enable us to hold a reference to each passed in data array,
+/// increasing its ref count preventing the gc from collecting it while we serialize.
+///
 /// Args:
 ///     tensor_dict (`Dict[str, Dict[Any]]`):
 ///         The tensor dict is like:
-///             {"tensor_name": {"dtype": "F32", "shape": [2, 3], "data": b"\0\0"}}
+///             {"tensor_name": {"dtype": "F32", "shape": [2, 3], "data_ptr": 1234, "data_len": 24}}
 ///     metadata (`Dict[str, str]`, *optional*):
 ///         The optional purely text annotations
 ///
@@ -163,10 +173,17 @@ fn serialize<'b>(
 
 /// Serializes raw data into file.
 ///
+/// NOTE: the caller is required to ensure any pointer passed via `data_ptr` is valid and will live
+/// long enough for the duration of the serialization.
+/// We will remove the need for the caller to hold references themselves when we drop support for
+/// python versions prior to 3.11 where the `PyBuffer` API is available.
+/// Creating a `PyBuffer` will enable us to hold a reference to each passed in data array,
+/// increasing its ref count preventing the gc from collecting it while we serialize.
+///
 /// Args:
 ///     tensor_dict (`Dict[str, Dict[Any]]`):
 ///         The tensor dict is like:
-///             {"tensor_name": {"dtype": "F32", "shape": [2, 3], "data": b"\0\0"}}
+///             {"tensor_name": {"dtype": "F32", "shape": [2, 3], "data_ptr": 1234, "data_len": 24}}
 ///     filename (`str`, or `os.PathLike`):
 ///         The name of the file to write into.
 ///     metadata (`Dict[str, str]`, *optional*):
--- a/safetensors/src/tensor.rs
+++ b/safetensors/src/tensor.rs
@@ -347,43 +347,9 @@ where
         return Err(SafeTensorError::HeaderTooLarge);
     }
 
-    // XXX: On Windows, we write to a temporary file first and then rename it.
-    // This avoids "error 1224" (ERROR_USER_MAPPED_FILE) which occurs when
-    // trying to write to a file that has an active memory-mapped section.
-    #[cfg(windows)]
-    {
-        use std::time::{SystemTime, UNIX_EPOCH};
+    buffered_write_to_file(filename, n, &header_bytes, &tensors)?;
 
-        let timestamp = SystemTime::now()
-            .duration_since(UNIX_EPOCH)
-            .map(|d| d.as_nanos())
-            .unwrap_or(0);
-        let process_id = std::process::id();
-        let temp_name = format!(".safetensors_tmp_{}_{}.tmp", process_id, timestamp);
-
-        let temp_path =
-            if let Some(parent) = filename.parent().filter(|p| !p.as_os_str().is_empty()) {
-                parent.join(&temp_name)
-            } else {
-                std::path::PathBuf::from(&temp_name)
-            };
-
-        buffered_write_to_file(&temp_path, n, &header_bytes, &tensors)?;
-
-        if let Err(e) = std::fs::rename(&temp_path, filename) {
-            let _ = std::fs::remove_file(&temp_path);
-            return Err(e.into());
-        }
-
-        Ok(())
-    }
-
-    #[cfg(not(windows))]
-    {
-        buffered_write_to_file(filename, n, &header_bytes, &tensors)?;
-
-        Ok(())
-    }
+    Ok(())
 }
 
 /// A structure owning some metadata to lookup tensors on a shared `data`
--- a/bindings/python/py_src/safetensors/__init__.pyi
+++ b/bindings/python/py_src/safetensors/__init__.pyi
@@ -20,10 +20,17 @@ def serialize(tensor_dict, metadata=None):
     """
     Serializes raw data.
 
+    NOTE: the caller is required to ensure any pointer passed via `data_ptr` is valid and will live
+    long enough for the duration of the serialization.
+    We will remove the need for the caller to hold references themselves when we drop support for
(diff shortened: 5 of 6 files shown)


README.md
```diff
@@ -1,195 +1,3 @@
...
-
-Rust
-[![Crates.io](https://img.shields.io/crates/v/safetensors.svg)](https://crates.io/crates/safetensors)
-[![Documentation](https://docs.rs/safetensors/badge.svg)](https://docs.rs/safetensors/)
-[![Codecov](https://codecov.io/github/huggingface/safetensors/coverage.svg?branch=main)](https://codecov.io/gh/huggingface/safetensors)
-[![Dependency status](https://deps.rs/repo/github/huggingface/safetensors/status.svg?path=safetensors)](https://deps.rs/repo/github/huggingface/safetensors?path=safetensors)
-
 # safetensors
 
-## Safetensors
```
danieldk, line 25: Thank you for submitting this! Could you please, before anyone reviews the PR:

- Check that the PR is in a good shape for submission. E.g. here a large part of the README is deleted), one of the new docs contains an ASCII emoji, etc.
- Make sure that the files are properly formatted with `cargo fmt`.
- Provide some reproducible data/benchmarks to verify.
- Split a change into multiple PRs if there are changes that can be separated into smaller units.
- Check if extending the API is really necessary.

TomQunChao: Thank you very much for the suggestions!
I have reverted all unintended changes (e.g., the README deletions and the ASCII emoji in the doc) and run cargo fmt on every touched file.
I also added a benchmark that compares the performance of the two `save_file` variants; the numbers are pasted in the PR body.
Regarding the necessity of the new API:
- In some training setups—at least in mine—check-pointing has to happen asynchronously.
  The current `safetensors.torch.save_file` never releases the GIL, so any background thread that calls it blocks every other Python thread for the entire serialization window.
An async-friendly entry point is therefore useful.
- The “thread-able” version is measurably faster on the benchmark, but it requires the user to guarantee that the CPU tensors being serialized will not be mutated during the call.
  That restriction makes it a non-drop-in replacement for the original function, so exposing it under a separate name keeps existing code safe while letting new code opt in to the extra performance.
- I’m still new to the safetensors code base, so I don’t have a clear picture of the long-term maintenance cost.
If the maintainers decide that the extra API surface is unacceptable, I’m happy to fold the functionality into the existing save_file behind an optional flag (or any other design you prefer).

Please let me know what direction you’d like me to take next.

bindings/python/src/lib.rs
```diff
@@ -48,50 +49,133 @@ impl View for &PyView<'_> {
...
+    tensor_dict: HashMap<String, PyBound<PyDict>>,
+    zero_copy: bool,
+) -> PyResult<HashMap<String, NdarrayView>> {
     let mut tensors = HashMap::with_capacity(tensor_dict.len());
     for (tensor_name, tensor_desc) in tensor_dict {
-        let mut shape: Vec<usize> = tensor_desc
-            .get_item("shape")?
+        let data_ptr: u64 = tensor_desc
+            .get_item("data_ptr")?
             .ok_or_else(|| SafetensorError::new_err(format!("Missing `shape` in {tensor_desc}")))?
```
ArthurZucker: error is wrong here
```suggestion
            .ok_or_else(|| SafetensorError::new_err(format!("Missing `data_ptr` in {tensor_desc}")))?
```

bindings/python/src/lib.rs
```diff
@@ -48,50 +49,133 @@ impl View for &PyView<'_> {
...
 
-fn prepare(tensor_dict: HashMap<String, PyBound<PyDict>>) -> PyResult<HashMap<String, PyView>> {
+struct NdarrayView {
+    shape: Vec<usize>,
+    dtype: Dtype,
+    data: Vec<u8>,
+    data_len: usize,
+    data_ptr: u64,
+    contained_data: bool,
+}
```
McPatate:
```suggestion
struct NumpyDataPointer {
  addr: u64,
  len: usize,
}

enum TensorData {
  Owned(Vec<u8>),
  Pointer(NumpyDataPointer),
}

struct NdarrayView {
    shape: Vec<usize>,
    dtype: Dtype,
    tensor_data: TensorData,
}
```

this way you don't need the `contained_data: bool` flag and you can simply `match self.tensor_data { // ... }` in the `impl View for &NdarrayView`.

bindings/python/src/lib.rs
```diff
@@ -48,50 +49,133 @@ impl View for &PyView<'_> {
...
+    fn data(&self) -> Cow<'_, [u8]> {
+        if self.contained_data {
+            Cow::Borrowed(&self.data)
+        } else {
+            let p = self.data_ptr as *const u8;
+            unsafe {
+                let slice = slice::from_raw_parts(p, self.data_len);
+                Cow::Borrowed(slice)
+            }
+        }
```
McPatate:
```suggestion
        match &self.tensor_data {
            TensorData::Owned(data) => Cow::borrowed(&data),
            TensorData::Pointer(ptr) => {
                let p = ptr.addr as *const u8;
                unsafe {
                    let slice = slice::from_raw_parts(p, ptr.len);
                    Cow::borrowed(slice)
                }
            },
        }
```
Not sure the `&` is needed in the `match` stmt, I'll let you check!

bindings/python/src/lib.rs
```diff
@@ -48,50 +49,133 @@ impl View for &PyView<'_> {
...
+    }
+
+    fn data_len(&self) -> usize {
+        self.data_len
```
McPatate:
```suggestion
        match &self.tensor_data {
            Owned(data) => data.len(),
            Pointer(ptr) => ptr.len,
        }
```
same here for the ref, you can run clippy to make sure it doesn't bark at you.

bindings/python/src/lib.rs
```diff
@@ -48,50 +49,133 @@ impl View for &PyView<'_> {
...
+            unsafe { slice::from_raw_parts(p, data_len).to_vec() }
+        };
+        let tensor = NdarrayView {
+            shape,
+            dtype,
+            data,
+            data_len,
+            data_ptr,
+            contained_data: !zero_copy,
+        };
```
McPatate:
```suggestion
        let tensor_data = if zero_copy {
            TensorData::new_pointer(data_ptr, data_len)
        } else {
            let p = data_ptr as *const u8;
            let data = unsafe { slice::from_raw_parts(p, data_len).to_vec() };
            TensorData::new_owned(data)
        };
        let tensor = NdarrayView {
            shape,
            dtype,
            tensor_data,
        };
```

bindings/python/benches/test_pt.py
```diff
@@ -87,6 +87,26 @@ def test_pt_sf_load_cpu_small(benchmark):
         tv = result[k]
         assert torch.allclose(v, tv)
 
+def test_sf_threadable_save_cpu(benchmark):
```
McPatate:
```suggestion
def test_pt_sf_threadable_save_cpu(benchmark):
```
Also small nit, could you move both `save_cpu` tests to the bottom of the file? I'd rather have benchmarks ordered by function, in our case load first and then save.

bindings/python/benches/test_pt.py
```diff
@@ -87,6 +87,26 @@ def test_pt_sf_load_cpu_small(benchmark):
...
+    weights = create_gpt2(12)
+    
+    # Benchmark save_file_threadable
+    with tempfile.NamedTemporaryFile(delete=False) as f_threadable:
+        benchmark(save_file_threadable, weights, f_threadable.name)
+    
+    # Clean up files
+    os.unlink(f_threadable.name)
+
+def test_sf_save_cpu(benchmark):
```
McPatate:
```suggestion
def test_pt_sf_save_cpu(benchmark):
```

bindings/python/benches/test_pt.py
McPatate, line 1: Could you add this benchmark as well:
```py
def test_sf_threadable_save_cpu_no_zero_copy(benchmark):
    # benchmark save_file_threadable vs save_file
    weights = create_gpt2(12)

    # Benchmark save_file_threadable
    with tempfile.NamedTemporaryFile(delete=False) as f_threadable:
        benchmark(save_file_threadable, weights, f_threadable.name, None, False)

    # Clean up files
    os.unlink(f_threadable.name)
```

McPatate (commented):
These are the results I'm getting from running the benchmarks (`python3 -m pytest benches/test_pt.py -k save_cpu`) in your repo in `feat/gil_free_serialize_file` (with my added no zero copy test):
```
-------------------------------------------------------------------------------------------- benchmark: 3 tests -------------------------------------------------------------------------------------------
Name (time in s)                                 Min                Max               Mean            StdDev             Median               IQR            Outliers     OPS            Rounds  Iterations
test_sf_threadable_save_cpu                   4.1726 (1.0)       4.1866 (1.0)       4.1789 (1.0)      0.0053 (1.0)       4.1773 (1.0)      0.0073 (1.0)           2;0  0.2393 (1.0)           5           1
test_sf_threadable_save_cpu_no_zero_copy      4.1736 (1.00)      4.1895 (1.00)      4.1818 (1.00)     0.0067 (1.25)      4.1798 (1.00)     0.0110 (1.50)          2;0  0.2391 (1.00)          5           1
test_sf_save_cpu                             11.2521 (2.70)     12.0204 (2.87)     11.6914 (2.80)     0.3391 (63.56)    11.8754 (2.84)     0.5627 (77.08)         1;0  0.0855 (0.36)          5           1
```
Which seem a bit more realistic compared to 22 fold speedup 😄 
Still very impressive results, kudos! Although I'm having trouble understanding why the no zero copy is the same speed as the zero copy one. Any idea what is going on here?

I ran the CI, there are some lint errors, could you fix them?

I think in follow-up PRs we could enable this with different libraries, would you be keen to add this @TomQunChao?

bindings/python/benches/test_pt.py
McPatate: you can ignore this!

bindings/python/tests/test_threadable.py
McPatate, line 1: before merging, we'll need to make extra sure files are exactly the same and that there is no regression.

You can remove `test_consistancy`, I'll write a script tmrw to serialize existing weights we have on the Hub to make sure there is no regression. We could definitely automate this and add a sort of `integration` test for consistency checks, but perhaps best to do in a subsequent PR.
If you have the bandwidth feel free, otherwise I'll tackle that at some point after your PR is merged.

bindings/python/src/lib.rs
McPatate: This too won't make much sense if you remove the non zero copy case. Feel free to just keep `data_ptr` & `data_len`.

bindings/python/src/lib.rs
TomQunChao: awesome, learned a new rust feature✍️ ^_^

bindings/python/benches/test_pt.py
TomQunChao: Thanks for the suggestions!
Just to confirm — are you referring to enabling this feature in other libraries such as transformers?
If so, I’d be very happy to work on that!

bindings/python/src/lib.rs
TomQunChao: ok, I'll check it.

TomQunChao:
@McPatate I've refactored the relevant parts in torch.py and lib.rs as you suggested — removing PyView, _flatten, and _tobytes, and replacing the old save_file with the GIL-free version.

However, I realized the workload is a bit larger than I expected, because the changes also need to be applied to NumPy/Paddle/TensorFlow backends. Unfortunately, I won’t be able to spend more time on this this week due to some other commitments.

Would you be able to help with the remaining changes?
If not, I’ll have time next Tuesday and can finish the rest then.

McPatate (approved):
I'm personally good with the state of this PR, would appreciate a review if you have time @danieldk

safetensors/src/tensor.rs
```diff
@@ -326,17 +347,43 @@ where
         return Err(SafeTensorError::HeaderTooLarge);
     }
 
-    let mut f = std::io::BufWriter::new(std::fs::File::create(filename)?);
-    f.write_all(n.to_le_bytes().as_ref())?;
-    f.write_all(&header_bytes)?;
+    // XXX: On Windows, we write to a temporary file first and then rename it.
+    // This avoids "error 1224" (ERROR_USER_MAPPED_FILE) which occurs when
+    // trying to write to a file that has an active memory-mapped section.
```
McPatate: Not 100% we want to include this, this mostly broke tests that read and wrote to the same file. I would err on the side of caution and leave it personally, as I don't believe writing then renaming would hurt performance.

bindings/python/benches/test_pt.py
```diff
@@ -152,3 +153,22 @@ def test_pt_sf_load_mps(benchmark):
...
+
+
+def test_pt_sf_save_cpu(benchmark):
+    weights = create_gpt2(12)
+
+    filename = "tmp.safetensors"
+
+    def setup():
+        try:
+            os.unlink(filename)
```
danieldk, line 168: I think it's worth adding a comment here, I'm sure we have forgotten why we do this in a year 😅.

bindings/python/py_src/safetensors/numpy.py
```diff
@@ -5,12 +5,7 @@
 import numpy as np
 
 from safetensors import deserialize, safe_open, serialize, serialize_file
-
-
-def _tobytes(tensor: np.ndarray) -> bytes:
-    if not _is_little_endian(tensor):
-        tensor = tensor.byteswap(inplace=False)
-    return tensor.tobytes()
+from torch import t
```
danieldk: This adds `torch` as a dependency, we can import torch in code paths that really require Torch.

danieldk: It's not used either if I'm not mistaken?

bindings/python/py_src/safetensors/numpy.py
```diff
@@ -40,10 +35,19 @@ def save(
...
-        k: {"dtype": v.dtype.name, "shape": v.shape, "data": _tobytes(v)}
-        for k, v in tensor_dict.items()
-    }
+    flattened = {}
+    keep_alive_buffer = []
+    for k, v in tensor_dict.items():
+        tensor = v
+        if not _is_little_endian(tensor):
+            tensor = tensor.byteswap(inplace=False)
+            keep_alive_buffer.append(tensor)
```
danieldk:
```suggestion
            keep_alive_buffer.append(tensor)  # to keep byteswapped tensors alive
```

bindings/python/py_src/safetensors/numpy.py
```diff
@@ -80,10 +84,19 @@ def save_file(
...
+        flattened[k] = {
+            "dtype": tensor.dtype.name,
+            "shape": tensor.shape,
+            "data_ptr": tensor.ctypes.data,
+            "data_len": tensor.nbytes,
+        }
```
danieldk: Nit: exactly the same as above, maybe nicer to factor out as a function that returns a tuple `flattened, keep_alive_buffer`?

bindings/python/src/lib.rs
```diff
@@ -26,82 +27,107 @@ static FLAX_MODULE: OnceLock<Py<PyModule>> = OnceLock::new();
 static MLX_MODULE: OnceLock<Py<PyModule>> = OnceLock::new();
 static PADDLE_MODULE: OnceLock<Py<PyModule>> = OnceLock::new();
 
-struct PyView<'a> {
+struct TensorDataPointer {
+    addr: u64,
+    len: usize,
+}
+
+struct TensorRawDataView {
```
danieldk, line 35: I might be missing something, but since data is now a raw pointer, shouldn't `TensorRawDataView` have a lifetime bound, tying it to the lifetime of the actual array? Or maybe even in `TensorDataPointer`

bindings/python/src/lib.rs
```diff
@@ -26,82 +27,107 @@ static FLAX_MODULE: OnceLock<Py<PyModule>> = OnceLock::new();
...
+        "float4_e2m1fn_x2" => Dtype::F4,
+        "complex64" => Dtype::C64,
+        dtype_str => {
+            return Err(SafetensorError::new_err(format!(
+                "dtype {dtype_str} is not covered",
+            )));
+        }
+    };
+    Ok(dtype)
+}
```
danieldk, line 101: Nit

```suggestion
}

```

safetensors/src/tensor.rs
```diff
@@ -326,17 +347,43 @@ where
...
-    for tensor in tensors {
-        f.write_all(tensor.data().as_ref())?;
+        buffered_write_to_file(&temp_path, n, &header_bytes, &tensors)?;
+
+        if let Err(e) = std::fs::rename(&temp_path, filename) {
+            let _ = std::fs::remove_file(&temp_path);
+            return Err(e.into());
+        }
+
+        Ok(())
```
danieldk: This seems to add more complicated logic and orthogonal to the rest of the PR? Also, doesn't nicely clean up if writing fails before getting to the removal and the user gets potentially large hidden files.

I think the original Windows error is nice, it tells people to just not write to a memory-mapped file.

bindings/python/benches/test_pt.py
McPatate: Ahah noted

bindings/python/src/lib.rs
McPatate: funny rustfmt doesn't take care of that automatically 🤔

bindings/python/py_src/safetensors/numpy.py
McPatate: unused indeed, removing

bindings/python/py_src/safetensors/numpy.py
McPatate: To be consistent with the other flatten fn, I'll pass in a ref to the keep alive buffer

bindings/python/src/lib.rs
McPatate: We don't have access to `PyBuffer` in rust since we want to support python versions pre 3.11. Will add a comment to document why we're doing things this way.

safetensors/src/tensor.rs
McPatate: updated the test to avoid getting the error!

PR #681, opened 2025-12-03
feat: add direct io backend

McPatate:
## **_Experimental changes to file loading_**

> [!WARNING]
> sorry for the messy thought dump you're about to read

Still measuring performance benefits. Probably implemented the wrong approach (full file loading in host memory), the goal with this impl is to try saturating disk read bandwidth on an nvme.

Running `pytest benches/test_pt.py -k test_pt_sf_load_direct` yields poor performance atm, but it's hard to compare to say `test_pt_sf_load_cpu` given os page cache is populated when writing the file to disk, which is done right before the `benchmark(load_file, ...)` call.

I want to continue experimenting with this loading method, perhaps instead imitating what is done in [fastsafetensors' `nogds_file_reader`](https://github.com/foundation-model-stack/fastsafetensors/blob/d9d23b6c72a3b188e4dbe8fc31f42d03e3beb851/fastsafetensors/cpp/ext.cpp#L470) where they have a bounce buffer allocated with `cudaHostAlloc` from which they run `cudaMemcpy` once they `pread` memory in said buffer.

I'd also like to try playing around with creating a `safetensors-distributed` package that'd be built on top of this infrastructure layer for fast loading of files in tensor parallel scenarios. We'll probably want to port over [some of the code in transformers](https://github.com/huggingface/transformers/blob/ad24f43c00c059ef9341635f102fa38657247ad6/src/transformers/core_model_loading.py#L558) in that package.

I'm not convinced we want to imitate fastsafetensors fully, where they scatter and broadcast tensors from gpu memory leveraging nvlink (iiuc). But at the same time, I'm not sure either we can slice the file like we currently do in our distributed weight loader in transformers while reading file in the bounce buffer, we want reads to be aligned when reading with `O_DIRECT` and this will get messy when sending over bytes to device. Need to investigate some more 🤕

### References

- [fastsafetensors paper](https://arxiv.org/html/2505.23072v1)
- [fastsafetensors repo](https://github.com/foundation-model-stack/fastsafetensors)

### DIFF
--- a/bindings/python/benches/test_pt.py
+++ b/bindings/python/benches/test_pt.py
@@ -2,10 +2,34 @@ import os
 import tempfile
 
 import pytest
-from safetensors import safe_open
 import torch
 
 from safetensors.torch import load_file, save_file
+
+import subprocess
+
+
+def drop_linux_page_cache():
+    """
+    Properly drop Linux page cache using sudo and tee.
+    """
+
+    print("dropping os page cache")
+
+    subprocess.run(["sync"], check=True)
+
+    proc = subprocess.run(
+        ["sudo", "tee", "/proc/sys/vm/drop_caches"],
+        input="3\n",
+        text=True,
+        capture_output=True,
+    )
+
+    print(f"stdout: {proc.stdout}")
+    print(f"stderr: {proc.stderr}")
+
+    if proc.returncode != 0:
+        raise RuntimeError(f"Failed to drop caches: {proc.stderr.strip()}")
 
 
 def create_gpt2(n_layers: int):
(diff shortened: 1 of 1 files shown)


bindings/python/src/lib.rs
```diff
@@ -441,9 +458,166 @@ struct Open {
...
+            let aligned_capacity = file_len.checked_next_multiple_of(ALIGN).ok_or_else(|| {
+                SafetensorError::new_err("overflow while computing aligned file len")
+            })?;
+            let layout = Layout::from_size_align(aligned_capacity, ALIGN)
+                .map_err(|_| SafetensorError::new_err("invalid layout"))?;
+            let ptr = alloc(layout);
+            if ptr.is_null() {
+                handle_alloc_error(layout);
+            }
+            let mut vec: Vec<u8> = Vec::from_raw_parts(ptr, 0, aligned_capacity);
```
danieldk, line 513: I might be mistaken, but I think `Vec` uses the alignment of the type, so you can do this more safely using something like:

```
let mut vec: Vec<[u8; ALIGN]> = Vec::with_capacity((file_len + ALIGN - 1) / ALIGN);
// Use vec as u8 rather than [u8; ALIGN].
```

bindings/python/src/lib.rs
```diff
@@ -441,9 +458,166 @@ struct Open {
...
+                let region_bytes = blocks_per_thread * block_size;
+
+                thread::scope(|scope| -> PyResult<()> {
+                    println!("creating buffer slice of len {aligned_prefix_len}");
+                    let buffer: &mut [u8] =
+                        std::slice::from_raw_parts_mut(vec.as_mut_ptr(), aligned_prefix_len);
+
+                    let mut handles = Vec::new();
+
+                    for (tid, chunk) in buffer.chunks_mut(region_bytes).enumerate() {
```
danieldk, line 534: Question for understanding, since `region_bytes = blocks_per_thread * block_size`. Suppose we have 16 blocks in the file and 4 threads. Then the responsibility by thread is:

```
0 0 0 0 1 1 1 1 2 2 2 2 3 3 3 3
```

I am curious how this compares to patterns like:

```
0 1 2 3 0 1 2 3 0 1 2 3 0 1 2 3 
```

PR #686, opened 2025-12-17
refactor: support reading padded/trailing whitespace header

McPatate:
in the future, we'll potentially want to support producing byte-aligned
tensor offsets. as trailing whitespace are already ignored by serde,
this won't break the format to allow trailing whitespaces. 3rd party
writers do need to take this into account and include the padding len in
the header size bytes at the start of the file. We can and probably will
add support for this in the future (alignment) thanks to this.

### DIFF
--- a/safetensors/src/tensor.rs
+++ b/safetensors/src/tensor.rs
@@ -16,8 +16,6 @@ const N_LEN: usize = size_of::<u64>();
 pub enum SafeTensorError {
     /// The header is an invalid UTF-8 string and cannot be read.
     InvalidHeader(Utf8Error),
-    /// The header's first byte is not the expected `{`.
-    InvalidHeaderStart,
     /// The header does contain a valid string, but it is not valid JSON.
     InvalidHeaderDeserialization(serde_json::Error),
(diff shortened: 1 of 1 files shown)


safetensors/src/tensor.rs
```diff
@@ -385,11 +385,10 @@ impl<'data> SafeTensors<'data> {
             return Err(SafeTensorError::InvalidHeaderLength);
         };
         let string = core::str::from_utf8(header_bytes).map_err(SafeTensorError::InvalidHeader)?;
-        // Assert the string starts with {
-        // NOTE: Add when we move to 0.4.0
-        // if !string.starts_with('{') {
-        //     return Err(SafeTensorError::InvalidHeaderStart);
-        // }
+        // Assert the string starts with `{`
+        if !string.starts_with('{') {
```
danieldk: Is it worth potentially breaking checkpoints over this (not sure if third-party writers made JSON with some initial space)? Maybe it's best to emit a warning for one release, warning that it will be enforced in 0.8 or 0.9?

PR #687, opened 2025-12-17
feat: add direct io write fast path on macos

McPatate:
This uses the `libc::F_NOCACHE` flag on the raw file descriptor to avoid going through the OS page cache. It yields about a 30% performance improvement.

I've been benchmarking different ways to write to disk and have found that setting the aforementioned flag to the file descriptor yields big perf benefits, see the results from https://github.com/huggingface/safetensors/tree/bench/different_write_methods:
```
-------------------------------------------------------------------------------------------- benchmark: 6 tests -------------------------------------------------------------------------------------------
Name (time in ms)                          Min                 Max                Mean             StdDev              Median                IQR            Outliers      OPS            Rounds  Iterations
test_pt_sf_save_cpu                    72.9900 (1.0)      176.8296 (1.11)     112.4869 (1.31)     34.4309 (1.30)      90.5884 (1.14)     65.4580 (11.04)        17;0   8.8899 (0.76)         50           1
test_pt_sf_save_cpu_direct             74.8230 (1.03)     249.0532 (1.56)      85.7739 (1.0)      30.4155 (1.15)      79.1634 (1.0)       5.9283 (1.0)           2;3  11.6586 (1.0)          50           1
test_pt_sf_save_cpu_vectored           80.3192 (1.10)     227.5461 (1.43)     121.0702 (1.41)     37.5942 (1.42)     142.9143 (1.81)     63.4260 (10.70)        13;0   8.2597 (0.71)         50           1
test_pt_sf_save_cpu_oneshot            90.7286 (1.24)     159.1498 (1.0)      122.9442 (1.43)     26.4695 (1.0)      128.4449 (1.62)     50.3530 (8.49)         21;0   8.1338 (0.70)         50           1
test_pt_sf_save_cpu_mmap              124.7373 (1.71)     277.0682 (1.74)     167.3294 (1.95)     39.2798 (1.48)     187.2467 (2.37)     67.1930 (11.33)        13;0   5.9762 (0.51)         50           1
test_pt_sf_save_cpu_parallel_mmap     142.0471 (1.95)     260.1278 (1.63)     205.1932 (2.39)     37.3936 (1.41)     224.3460 (2.83)     73.0463 (12.32)        19;0   4.8735 (0.42)         50           1
```

I believe setting `F_NOCACHE` for write operations should mostly be fine given we'll probably never read-after-write (negating the potential benefits of mmap), curious to hear if there is an existing workflow that needs read-after-write.

### DIFF
--- a/safetensors/src/tensor.rs
+++ b/safetensors/src/tensor.rs
@@ -307,43 +307,32 @@ fn buffered_write_to_file<V: View>(
     n: u64,
     header_bytes: &[u8],
     tensors: &[V],
-) -> Result<(), SafeTensorError> {
-    let mut f = std::io::BufWriter::new(std::fs::File::create(path)?);
-    f.write_all(n.to_le_bytes().as_ref())?;
-    f.write_all(header_bytes)?;
-    for tensor in tensors {
-        f.write_all(tensor.data().as_ref())?;
-    }
-    f.flush()?;
-    Ok(())
-}
-
-#[cfg(all(feature = "std", target_os = "macos"))]
-fn buffered_write_to_file_direct<V: View>(
-    filename: &Path,
-    n: u64,
-    header_bytes: &[u8],
-    tensors: &[V],
     total_size: usize,
 ) -> Result<(), SafeTensorError> {
-    use std::os::fd::AsRawFd;
+    let file = std::fs::File::create(path)?;
 
-    let file = std::fs::OpenOptions::new()
-        .write(true)
-        .create(true)
-        .truncate(true)
-        .open(filename)?;
     file.set_len(total_size as u64)?;
+
+    // Serialize tensors to a file using direct I/O (bypassing page cache) using F_NOCACHE.
+    // This yields ~30% performance improvement.
+    #[cfg(target_os = "macos")]
     unsafe {
+        use std::os::fd::AsRawFd;
+
         libc::fcntl(file.as_raw_fd(), libc::F_NOCACHE, 1);
     }
+
     let mut f = std::io::BufWriter::with_capacity(1024 * 1024, file);
+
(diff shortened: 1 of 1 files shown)


safetensors/src/tensor.rs
```diff
@@ -318,6 +318,35 @@ fn buffered_write_to_file<V: View>(
...
+    total_size: usize,
+) -> Result<(), SafeTensorError> {
+    use std::os::fd::AsRawFd;
+
+    let file = std::fs::OpenOptions::new()
+        .write(true)
+        .create(true)
+        .truncate(true)
+        .open(filename)?;
+    file.set_len(total_size as u64)?;
```
McPatate, line 314: not even sure this is needed, won't hurt though.

safetensors/src/tensor.rs
```diff
@@ -318,6 +318,35 @@ fn buffered_write_to_file<V: View>(
...
+    unsafe {
+        libc::fcntl(file.as_raw_fd(), libc::F_NOCACHE, 1);
+    }
+    let mut f = std::io::BufWriter::with_capacity(1024 * 1024, file);
```
McPatate, line 325: same here, I'm not sure explicitly setting the capacity has any benefits, but I'd rather be explicit so leaving it (tested with simply `BufWriter::new` and results are the same).
I've tested with larger buffers, with 1 to 8 results are about the same but perf degrades with 16 & 32Mb.

safetensors/src/tensor.rs
McPatate: From my tests, this doesn't change performance at all. Ok to remove it if someone feels strongly about it.

danieldk: I like it. Avoids that if the file is too large for the disk, it fails as early as possible.

PR #690, opened 2025-12-19
feat: add io_uring + O_DIRECT write for linux fast path

McPatate:
Following on #687, taking a look at linux now. The performance boost for this vary quite a bit depending on how much data is written at once, but see preliminary results for 5 rounds of benchmarking on gpt2-like model of around ~522mb:
```
-------------------------------------------------------------------------------------------- benchmark: 2 tests -------------------------------------------------------------------------------------------
Name (time in ms)                           Min                 Max                Mean             StdDev              Median                IQR            Outliers     OPS            Rounds  Iterations
test_pt_sf_save_cpu_linux_io_uring     114.4863 (1.0)      141.0581 (1.0)      122.5623 (1.0)      10.9727 (3.20)     116.9135 (1.0)      12.5483 (2.27)          1;0  8.1591 (1.0)           5           1
test_pt_sf_save_cpu                    330.5188 (2.89)     338.7367 (2.40)     334.2241 (2.73)      3.4262 (1.0)      335.1040 (2.87)      5.5184 (1.0)           2;0  2.9920 (0.37)          5           1
```

Consistency seems to be good, produced files are identical in size and have the same `sha1sum`.

This is preliminary work, imo there are still things to do wrt compatibility of the underlying system with io_uring for instance and the tuning (alignment, buf_size). But the performance looks encouraging and we should investigate!

cc @botirk38 would appreicate your feedback!

### DIFF
--- a/safetensors/src/tensor.rs
+++ b/safetensors/src/tensor.rs
@@ -481,8 +481,6 @@ fn linux_io_uring_write<V: View>(
         page_size as usize
     };
 
-    println!("alignment: {alignment}");
-
     let file = std::fs::OpenOptions::new()
         .write(true)
         .create(true)
@@ -602,8 +600,15 @@ fn linux_io_uring_write<V: View>(
             if let Some((buf_idx, expected_len)) = in_flight.remove(&ud) {
                 let result = cqe.result();
                 if result < 0 {
+                    let errno = -result;
+                    if errno == libc::EINVAL {
+                        return Err(SafeTensorError::IoError(std::io::Error::new(
+                            std::io::ErrorKind::Unsupported,
+                            "io_uring Write operation not supported on this kernel (requires 5.6+)",
+                        )));
+                    }
                     return Err(SafeTensorError::IoError(std::io::Error::from_raw_os_error(
-                        -result,
+                        errno,
                     )));
                 }
                 if (result as usize) != expected_len {


safetensors/src/tensor.rs
```diff
@@ -336,6 +349,291 @@ fn buffered_write_to_file<V: View>(
...
+    n: u64,
+    header_bytes: &[u8],
+    tensors: &[V],
+    total_size: usize,
+) -> Result<(), SafeTensorError> {
+    use io_uring::{opcode, types, IoUring};
+    use std::collections::VecDeque;
+    use std::os::{fd::AsRawFd, unix::fs::OpenOptionsExt};
+
+    const ALIGNMENT: usize = 4096;
```
botirkhaltaev: is it safe upstream to hardcode 4kb page alignment, some OSes may have various page sizes. ```libc::sysconf(_SC_PAGESIZE)```

safetensors/src/tensor.rs
```diff
@@ -336,6 +349,291 @@ fn buffered_write_to_file<V: View>(
...
+    const BUF_SIZE: usize = 256 * 1024;
+    const QUEUE_DEPTH: u32 = 64;
+
+    let file = std::fs::OpenOptions::new()
+        .write(true)
+        .create(true)
+        .truncate(true)
```
botirkhaltaev, line 487: again not sure, but if the FS does not support direct I/O this will fail and there is no fallback for this

safetensors/src/tensor.rs
```diff
@@ -336,6 +349,291 @@ fn buffered_write_to_file<V: View>(
...
+
+            data_source.fill_buffer(buffer);
+
+            if buffer.len == 0 {
+                free_buffer_indices.push_front(buf_idx);
+                break;
+            }
+
+            // For O_DIRECT, write length must be aligned.
+            // Only pad the last buffer - intermediate buffers should always be full (BUF_SIZE)
```
botirkhaltaev: im not a fan of using O_DIRECT when we can't emit requests that are fully aligned so we force padding like this, in my benches, its better to let the kernel do a buffered write in this case

safetensors/src/tensor.rs
```diff
@@ -336,6 +349,291 @@ fn buffered_write_to_file<V: View>(
...
+                        format!("short write: expected {}, got {}", expected_len, result),
+                    )));
+                }
+                free_buffer_indices.push_back(buf_idx);
+            }
+        }
+    }
+
+    // Truncate to actual size (we potentially padded the last buffer for O_DIRECT alignment)
+    unsafe {
```
botirkhaltaev: yea this is a consequence of padding, ideally as above, two paths when requests are naturally aligned and when not to do buffered writes, so we don't end up in this scenario

safetensors/src/tensor.rs
```diff
@@ -336,6 +349,291 @@ fn buffered_write_to_file<V: View>(
...
+                    return Err(SafeTensorError::IoError(std::io::Error::from_raw_os_error(
+                        -result,
+                    )));
+                }
+                if (result as usize) != expected_len {
+                    return Err(SafeTensorError::IoError(std::io::Error::new(
+                        std::io::ErrorKind::WriteZero,
+                        format!("short write: expected {}, got {}", expected_len, result),
+                    )));
+                }
```
botirkhaltaev, line 619: i also think we are missing a fsync :), never trust the kernel!!!

McPatate: that may explain the huge perf boost lol, let me check

safetensors/src/tensor.rs
```diff
@@ -336,6 +349,291 @@ fn buffered_write_to_file<V: View>(
...
+        .open(path.as_ref())?;
+
+    file.set_len(total_size.next_multiple_of(ALIGNMENT) as u64)?;
+
+    let fd = types::Fd(file.as_raw_fd());
+
+    let mut ring = IoUring::new(QUEUE_DEPTH)
+        .map_err(|e| SafeTensorError::IoError(std::io::Error::other(e)))?;
+
+    let mut buffers: Vec<AlignedBuffer> = Vec::with_capacity(QUEUE_DEPTH as usize);
```
botirkhaltaev, line 509: might be worth to register these with io_uring so we can avoid repinning and revalidating each time, but that's a micro thing

safetensors/src/tensor.rs
```diff
@@ -336,6 +349,291 @@ fn buffered_write_to_file<V: View>(
...
+    use io_uring::{opcode, types, IoUring};
+    use std::collections::VecDeque;
+    use std::os::{fd::AsRawFd, unix::fs::OpenOptionsExt};
+
+    const ALIGNMENT: usize = 4096;
+    const BUF_SIZE: usize = 256 * 1024;
+    const QUEUE_DEPTH: u32 = 64;
+
+    let file = std::fs::OpenOptions::new()
+        .write(true)
```
botirkhaltaev, line 485: if we grab the fd from the file we can register that with io_uring as well but may not be important here as well

McPatate:
closing in favour of our hmll work

PR #704, opened 2026-02-08
Make SafetensorError pickle-able (fixes #684)

### LINKED ISSUE #684: SafetensorError isn't pickle-able, it should be (smoke test for issue `import of module 'safetensors_rust' failed`)
### System Info

(`transformers-cli env` breaks with `jax` issues right now, so here's the relevant version)
```
safetensors==0.7.0
```

### Information

- [x] My own modified scripts

### Reproduction

```
from safetensors import SafetensorError
import pickle
pickle.dumps(SafetensorError())
```

fails with

```
  File "<stdin>", line 1, in <module>
_pickle.PicklingError: Can't pickle <class 'safetensors_rust.SafetensorError'>: import of module 'safetensors_rust' failed
```

### Expected behavior

many integrations expect errors to be picklable or serializable. they don't have to be, but they reduce pain everywhere when they are.

another POV is this is probably a smoke test for a deeper issue that may be relevant

jayzuccarelli:
Fixes #684

**Problem:** `SafetensorError` is defined in the Rust extension (`_safetensors_rust`), so `pickle.dumps(SafetensorError())` fails (e.g. "Can't pickle: import of module 'safetensors_rust' failed"). Many integrations (multiprocessing, joblib, distributed workers) expect exceptions to be picklable.

**Solution:** Introduce a Python subclass of the Rust `SafetensorError` that implements `__reduce__`, and export it as `SafetensorError`. Exceptions raised from Rust are still caught by `except SafetensorError` (same type hierarchy). Pickling/unpickling now uses the Python class, so the Rust module is not required when unpickling.

- `bindings/python/py_src/safetensors/__init__.py`: Python `SafetensorError` subclass with `__reduce__(self) -> (SafetensorError, self.args)`.
- `bindings/python/tests/test_safetensor_error_pickle.py`: Tests for pickle round-trip (empty and with message).

### DIFF
--- a/bindings/python/py_src/safetensors/__init__.py
+++ b/bindings/python/py_src/safetensors/__init__.py
@@ -1,6 +1,6 @@
 # Re-export this
 from ._safetensors_rust import (  # noqa: F401
-    SafetensorError,
+    SafetensorError as _RustSafetensorError,
     __version__,
     deserialize,
     safe_open,
@@ -8,3 +8,13 @@ from ._safetensors_rust import (  # noqa: F401
     serialize,
     serialize_file,
 )
+
+
+class SafetensorError(_RustSafetensorError):
+    """
+    Custom Python Exception for Safetensor errors.
+    Subclasses the Rust exception so it remains picklable (e.g. for multiprocessing).
+    """
+
+    def __reduce__(self):
+        return (SafetensorError, self.args)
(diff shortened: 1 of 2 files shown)


McPatate:
Hi, thank you for taking a look. I'm just wondering, why do we need anything safetensors related to pickle-able?

McPatate:
closing in favour of #753

PR #713, opened 2026-03-02
feat: add hmll io uring multi device

McPatate:
hmll backend with io_uring support

TODO: PR desc

### DIFF
--- a/bindings/python/Cargo.toml
+++ b/bindings/python/Cargo.toml
@@ -13,6 +13,11 @@ crate-type = ["cdylib"]
 pyo3 = { version = "0.25", features = ["abi3", "abi3-py38"] }
 memmap2 = "0.9"
 serde_json = "1.0"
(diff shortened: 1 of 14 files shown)


changhoon-sung:
Could you share any plans for the future development of hmll for safetensors? Ultimately, is there an intention to integrate it into safetensors, or will it continue to develop as a separate library with independent development?

McPatate:
Sorry for the delay! Atm, it isn't clear what we're going to do for these kinds of optims.

changhoon-sung:
It seems like these optims could improve performance for model loading and host-offloaded tensors. Are there any technical or maintenance concerns that make them difficult to adopt?

McPatate:
The issue is simply with integrating `hmll` vs hand-rolling a rust backend. This is what I meant by "it isn't clear what we're going to do". But definitely will be spending time on making the library faster / more efficient!

I'm currently working on tokenizers, but in the coming weeks I'll circle back on this!

PR #728, opened 2026-04-11
Opt in to disable PyTorch file-backed storage to fix OOM with unified memory

woct0rdho:
## Motivation

See the previous discussion in https://github.com/huggingface/transformers/issues/44756 .

When loading large models to GPU, some unified memory systems (such as Strix Halo) behave poorly with PyTorch's file-backed storage. For example, the Qwen3.5 35B fp16 model takes 67 GB memory, and I have 125 GB TTM memory, so it should not OOM. But when I test it, it loads very slowly and eventually OOMs.

If I understand correctly, this is because two copies of the model are taking TTM memory, one for GPU and the other for file system cache. We can avoid this using `Storage::Mmap` rather than `torch.ByteStorage.from_file`, which moves each tensor to GPU with only a short-lived CPU buffer. Note that previously people would understand it as 'disable mmap' but actually it's not.

## Implementation

I propose an environment variable `SAFETENSORS_DISABLE_TORCH_FROM_FILE` to opt in this feature. Also, previously `Storage::Mmap` did not support indexing with ellipsis and I implemented it, which is used in transformers.

After this PR, it takes about 30s to load the Qwen3.5 35B fp16 model without OOM, and we no longer need the workaround in https://github.com/huggingface/transformers/issues/44756 and in [ComfyUI](https://github.com/Comfy-Org/ComfyUI/blob/593be209a45a8a306c26de550e240a363de405a7/comfy/utils.py#L137).

I've tested the inference speed of some smaller models on GPU before and after this PR, and there is no regression.

When viewing the diff on GitHub, it's better to enable 'Hide whitespace'.

I'd like to ask users of other unified memory systems (such as Mac and DGX Spark) to test this PR. Also, I'm not sure what's the best way to automatically run the relevant unit tests with `SAFETENSORS_DISABLE_TORCH_FROM_FILE`.

## AI model use

version that you used: GPT-5.4


woct0rdho:
Hi @McPatate , do you think this will be addressed in hmll?

RedL0tus:
Can confirm it works on GB10 (NVIDIA DGX Spark) as well. Would be great if it can get merged.

McPatate:
Fix should be in, I'll close the PR for now. Feel free to re-open if #760 didn't do what you need.

woct0rdho:
I can confirm that after #760, `backend="pread"` does not OOM, but `backend="mmap"` still OOMs on Strix Halo.

I think an env var makes more sense than passing `backend` in all downstream frameworks including transformers, ComfyUI, and more. I've rebased it in https://github.com/woct0rdho/safetensors/tree/mmap . @McPatate @johnnynunez Do you think so?

McPatate:
I'm more inclined towards an explicit param than an env var. Paves the way for the new backend we're working on as well.

Strange that mmap still OOMs though, is it CUDA + torch?

woct0rdho:
It's ROCm + torch. `backend="mmap"` is like `SAFETENSORS_DISABLE_TORCH_FROM_FILE=0` in this PR, which uses torch from_file.

McPatate:
> It's ROCm + torch. backend="mmap" is like SAFETENSORS_DISABLE_TORCH_FROM_FILE=0 in this PR, which uses torch from_file.

I see, indeed the mmap fix I have is for cuda only. Could be worth revisiting your PR for other targets.

woct0rdho:
I guess Mac already works well after #756 , so the problem only remains for Strix Halo. In the safetensors repo I don't see it handling ROCm and CUDA differently. Maybe we should somehow fix torch from_file or amdgpu driver so it provides the same mmap semantics for ROCm as for CUDA.

McPatate:
Does the `pread` backend not do the job for the moment? Do you specifically need mmap?

woct0rdho:
The pread backend does the job for the moment. I just think it's still worth investigating that the mmap backend now does not cause the problem for DGX Spark but causes the problem for Strix Halo. Also, the mmap backend is the default, so it takes some effort to pass the pread backend in downstream frameworks.

McPatate:
I added a special "bypass" for cuda + torch, so it makes sense that rocm + torch with mmap still fails. We write in a `PinnedCpuDest` which is initialized via torch `pin_memory=True` and then send via `.to(device)` rather than relying on torch's untyped storage.

Could be worth investigating on rocm devices, would have to think how to handle each `(device, framework)` pair a bit more idiomatically. Perhaps memory pinning works out of the box with rocm and it's a simple fix? don't have any amd gpus on hand to test though.

PR #734, opened 2026-04-13
fix(torch): cover f8 & f4 in big endian & unskip tests

McPatate:
Two tests were skipped after the hasattr PR fix landed, `float8` and `float4` are not real torch dtypes.

`torch.float8_e8m0fnu` and `torch.float4_e2m1fn_x2` were missing the `NPDTYPES` table, it would've raised a `KeyError` on s390x.


danieldk (approved):
Nice!

PR #735, opened 2026-04-13
fix: multithreaded test on windows + hard fail on thread err

McPatate:
background thread was failing due to a reopen error on windows: it isn't possible to re-open a fail opened in readonly (namedtemporaryfile delete-on-close semantics).

two changes:
- change namedtemporaryfile to temporarydirectory + path join
- hard fail on thread execptions


danieldk (approved):
Interesting Windows behavior! Looks good!

Issue #729, opened 2026-04-12
Direct GPU loading via safe_open(device="cuda") 1.2-11x slower than CPU + pinned async path


davidpeden3:
**TL;DR**: `safe_open(device="cuda")` is 1.2-11x slower than the manual `safe_open(device="cpu")` + `pin_memory()` + async `.to("cuda")` path. Root cause is serialized kernel page faults during direct `cudaMemcpy` from mmap'd regions on cold loads. A transparent internal fallback to the faster path would fix it for all users and libraries.

## Summary

Directly loading safetensors to GPU with `safe_open(device="cuda:0")` is substantially slower than loading to CPU first, pinning, and using non-blocking transfer — up to 11x on large/oversubscribed models, and consistently 1.2-3.5x even for models that fit in VRAM.

## Root Cause

When `safe_open` is opened with `device="cuda:0"`, `get_tensor()` performs `cudaMemcpy` directly from the mmap'd file region. This triggers per-4KB page faults during the DMA copy — each fault serializes: page fault → kernel disk read → resume copy. Disk I/O and GPU transfer cannot overlap.

The CPU path benefits from efficient kernel readahead during the initial read + bulk pinned memcpy, while the direct path suffers serialized faults during the GPU DMA copy. `safe_open(device="cpu")` lets the kernel handle page faults efficiently, then `pin_memory()` copies from page cache to page-locked memory in one bulk operation, and `non_blocking .to("cuda:0")` uses fast async DMA.

Adding `madvise(MADV_SEQUENTIAL)` before loading further improves performance by enabling aggressive kernel readahead.

## Benchmark Results

**Hardware:** NVIDIA RTX 5090 (32GB VRAM), AMD Ryzen 9950X3D, PCIe 5.0 x16, NVMe SSD
**Software:** safetensors 0.7.0, PyTorch 2.9.1, CUDA 12.8, WSL2/Docker (Linux 6.6.87 kernel)
**Method:** Best of 3 runs, page cache dropped between runs via `posix_fadvise(POSIX_FADV_DONTNEED)`

| Model | Size | Files | `device=cuda` | cpu+pin+async | Speedup |
|-------|------|-------|---------------|---------------|---------|
| microsoft/phi-2 | 5.2 GB | 2 | 2.9s | 0.9s | **3.5x** |
| mistralai/Mistral-7B-v0.1 | 13.5 GB | 2 | 4.8s | 1.7s | **2.9x** |
[...]

PR #736, opened 2026-04-13
feat(python): opt-in CUDA fast path for safe_open (#729)

davidpeden3:
## Summary

Adds opt-in CUDA fast path for `safe_open`, gated by `SAFETENSORS_FAST_CUDA=1` env var. Torch-only; other frameworks untouched.

Addresses #729.

## Mechanism

The default `safe_open(device="cuda")` path copies from mmap regions directly into GPU memory. Each cudaMemcpy serializes on kernel page faults, preventing overlap between disk I/O and GPU transfer. The wrapper:

1. Opens the file via the existing Rust path with `device="cpu"`.
2. Applies `madvise(MADV_SEQUENTIAL)` for aggressive kernel readahead.
3. For each `get_tensor`, pins the CPU tensor and issues `.to(cuda, non_blocking=True)` on a dedicated CUDA stream.
4. Synchronizes the stream on context exit.

## Benchmarks

RTX 5090 / Ryzen 9950X3D, loading Gemma 4 26B-A4B (~48GB safetensors):

| Path                                  | Time | Throughput |
|---------------------------------------|------|------------|
| Current `safe_open(device="cuda")`    | 576s | 85 MB/s    |
| CPU + pin + async                     | 90s  | 550 MB/s   |
| CPU + pin + async + madvise           | 52s  | 956 MB/s   |

**Environment disclosure:** benchmarks run under WSL2 / Docker Desktop on Windows (WDDM driver). I don't have bare-metal Linux hardware; the mechanism (page-fault-serialized cudaMemcpy from mmap) is platform-agnostic so the pattern should reproduce on native Linux, but I cannot confirm.

## Opt-in rationale

- Peak memory briefly grows by the size of the largest tensor (pinned CPU + GPU simultaneously during transfer).
- Correct behavior depends on caller using `safe_open` as a context manager so the stream is synced before tensors are consumed.

These trade-offs argue for opt-in rather than default. Env var keeps the public API unchanged.

## Stopgap framing

Per discussion on #729, `mfuntowicz/hmll` is the long-term direction. This PR is intended as a stopgap that can be removed when hmll provides a CUDA path. The change is isolated to the Python binding layer; the Rust core is untouched.

## Tests

- Dispatcher gating (unit, no CUDA required): 4 tests
- Round-trip equality vs. Rust path (CUDA-gated, skipped when unavailable): 1 test
- Existing torch test suite: 55 pass unchanged

### DIFF
--- a/bindings/python/py_src/safetensors/__init__.py
+++ b/bindings/python/py_src/safetensors/__init__.py
@@ -3,8 +3,24 @@ from ._safetensors_rust import (  # noqa: F401
     SafetensorError,
     __version__,
     deserialize,
-    safe_open,
     _safe_open_handle,
     serialize,
     serialize_file,
 )
+from ._safetensors_rust import safe_open as _rust_safe_open
+
+
+def safe_open(filename, framework, device="cpu"):
+    """Open a safetensors file lazily.
+
+    Dispatches to the Rust implementation by default. When the environment
+    variable ``SAFETENSORS_FAST_CUDA=1`` is set and the caller requests a
+    torch tensor on a CUDA device, a Python wrapper routes reads through
+    CPU + pinned memory + async transfer on a dedicated stream. See
+    ``safetensors._fast_cuda`` for details.
+    """
+    from ._fast_cuda import FastCudaSafeOpen, fast_cuda_enabled
+
+    if fast_cuda_enabled(framework, device):
+        return FastCudaSafeOpen(filename, framework=framework, device=device)
+    return _rust_safe_open(filename, framework=framework, device=device)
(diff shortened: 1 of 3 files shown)


bindings/python/py_src/safetensors/__init__.py
```diff
@@ -3,8 +3,24 @@
...
+
+    Dispatches to the Rust implementation by default. When the environment
+    variable ``SAFETENSORS_FAST_CUDA=1`` is set and the caller requests a
+    torch tensor on a CUDA device, a Python wrapper routes reads through
+    CPU + pinned memory + async transfer on a dedicated stream. See
+    ``safetensors._fast_cuda`` for details.
+    """
+    from ._fast_cuda import FastCudaSafeOpen, fast_cuda_enabled
+
+    if fast_cuda_enabled(framework, device):
```
McPatate: perhaps worth adding a device is cuda check here as well

davidpeden3: you got it. done in b071bc2.

McPatate:
I think my latest PR #760 might've addressed this issue, could you confirm main fixes on your target platform/device?

PR #738, opened 2026-04-14
feat!: add `TensorSpec` param to `serialize*` fns

McPatate:
🧹 🧹 

Cleaner API with proper rust type

### DIFF
--- a/bindings/python/py_src/safetensors/__init__.pyi
+++ b/bindings/python/py_src/safetensors/__init__.pyi
@@ -5,10 +5,10 @@
 #   - `__version__: str`
 #   - type annotations on `TensorSpec` / `serialize` / `serialize_file`
 #
-# TODO: once we upgrade pyo3 to >= 0.28 (assuming `experimental-inspect`,
-# replace `stub.py` with a dedicated `tools/stub-gen` binary using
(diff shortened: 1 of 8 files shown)


bindings/python/py_src/safetensors/__init__.pyi
```diff
@@ -69,6 +90,66 @@ def serialize_file(tensor_dict, filename, metadata=None):
...
+    SAFETY: `data_ptr` is a raw memory address. The caller must ensure the
+    underlying buffer stays alive for the duration of every `serialize` /
+    `serialize_file` call that consumes this spec.
+    """
+    def __init__(
+        self,
+        dtype: str,
+        shape: Sequence[int],
+        data_ptr: int,
+        data_len: int,
```
danieldk, line 118: I would

```suggestion
        *,
        dtype: str,
        shape: Sequence[int],
        data_ptr: int,
        data_len: int,
```

and avoid accidental ptr/len mixups (since both are `int`) (+ on the Rust rside)

danieldk (commented):
Love it! Added one small comment that might avoid potential incorrect use (especially because pointers are the sensitive part).

PR #746, opened 2026-04-16
feat: upstream `convert.py` script

McPatate:
upstreaming the latest version of the `convert.py` script from the space: https://huggingface.co/spaces/safetensors/convert/


LysandreJik (approved):
yes nice!

PR #749, opened 2026-04-16
fix(python): use checked_mul for F4 last-dim doubling

willardjansen:
## Summary

`TensorSpec.__new__` doubles the last shape dimension for `float4_e2m1fn_x2` (F4) in order to convert the storage shape to the logical element count recorded in the safetensors header (`bindings/python/src/lib.rs:64-67`). The multiplication uses raw `*=` on `usize`, so a last dim greater than `usize::MAX / 2` panics in debug builds and wraps silently in release builds.

- **Debug builds:** `thread '<unnamed>' panicked at src/lib.rs:66:13: attempt to multiply with overflow`
- **Release builds:** the wrap succeeds silently and a `TensorSpec` is returned with a corrupted shape field. The serializer will then record the wrapped shape into the file header. The resulting file is self-consistent under `Metadata::validate` (because `validate` re-derives size with `checked_mul`), but it does not match the caller's intent — a `[2**63 + 1]` F4 tensor is silently serialized as a `[2]` tensor.

This is a correctness issue at the FFI boundary rather than a memory-safety problem (the `checked_mul` gates in the Rust core hold for every read path I looked at), but it violates the principle of least surprise for anyone building `TensorSpec` from user-derived shapes.

### Reproduction (release build, before fix)

```python
import ctypes
from safetensors import _safetensors_rust as _rs

buf = ctypes.create_string_buffer(1)
spec = _rs.TensorSpec(
    dtype="float4_e2m1fn_x2",
    shape=[2**63 + 1],
    data_ptr=ctypes.addressof(buf),
    data_len=1,
)
print(spec.shape)                 # [2] — silently corrupted
serialized = _rs.serialize({"test": spec})
# Header records: {"test":{"dtype":"F4","shape":[2],"data_offsets":[0,1]}}
```

## Change

Replace the raw multiplication with `checked_mul(2)` and return a clear `SafetensorError` on overflow.

```rust
shape[n - 1] = shape[n - 1].checked_mul(2).ok_or_else(|| {
    SafetensorError::new_err(format!(
        "F4 last-dim {} doubled to logical shape overflows usize",
        shape[n - 1]
    ))
})?;
```

## Test plan

- [x] New regression test `test_f4_shape_last_dim_overflow` in `bindings/python/tests/test_simple.py` covering the overflow case
- [x] Manual verification that valid small F4 shapes still round-trip correctly (`shape=[4]` → stored `[8]`)
- [x] `maturin develop --release` builds cleanly

Happy to adjust the error message or test style to match your conventions.

### DIFF
--- a/bindings/python/tests/test_simple.py
+++ b/bindings/python/tests/test_simple.py
@@ -257,6 +257,19 @@ class ReadmeTestCase(unittest.TestCase):
                 }
             )
 
+    def test_f4_shape_last_dim_overflow(self):
+        # The F4 branch doubles the last shape dim (storage -> logical). If the
+        # last dim > usize::MAX / 2, the multiplication must error rather than
+        # wrap silently (release) or panic (debug).
+        huge = 2**63 + 1
+        with self.assertRaises(SafetensorError):
+            TensorSpec(
+                dtype="float4_e2m1fn_x2",
+                shape=[huge],
+                data_ptr=0,
+                data_len=1,
+            )
+
     def test_torch_slice(self):
         A = torch.randn((10, 5))
         tensors = {
--- a/bindings/python/tests/test_simple.py
+++ b/bindings/python/tests/test_simple.py
@@ -257,19 +257,6 @@ class ReadmeTestCase(unittest.TestCase):
                 }
             )
 
-    def test_f4_shape_last_dim_overflow(self):
-        # The F4 branch doubles the last shape dim (storage -> logical). If the
-        # last dim > usize::MAX / 2, the multiplication must error rather than
-        # wrap silently (release) or panic (debug).
-        huge = 2**63 + 1
-        with self.assertRaises(SafetensorError):
-            TensorSpec(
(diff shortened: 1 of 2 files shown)


bindings/python/tests/test_simple.py
```diff
@@ -257,6 +257,19 @@ def test_exception(self):
...
+        # wrap silently (release) or panic (debug).
+        huge = 2**63 + 1
+        with self.assertRaises(SafetensorError):
+            TensorSpec(
+                dtype="float4_e2m1fn_x2",
+                shape=[huge],
+                data_ptr=0,
+                data_len=1,
+            )
+
```
McPatate: no need to test this!
```suggestion
```

McPatate (changes requested):
thanks for adding the check!

willardjansen:
Applied the suggested change — removed the test function per your feedback. Ready for re-review.

McPatate (approved):

PR #756, opened 2026-04-22
[MPS] Fast load of the tensors

Isalia20:
# What does this PR do?

Implements fast loading of safetensors onto MPS. New API from PyTorch allows us to grab the host-wrieable pointer via `torch.mps._host_alias_storage` and parallel `preads` straight to the MTLBuffers.

Falls back to the existing `safe_open` path when the torch API is missing or on non-macOS.

## Speedup (Tested on M5 Pro)

- 2 GB warm: 7 → 50 GB/s (~5×)
- 16 GB cold: 2.83 → 6.81 GB/s (2.4×)

### DIFF
--- a/.github/workflows/build_documentation.yml
+++ b/.github/workflows/build_documentation.yml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   build:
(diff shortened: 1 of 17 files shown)


McPatate (approved):
🔥

danieldk (approved):

PR #759, opened 2026-04-29
feat(python): add opt-in no-mmap loading

johnnynunez:
## Summary
- Adds a keyword-only `mmap` option to `safe_open`, defaulting to the existing mmap-backed behavior.
- Adds an opt-in `mmap=False` direct file-backed storage path that reads tensors on demand without memory-mapping the file.
- Threads `mmap` through `safetensors.torch.load_file` and `load_model`, and adds focused tests for the no-mmap path.
- https://github.com/safetensors/safetensors/issues/758

## Motivation
On unified CPU/GPU memory systems such as NVIDIA DGX Spark / Grace Blackwell and Jetson-class devices, mmap-backed file pages and framework/device tensors can reside in the same physical memory pool. For large model files, that can effectively double resident memory while loading weights. An opt-in no-mmap path gives downstream frameworks a way to avoid retaining mmap-backed file storage while preserving the current default behavior for discrete-memory systems.

Related downstream reports:
- https://github.com/Comfy-Org/ComfyUI/issues/10896
- https://forums.developer.nvidia.com/t/buyers-beware-dgx-spark-limited-to-64gb-in-comfyui/356573

## Test plan
- `cargo fmt --manifest-path bindings/python/Cargo.toml --check`
- `cargo check --manifest-path bindings/python/Cargo.toml`
- `uvx maturin build --manifest-path bindings/python/Cargo.toml --release -o bindings/python/dist`
- `PYTHONPATH=bindings/python/dist/wheel_extract python -m pytest tests/test_no_mmap.py tests/test_simple.py tests/test_pt_comparison.py -q`

PR #760, opened 2026-04-30
feat: add `backend` with pread file

### LINKED ISSUE #758: safe_open() always mmaps: doubles peak memory on unified-memory devices (Grace Blackwell / DGX Spark / Apple Silicon, etc.)
## Summary

`safetensors.safe_open(...)` (and therefore `safetensors.torch.load_file`) always memory-maps the file in the Rust core. There is currently no public API to opt out of `mmap` and instead stream the tensor bytes from disk into a freshly-allocated buffer.

On systems with **unified CPU/GPU memory** — NVIDIA Grace Blackwell / DGX Spark, Apple Silicon (`mps`), AMD APUs, NVIDIA Jetson / IGX, integrated iGPUs, etc. — this is much more than a stylistic choice: it makes loading large checkpoints **OOM well before the hardware limit is reached**.

## The problem

On a discrete-GPU box the page cache backing the `mmap` lives in host DRAM and the device copy lives in VRAM, so they are accounted to two different physical pools and total peak memory ≈ `1×model`.

On a unified-memory device there is only **one** physical pool. After:

1. `safe_open()` `mmap`s the `.safetensors` file → kernel page cache populates with model bytes (pool A: ~1×model).
2. `f.get_tensor(k)` / `.to(device=...)` allocates a device tensor and copies into it (pool B: ~1×model).

…both A and B are the same physical RAM. Peak memory is **~2×model**, and large models (FLUX.2-dev FP16, Wan, Hunyuan, big LLMs, …) reliably OOM on 128 GB DGX Spark / Grace Blackwell / large Apple Silicon machines that have *more* than enough memory to hold a single copy of the model plus the rest of the pipeline.

The page cache *should* be evictable, but in practice, while the copy is in flight (and especially with prefetch / readahead and refcounted mmap views held by the Python tensors), the kernel has no opportunity to drop those pages, so the working set really is ~2×model.

This was originally reported against ComfyUI ([Comfy-Org/ComfyUI#10896](https://github.com/Comfy-Org/ComfyUI/issues/10896)) but it is a property of `safetensors` itself, not of the consumer.

## Reproducer

```python
import safetensors.torch, torch, resource

# any reasonably large checkpoint, e.g. FLUX.2-dev FP16 (~32 GB)
sd = safetensors.torch.load_file(\"flux2-dev.safetensors\", device=\"cuda\")
print(\"peak RSS (GB):\", resource.getrusage(resource.RUSAGE_SELF).ru_maxrss / 1024 / 1024)
```

On a DGX Spark / Grace Blackwell box (unified 128 GB pool) the peak working set is roughly `2 × file_size` instead of `1 × file_size`.

`--disable-mmap` in ComfyUI does **not** fix this on its own, because at the safetensors level we still go through `safe_open()` which mmaps regardless.

## Existing workaround (downstream)

ComfyUI now ships a manual safetensors loader that bypasses `safe_open` entirely and reads each tensor straight from disk into a per-tensor `bytearray` via `readinto`, then zero-copy-wraps it as a torch tensor with `torch.frombuffer`:

- [`ComfyUI/comfy/utils.py` `load_safetensors_no_mmap`](https://github.com/comfyanonymous/ComfyUI/blob/master/comfy/utils.py)

Sketch:

```python
header_size = struct.unpack(\"<Q\", f.read(8))[0]
header = json.loads(f.read(header_size))
data_base_offset = 8 + header_size
for name, info in header.items():
    start, end = info[\"data_offsets\"]
    n = end - start
    buf = bytearray(n)
    f.seek(data_base_offset + start)
    view = memoryview(buf)
    off = 0
    while off < n:
        off += f.readinto(view[off:])
    t = torch.frombuffer(buf, dtype=DTYPES[info[\"dtype\"]]).reshape(info[\"shape\"])
    if device.type != \"cpu\":
        t = t.to(device=device)   # one tensor in flight, then 'buf' is freed
```

With this path, peak memory on DGX Spark is **one** model copy (plus the bytes of a single tensor in flight), and FLUX.2-dev FP16 + text encoder loads cleanly in ~90 GB on a 128 GB Spark — same as the GGUF path via llama.cpp's `--no-mmap`.

llama.cpp / llama-server / LM Studio report the exact same problem on the same class of hardware and solve it with `--no-mmap`. So this isn't ComfyUI-specific; it's the unified-memory + mmap interaction.

## Feature request

Please add a first-class way to load a `.safetensors` file **without `mmap`**, exposed through both the Rust core and the Python bindings. Some shape suggestions, in rough order of preference:

1. **`mmap=False` (or `use_mmap=False`) kwarg on `safe_open()` / `load_file()`**

   ```python
   with safetensors.safe_open(path, framework=\"pt\", device=\"cuda\", mmap=False) as f:
       ...
   safetensors.torch.load_file(path, device=\"cuda\", mmap=False)
   ```

   Internally this would `read`/`pread` the requested byte ranges into a fresh allocation per tensor (or one big allocation, your choice), bypassing the page cache. Equivalent of llama.cpp's `--no-mmap`.

2. **An environment variable**, e.g. `SAFETENSORS_NO_MMAP=1`, mirroring `HF_HUB_*` knobs, so consumers like Transformers, Diffusers, vLLM, ComfyUI, etc. can flip it globally for unified-memory deployments without each tool having to plumb a kwarg.

3. **Auto-detect unified memory** (lower priority — heuristics here are messy: `iGPU`, `mps`, `cuda` on Tegra/Grace, ROCm APUs, …). A manual switch is enough; auto-detect can come later.

Whichever shape you prefer, the key request is: **the Python user must be able to ask `safetensors` not to `mmap`**, and have it instead stream from the file descriptor.

## Why fix it upstream

Right now every framework on top of safetensors (Transformers, Diffusers, vLLM, ComfyUI, SD-WebUI, kohya, …) is going to need its own bespoke \"manual safetensors parser\" workaround to support DGX Spark / Grace Blackwell / Apple Silicon properly for large models. That's a lot of duplicated, slightly-different reimplementations of the on-disk format, and it defeats the point of having a single shared loader.

A single `mmap=False` knob in `safetensors` itself fixes it for the entire ecosystem.

## Environment

- `safetensors`: latest (verified up to 0.7.x)
- Hardware where this matters in practice: NVIDIA DGX Spark / GB200 / GB10 (Grace Blackwell, unified LPDDR5x), Apple Silicon, AMD APUs, NVIDIA Jetson / IGX, integrated iGPUs.
- Models that reliably trigger it on DGX Spark (128 GB unified): FLUX.2-dev FP16, Wan, Hunyuan-Video, large Qwen / Llama checkpoints, …

Happy to send a PR if a kwarg-style API (option 1) sounds acceptable.

McPatate:
Add a `Backend` parameter to `safe_open` to offer an efficient enough fallback methods for devices where our current mmap implementation causes OOMs (unified memory devices when using PyTorch native mmap) or is slow, e.g. loading a model from an NFS. We now have two backends:
- `Backend::Mmap` (`"mmap"`): similar as before, default
- `Backend::ReadFile` (`"read_file"`): keep an open file handle and `pread(2)` into temp host buffers dropped after device transfer.

Independently of the backend, for cuda + pytorch we now load bytes into an intermediate cpu pinned memory buffer before sending to device. This should fix the OOM reported in #728 and #759 while leaving the option of using mmap backends for users on unified memory devices. I suspect the issue comes from how PyTorch manages its memory when opening files with mmap, #728 fix points in that direction at least because using `Storage::Mmap` seems to fix.

Closes #758, supersedes #728, #759

### DIFF
--- a/bindings/python/py_src/safetensors/__init__.pyi
+++ b/bindings/python/py_src/safetensors/__init__.pyi
@@ -165,8 +165,12 @@ class safe_open:
 
         device (`str`, defaults to `"cpu"`):
             The device on which you want the tensors.
+
+        backend (`str`, *keyword-only*, defaults to `"mmap"`):
+            Storage backend used to serve tensor bytes. `"mmap"` (default) and
+            `"read_file"` uses `pread(2)` to read tensor bytes.
     """
-    def __init__(self, filename, framework, device=...):
+    def __init__(self, filename, framework, device=..., *, backend: str = "mmap"):
         pass
 
     def __enter__(self):
--- a/bindings/python/py_src/safetensors/__init__.pyi
+++ b/bindings/python/py_src/safetensors/__init__.pyi
@@ -168,7 +168,7 @@ class safe_open:
 
         backend (`str`, *keyword-only*, defaults to `"mmap"`):
             Storage backend used to serve tensor bytes. `"mmap"` (default) and
-            `"read_file"` uses `pread(2)` to read tensor bytes.
+            `"pread"` uses `pread(2)` to read tensor bytes.
     """
     def __init__(self, filename, framework, device=..., *, backend: str = "mmap"):
         pass
--- a/bindings/python/src/lib.rs
+++ b/bindings/python/src/lib.rs
@@ -334,16 +334,17 @@ fn slice_to_indexer(
 enum Backend {
     Mmap,
     /// Keeps the file handle open and serves each `get_tensor` /
-    /// `get_slice` via `pread(2)`, dispatching on `(framework, device)` to
-    /// write directly into a destination buffer chosen for performance.
-    ReadFile,
+    /// `get_slice` via `pread(2)` (or its Windows equivalent),
+    /// dispatching on `(framework, device)` to write directly into a
+    /// destination buffer chosen for performance.
+    Pread,
 }
 
 impl fmt::Display for Backend {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.write_str(match *self {
             Backend::Mmap => "mmap",
-            Backend::ReadFile => "read_file",
+            Backend::Pread => "pread",
         })
     }
 }
@@ -355,9 +356,9 @@ impl<'a, 'py> FromPyObject<'a, 'py> for Backend {
         let name: String = ob.extract()?;
         match &name[..] {
             "mmap" => Ok(Backend::Mmap),
-            "read_file" => Ok(Backend::ReadFile),
+            "pread" => Ok(Backend::Pread),
             name => Err(SafetensorError::new_err(format!(
-                "backend {name:?} is invalid (expected one of: \"mmap\", \"read_file\")"
+                "backend {name:?} is invalid (expected one of: \"mmap\", \"pread\")"
             ))),
         }
     }
@@ -537,7 +538,7 @@ enum Storage {
     /// Holds an open file handle and
     /// serves each tensor via `pread(2)` into a fresh per-tensor host
     /// buffer, with framework/device-specific buffer choices for performance.
-    ReadFile(Arc<File>),
+    Pread(Arc<File>),
 }
 
 #[derive(Debug, PartialEq, Eq, PartialOrd)]
@@ -643,14 +644,14 @@ impl Open {
             Ok(())
         })?;
 
-        if backend == Backend::ReadFile {
+        if backend == Backend::Pread {
             return Ok(Self {
                 filename,
                 metadata,
                 offset,
                 framework,
                 device,
-                storage: Arc::new(Storage::ReadFile(Arc::new(file))),
+                storage: Arc::new(Storage::Pread(Arc::new(file))),
             });
         }
 
@@ -1008,11 +1009,11 @@ impl Open {
                     Ok(tensor.into_pyobject(py)?.into())
                 })
             }
-            Storage::ReadFile(file) => self.get_tensor_read_file(name, info, file),
+            Storage::Pread(file) => self.get_tensor_pread(name, info, file),
         }
     }
 
-    fn get_tensor_read_file(
+    fn get_tensor_pread(
         &self,
         name: &str,
         info: &TensorInfo,
@@ -1056,7 +1057,7 @@ impl Open {
 
         Python::attach(|py| -> PyResult<Py<PyAny>> {
             let torch = get_module(py, &TORCH_MODULE)?;
-            let dest = prepare_torch_pinned_cpu_dest(py, torch, info.dtype, &info.shape, nbytes)?;
+            let dest = PinnedCpuDest::new(py, torch, info.dtype, &info.shape, nbytes)?;
 
             if nbytes > 0 {
                 let write_ptr = dest.write_ptr;
@@ -1073,7 +1074,7 @@ impl Open {
                             dst.copy_from_slice(src);
                         });
                     }
-                    Storage::ReadFile(file) => {
+                    Storage::Pread(file) => {
                         let file_offset = (self.offset + begin) as u64;
                         let read_result: std::io::Result<()> = py.detach(|| {
                             // SAFETY: write_ptr/nbytes name `dest.tensor`'s
@@ -1180,7 +1181,7 @@ impl Open {
                 let (begin, end) = info.data_offsets;
                 let nbytes = end - begin;
 
-                let dest = prepare_mps_dest(py, torch, &name, info.dtype, &info.shape, nbytes)?;
+                let dest = MpsDest::new(py, torch, &name, info.dtype, &info.shape, nbytes)?;
                 if let Some(alias) = dest.host_alias {
                     host_aliases.push(alias);
                     jobs.push(PreadJob {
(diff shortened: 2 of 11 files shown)


McPatate:
I'm on a DGX Spark and can confirm main OOMs the device with a file size `>total_mem/2`.

With this branch, regardless of the chosen backend, memory pressure is maybe a little extra than the actual footprint. Surprised that the `read_file` backend is faster, especially given warm cache + we're not issuing `pread` calls in parallel.

Measuring the `load_file` call:
```
(safetensors) luc@spark:~/safetensors$ python3 oom_repro.py /tmp --backend mmap
loading 1 shards (69.88 GB total) on cuda:0 via backend=mmap

  [ 1/1] synthetic_70gb.safetensors               69.88GB  64.57s  host_rss= 57.81GB  cuda_alloc= 69.89GB  cuda_peak= 69.89GB

loaded 1695 tensors in 64.6s (1.08 GB/s)
peak host RSS:       57.81 GB
peak cuda alloc:     69.89 GB
peak cuda reserved:  70.62 GB
(safetensors) luc@spark:~/safetensors$ python3 oom_repro.py /tmp --backend read_file
loading 1 shards (69.88 GB total) on cuda:0 via backend=read_file

  [ 1/1] synthetic_70gb.safetensors               69.88GB  40.62s  host_rss=  2.75GB  cuda_alloc= 69.89GB  cuda_peak= 69.89GB

loaded 1695 tensors in 40.6s (1.72 GB/s)
peak host RSS:        2.75 GB
peak cuda alloc:     69.89 GB
peak cuda reserved:  70.62 GB
```

bindings/python/py_src/safetensors/__init__.pyi
```diff
@@ -165,8 +165,12 @@ class safe_open:
 
         device (`str`, defaults to `"cpu"`):
             The device on which you want the tensors.
+
+        backend (`str`, *keyword-only*, defaults to `"mmap"`):
+            Storage backend used to serve tensor bytes. `"mmap"` (default) and
+            `"read_file"` uses `pread(2)` to read tensor bytes.
```
danieldk: Not super-opinionated on this, but maybe it's `read_file` -> `pread`?

bindings/python/src/lib.rs
```diff
@@ -329,6 +329,40 @@ fn slice_to_indexer(
...
 }
 
+/// Storage backend used to serve tensor bytes.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum Backend {
+    Mmap,
+    /// Keeps the file handle open and serves each `get_tensor` /
+    /// `get_slice` via `pread(2)`, dispatching on `(framework, device)` to
+    /// write directly into a destination buffer chosen for performance.
+    ReadFile,
```
danieldk: Maybe `Pread`?

bindings/python/src/lib.rs
```diff
@@ -1622,6 +1732,209 @@ fn get_module<'a>(
...
+    {
+        use std::os::unix::fs::FileExt;
+        file.read_exact_at(buf, offset)
+    }
+    #[cfg(windows)]
+    {
+        use std::os::windows::fs::FileExt;
+        let mut written = 0;
+        while written < buf.len() {
+            let n = file.seek_read(&mut buf[written..], offset + written as u64)?;
```
danieldk, line 1765: This doesn't seem safe under multi-threading, since the file position is a global state. I am not really sure how this work with Python free-threaded interpreters. Can Rust code only be used single-threaded (is there still a lock?).

bindings/python/src/lib.rs
```diff
@@ -1622,6 +1732,209 @@ fn get_module<'a>(
...
+
+/// Pre-allocated MPS destination ready for a `pread` write. `write_ptr` is the
+/// host-aliased CPU pointer to the MPS tensor's MTLBuffer; `host_alias` keeps
+/// that alias storage alive across the writes — drop only after the trailing
+/// `mps.synchronize()`.
+#[cfg(target_os = "macos")]
+struct MpsDest<'py> {
+    tensor: PyBound<'py, PyAny>,
+    host_alias: Option<PyBound<'py, PyAny>>,
+    write_ptr: usize,
```
danieldk, line 1870: So `write_ptr` is always valid because `host_alias` is held?

> drop only after the trailing`mps.synchronize()`

It feels like this could be a bit error-prone? Maybe `MpsDest` should be bundled in a container (`MpsDests`?) that synchronizes in a `Drop` implementation?

bindings/python/src/lib.rs
```diff
@@ -1622,6 +1732,209 @@ fn get_module<'a>(
...
+}
+
+/// Allocate `torch.empty(...,device="mps")` and acquire a CPU-writable host
+/// alias to its storage.
+#[cfg(target_os = "macos")]
+fn prepare_mps_dest<'py>(
```
danieldk: Making this a `new` method on `MpsDest` seems cleaner?

bindings/python/src/lib.rs
```diff
@@ -1622,6 +1732,209 @@ fn get_module<'a>(
...
+
+/// Pre-allocated pinned-CPU torch destination.
+struct PinnedCpuDest<'py> {
+    tensor: PyBound<'py, PyAny>,
+    write_ptr: usize,
+}
+
+/// Allocate `torch.empty(...,device="cpu", pin_memory=True)` and extract the
+/// data pointer for direct pread.
+fn prepare_torch_pinned_cpu_dest<'py>(
```
danieldk: Could also be `new`.

bindings/python/src/lib.rs
McPatate: It does indeed move cursor position but as we don't use it to read (it's set as the second parameter) this is perfectly ok!

McPatate: got claude to detail the mechanism a bit more in the comment :)

danieldk (approved):

PR #764, opened 2026-05-11
fix: write to tempfile + rename to preserve mmap source

### LINKED ISSUE #762: [0.8.0rc0] save_file overwriting same path returns zeros when tensors come from prior load_file
## TL;DR

In `safetensors==0.8.0-rc.0`, calling `save_file(tensors, path)` where `tensors`
were just returned by `load_file(path)` produces a **zero-filled output file** —
both on disk and in the in-memory tensors. This appears to be a regression
vs `0.7.0` (last stable), which handles this pattern correctly.

The trigger is the path collision; the bug fires whether or not the dict is
modified between load and save.

## Repro

```python
import os, tempfile, torch
from safetensors.torch import load_file, save_file
import safetensors

print("safetensors:", safetensors.__version__)  # 0.8.0-rc.0
print("torch:", torch.__version__)              # 2.10.0

with tempfile.TemporaryDirectory() as tmp:
    path = os.path.join(tmp, "model.safetensors")

    # Write a file with non-zero data
    save_file({"w": torch.tensor([1.0, 2.0, 3.0])}, path)

    # Load it back, then save (no modification) to the SAME path
    loaded = load_file(path)
    save_file(loaded, path)

    # Re-read to inspect
    reloaded = load_file(path)
    print("on-disk after round-trip:", reloaded["w"])
    print("in-memory after save:    ", loaded["w"])
```

**Output on `0.8.0rc0`**:
```
safetensors: 0.8.0-rc.0
torch: 2.10.0
on-disk after round-trip: tensor([0., 0., 0.])
in-memory after save:     tensor([0., 0., 0.])
```

**Expected (and actual on `0.7.0`)**:
```
on-disk after round-trip: tensor([1., 2., 3.])
in-memory after save:     tensor([1., 2., 3.])
```

## Variant matrix

I tested seven variations to characterize the trigger:

| # | Pattern | 0.8.0rc0 result |
|---|---|---|
| 1 | load → rename keys → save same path | **fails (zeros)** |
| 2 | load → rename keys → save **different** path | passes |
| 3 | load → rename + `.clone()` → save same path | passes |
| 4 | load → save same path, **no modification** | **fails (zeros)** |
| 5 | load → in-place `tensor.add_(100)` → save same path | **fails (zeros)** |
| 6 | load → rename → `del` + `gc.collect()` → save same path | **fails (zeros)** |
| 7 | load → rename + `.detach().clone()` → save same path | passes |

So:
- The bug is **independent of whether the dict was modified** (variants 1, 4, 5 all fail).
- It is triggered **only when the save target is the same path** as the prior load (variant 2 passes).
- It is avoided by **materializing the tensors via `.clone()`** before save (variants 3, 7 pass).
- A `del` + `gc.collect()` is **not sufficient** to release the mmap before save (variant 6 fails).

## Likely cause (speculation)

`load_file` returns memory-mapped tensors backed by the input file. When
`save_file` is then called on the same path, the writer evidently truncates the
file before serializing, which zeroes the underlying mmap pages. The serializer
then reads from the (now-zero) tensors and writes zeros back to disk. The
in-memory tensors are also zero because they're mmap views of the same file.

If this analysis is correct, the fix is one of:
- Detect the path collision and either copy first, or refuse with a clear error
- Always read each input tensor into a CPU buffer before truncating the output

## Workarounds (for downstream users hitting this now)

1. **Save to a different path** (and rename atomically afterwards if needed).
2. **Clone every tensor** before passing to `save_file`:
   ```python
   save_file({k: v.clone() for k, v in loaded.items()}, path)
   ```

## Environment

- safetensors 0.8.0-rc.0 (installed via `pip install --pre safetensors==0.8.0rc0`)
- Python 3.13
- torch 2.10.0+rocm7.1 (also reproducible with CPU-only torch — the bug is
  not GPU-related)
- Linux 6.8.0 / Ubuntu 22.04
- Hardware: AMD MI300X (irrelevant — bug is in CPU file I/O)

Confirmed **does not** reproduce on `safetensors==0.7.0` with the same Python /
torch versions.

## Question

Is a fix already in `main` past `0.8.0rc0`? If so, happy to retest against the
current dev build. Otherwise, would a same-path-detection + auto-copy in
`save_file` be the preferred fix, or a clear error directing users to
`.clone()` / save-to-different-path?

McPatate:
Calling `save_file(loaded, path)` on tensors that were previously loaded with `loaded = load_file(path)` on Linux would produce a zero-filled file.

When writing, we open files with `O_TRUNC` using `File::create` and then `set_len(file_size)`, zeroing the file for `file_size` bytes which in the read and write to same-file scenario invalidates the active mmap pages and re-faults them as zeroes, causing a silent error & regression.

We now write to a temporary file and rename said file to `path` once write finishes, preserving the old data in memory as long as a reference exists while also outputting a valid file.

Closes #762


safetensors/src/tensor.rs
```diff
@@ -306,33 +310,99 @@ fn buffered_write_to_file<V: View>(
...
+                Ok(file) => return Ok(Self { path, file }),
+                Err(e) if e.kind() == std::io::ErrorKind::AlreadyExists => continue,
+                Err(e) => return Err(SafeTensorError::IoError(e)),
+            }
+        }
+        Err(SafeTensorError::IoError(std::io::Error::new(
+            std::io::ErrorKind::AlreadyExists,
+            "could not allocate a unique temporary file next to destination",
+        )))
+    }
```
danieldk: I am not a big fan of hand-rolling a custom temp file implementation. Here it's easily guessable, vulnerable to collision attacks (e.g. exhausting all possibilities).

I think it's better to use the `tempfile` create, which is very widely used and optimized for security. Also has a `persist` method that does an atomic rename, plus can also create a temporary file in the same directory, so I think it does everything we need it to do.

McPatate: Yeah I hesitated quite a bit, since we only need rename I was thinking this'd be fine. I didn't want to add too many dependencies to Safetensors which is already pretty light in that regard, and felt the security concerns were minimal.

But totally ok to add temp file, will update!

danieldk (approved):
Looks good!

Seems that Windows CI is still unhappy?

safetensors/src/tensor.rs
```diff
@@ -306,29 +306,37 @@ fn buffered_write_to_file<V: View>(
...
 ) -> Result<(), SafeTensorError> {
-    let file = std::fs::File::create(path)?;
+    let path = path.as_ref();
+    // Write to a sibling tempfile then rename, so an existing `path` is never
+    // truncated under any mmap of it (e.g. tensors returned by `load_file`).
+    let parent = path.parent().unwrap_or_else(|| Path::new("."));
+    let temp = tempfile::NamedTempFile::new_in(parent)?;
 
-    file.set_len(total_size as u64)?;
+    temp.as_file().set_len(total_size as u64)?;
```
nicolasdalsass, line 315: Hello :-) 

FYI, I noticed this change because it doesn't have the same file rights semantic as the previous version. Previously, the file at path would be created with the equivalent of a chmod 666 (with `std::fs::File::create`), whereas the temp file is created with the equivalent of a chmod 600 (with `tempfile::NamedTempFile::new_in`).

So in the end, the files created throught this path are more restrictive than before, which for me caused the model to not be usable by multiple users sharing a group.

While it's not something you may want to change, I thought you may want to be aware of that.

Cheers !

McPatate: Oh that's a regression that was not intended. Will try to make sure we create the file as 666!

Could you open an issue so I can track?

nicolasdalsass: I created https://github.com/safetensors/safetensors/issues/782 :-)

PR #765, opened 2026-05-12
Return ValidationOverflow from TensorView size overflow

GHX5T-SOL:
## Summary
- Use checked arithmetic in `TensorView::new` for the shape product and dtype bit-size multiplication.
- Return the existing `ValidationOverflow` error instead of panicking in debug builds or relying on wrapping arithmetic in release builds.
- Add regression coverage for both shape-product and dtype-size overflow cases.

## Validation
- `cargo fmt --check --manifest-path safetensors/Cargo.toml`
- `cargo test --manifest-path safetensors/Cargo.toml`
- `git diff --check -- safetensors/src/tensor.rs`
- `git diff -- safetensors/src/tensor.rs | gitleaks stdin --no-banner --redact --exit-code 1`

### DIFF
--- a/safetensors/src/tensor.rs
+++ b/safetensors/src/tensor.rs
@@ -752,9 +752,15 @@ impl<'data> TensorView<'data> {
         shape: Vec<usize>,
         data: &'data [u8],
     ) -> Result<Self, SafeTensorError> {
-        let n_elements: usize = shape.iter().product();
+        let n_elements: usize = shape
+            .iter()
+            .copied()
+            .try_fold(1usize, usize::checked_mul)
+            .ok_or(SafeTensorError::ValidationOverflow)?;
 
-        let nbits = n_elements * dtype.bitsize();
+        let nbits = n_elements
+            .checked_mul(dtype.bitsize())
+            .ok_or(SafeTensorError::ValidationOverflow)?;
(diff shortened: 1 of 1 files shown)


McPatate:
Hey, thank you for the PR. Do you have a real-world example where a check of this nature is needed? This feels overly defensive as I don't see any code paths that would hit this, we validate the sizes in our metadata validation fn.

GHX5T-SOL:
Thanks, that is a fair concern. I do not have a malformed `.safetensors` file path that bypasses `Metadata::validate`; deserialization is already covered there.

The reachable case I was addressing is direct Rust API use of `TensorView::new(...)`, where a caller constructs a view from external shape/dtype/data for serialization or slicing. On current main, shapes such as `vec![usize::MAX, 2]` can panic in debug during the unchecked product/multiplication, and wrap in release before the length comparison. This PR makes that constructor return the existing `ValidationOverflow`, matching the checked arithmetic already used by `Metadata::validate`.

I rechecked the focused regression locally:

```text
cargo test --manifest-path safetensors/Cargo.toml tensor::tests::test_tensor_view_validation_overflow -- --exact
# 1 passed
```

If you prefer not to carry this public-constructor hardening without an in-repo caller, I am fine closing it.

PR #766, opened 2026-05-12
Avoid stride overflow for empty tensor slice requests

GHX5T-SOL:
## Summary

`SliceIterator::new` used to walk the tensor shape even when no slice indexers were provided. For zero-sized tensors with a very large dimension, that no-op slice path could overflow in debug builds while accumulating row-major stride state before falling back to the full tensor buffer.

This returns the full-buffer iterator immediately when `slices` is empty. It preserves the existing no-op slice behavior and avoids doing stride arithmetic that is not needed for that case.

## Validation

- `cargo fmt --check --manifest-path safetensors/Cargo.toml`
- `cargo test --manifest-path safetensors/Cargo.toml slice::tests::test_zero_sized_tensor_with_large_dim_does_not_panic -- --exact`
- `cargo test --manifest-path safetensors/Cargo.toml`
- `cargo test --release --manifest-path safetensors/Cargo.toml slice::tests::test_zero_sized_tensor_with_large_dim_does_not_panic -- --exact`
- `cargo test --manifest-path safetensors/Cargo.toml --all-features`
- `git diff --check -- safetensors/src/slice.rs`

No linked issue; this came out of a local slice-path audit.

### DIFF
--- a/safetensors/src/slice.rs
+++ b/safetensors/src/slice.rs
@@ -283,6 +283,13 @@ impl<'data> SliceIterator<'data> {
         if n_slice > n_shape {
             return Err(InvalidSlice::TooManySlices);
         }
+        if slices.is_empty() {
+            return Ok(Self {
+                view,
+                indices: vec![(0, view.data().len())],
+                newshape: view.shape().to_vec(),
+            });
+        }
         let mut newshape = Vec::with_capacity(view.shape().len());
 
         // Minimum span is the span of 1 item;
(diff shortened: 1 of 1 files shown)


McPatate:
Same as for #765 , any real world use case where this is an issue?

GHX5T-SOL:
Same boundary as #765: I do not have a file-deserialization bypass here. This is a direct public API edge case around zero-sized tensor views.

The concrete case is `TensorView::new(Dtype::U8, vec![0, usize::MAX], &[])`, which is a valid zero-byte tensor view. Asking for the whole tensor with an empty slice list should not need any stride multiplication, but current main still walks the shape and reaches unchecked `span *= shape`, which can panic in debug even though no slicing is requested. The patch short-circuits empty slice lists to return the full byte range.

Focused regression still passes locally:

```text
cargo test --manifest-path safetensors/Cargo.toml slice::tests::test_zero_sized_tensor_with_large_dim_does_not_panic -- --exact
# 1 passed
```

If this is too defensive for the API surface, I am fine closing it.

PR #767, opened 2026-05-13
refactor: replace `_host_alias_storage` w/ `MTLBuffer`

McPatate:
After applying the following diff to transformers (starting from https://github.com/huggingface/transformers/pull/46239):
```diff
diff --git a/src/transformers/core_model_loading.py b/src/transformers/core_model_loading.py
index 271e1898d4..b6d47492e3 100644
--- a/src/transformers/core_model_loading.py
+++ b/src/transformers/core_model_loading.py
@@ -993,7 +993,11 @@ GLOBAL_WORKERS = min(4, os.cpu_count() or 4)

 def _materialize_copy(tensor: torch.Tensor, device=None, dtype=None) -> torch.Tensor:
     # This slicing is what actually loads the tensor from the safetensors slice object
-    tensor = tensor[...]
+    # FIXME: update this when ellipsis support for MPS/pread lands
+    try:
+        tensor = tensor[...]
+    except TypeError:
+        tensor = tensor[:]
     if dtype is not None or device is not None:
         tensor = tensor.to(device=device, dtype=dtype)
     return tensor
diff --git a/src/transformers/modeling_utils.py b/src/transformers/modeling_utils.py
index db241909eb..802e6c2a81 100644
--- a/src/transformers/modeling_utils.py
+++ b/src/transformers/modeling_utils.py
@@ -4497,7 +4497,12 @@ class PreTrainedModel(nn.Module, EmbeddingAccessMixin, ModuleUtilsMixin, PushToH
                         with open(file, "rb") as _fh:
                             merged_state_dict.update(_safe_load_bytes(_fh.read()))
                         continue
-                    file_pointer = safe_open(file, framework="pt", device="cpu")
+                    is_mps = load_config.device_map is not None and any(
+                        (d.type if isinstance(d, torch.device) else d) == "mps"
+                        for d in load_config.device_map.values()
+                    )
+                    backend, device = ("pread", "mps") if is_mps else ("mmap", "cpu")
+                    file_pointer = safe_open(file, framework="pt", device=device, backend=backend)
                     all_pointer.add(file_pointer)
                     for k in file_pointer.keys():
                         merged_state_dict[k] = file_pointer.get_slice(k)  # don't materialize yet
```

### Benchmarks on M3 Pro 36gb ram

Note that I'm running benchmarks on my computer with everything open, so emulating a user browsing the web, messages, music and whatnot while having his local model running in the background.
In case anyone is wondering why swap / rss is still significant.

#### pread
```
  safetensors 0.8.0-dev.0   transformers 5.10.0.dev0
  model: Qwen/Qwen2.5-14B   dtype: bfloat16   device_map: mps
  at start: swap 2.75G  wired 3.33G  compressed 0.34G

  W0527 22:00:02.496000 32515 torch/distributed/elastic/multiprocessing/redirects.py:35] NOTE: Redirects are currently not supported in MacOs.
  Loading weights: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 579/579 [00:09<00:00,
  61.15it/s]
    from_pretrained   13.28s   params 14.8B
    peak  rss  10.9G  phys  27.9G  wired   3.2G  comp  23.7G  swap  12.0G
    MPS driver  27.5G  current   0.0G
```
#### mmap
```
  safetensors 0.8.0-dev.0   transformers 5.10.0.dev0
  model: Qwen/Qwen2.5-14B   dtype: bfloat16   device_map: mps
  at start: swap 2.65G  wired 3.12G  compressed 0.63G

  W0527 22:07:57.854000 39328 torch/distributed/elastic/multiprocessing/redirects.py:35] NOTE: Redirects are currently not supported in MacOs.
  Loading weights: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 579/579 [00:50<00:00,
  11.43it/s]
    from_pretrained   60.01s   params 14.8B
    peak  rss  20.0G  phys  27.9G  wired   3.1G  comp  19.4G  swap  12.3G
    MPS driver  27.5G  current   0.0G

```

#### pre-changes

OOMs 2 out of 3 times, swap goes to 34+Gb, physical memory is >29gb and timing were in the 75+ secs range.

The cause of this memory footprint + oom is coming from the fact that in transformers we `safe_open(device="cpu").to(mps)`.
Opening the file with the `UntypedStorage.from_file()` call loads the full file in memory, followed by the `.to(mps)` call effectively doubles the memory usage since we're on unified memory (apple silicon). Pytorch's mps cache allocator retains the freed buffers instead of releasing the memory.
The CPU storage memory is held until the `safe_open` context is closed and freed MPS buffers aren't reclaimed until `torch.mps.empty_cache()` is called.


bindings/python/src/dlpack.rs
```diff
@@ -0,0 +1,268 @@
...
+/// and not deleted before.
+unsafe extern "C" fn managed_tensor_deleter<B: AsDevicePtr>(self_ptr: *mut DLManagedTensor) {
+    if self_ptr.is_null() {
+        return;
+    }
+    let ctx_ptr = unsafe { (*self_ptr).manager_ctx as *mut ManagedCtx<B> };
+    if !ctx_ptr.is_null() {
+        unsafe { drop(Box::from_raw(ctx_ptr)) };
+    }
+    unsafe { drop(Box::from_raw(self_ptr)) };
```
danieldk: Some questions:

* Why wrapped in a `Box` and then dropped, instead of a raw free?
* Is this safe? What if the pointers were allocated with a different allocator?

bindings/python/src/dlpack.rs
```diff
@@ -0,0 +1,268 @@
...
+        device_type: DLDeviceType::Cuda,
+        device_id: ordinal,
+    }
+}
+
+#[cfg(target_os = "macos")]
+pub(crate) fn metal_device() -> DLDevice {
+    DLDevice {
+        device_type: DLDeviceType::Metal,
+        device_id: 0,
```
danieldk, line 212: Is this always the case? E.g. Mac Pro x86_64 with multiple MPX modules. I am not sure if this Metal support works at all, since the assumptions are very different (e.g. Metal buffers in main memory for Apple Silicon vs. Metal buffers in VRAM for x86_64 with discrete GPUs). I think we still support Intel Macs, if so, maybe we need to make some paths conditional on ARM64, or even better check if the memory is CPU memory and not (discrete) GPU memory?

bindings/python/src/dlpack.rs
```diff
@@ -0,0 +1,268 @@
...
+
+    let ctx = Box::new(ManagedCtx {
+        _device_buf: device_buf,
+        shape: shape.into_boxed_slice(),
+    });
+
+    let shape_ptr = ctx.shape.as_ptr() as *mut i64;
+    let ctx_ptr = Box::into_raw(ctx);
+
+    let managed = Box::into_raw(Box::new(DLManagedTensor {
```
danieldk, line 240: Maybe worth adding a comment here that this is safe to pass to C-land because the deallocation of not only the `DLManagedTensor` is handled by the `managed_tensor_deleter`, but `DLManagedTensor` itself as well.

bindings/python/src/lib.rs
```diff
@@ -292,41 +294,86 @@ fn deserialize(py: Python, bytes: &[u8]) -> PyResult<Vec<(String, HashMap<String
...
-) -> Result<TensorIndexer, PyErr> {
-    match slice_index {
-        SliceIndex::Slice(slice) => {
-            let py_start = slice.getattr(intern!(slice.py(), "start"))?;
-            let start: Option<usize> = py_start.extract()?;
-            let start = if let Some(start) = start {
-                Bound::Included(start)
-            } else {
-                Bound::Unbounded
+/// Convert a Python slicing argument into the crate's `TensorIndexer` form.
```
danieldk: Not 100% sure how this change is related to the rest?

bindings/python/src/lib.rs
```diff
@@ -1496,6 +1586,112 @@ impl PySafeSlice {
...
         })
     }
+
+    /// MPS slice path: alloc a Shared `MTLBuffer` sized for the post-slice
+    /// tensor, fill each source segment directly into it via pread/memcpy,
+    /// then hand off via DLPack. Skips the bytearray + CPU tensor +
+    /// `.to("mps")` round-trip the general path takes.
+    #[cfg(target_os = "macos")]
+    fn slice_mps(&self, slices: &PyBound<'_, PyAny>) -> PyResult<Py<PyAny>> {
+        let indexers = parse_indexers(slices, &self.info.shape)?;
```
danieldk, line 1580: Ah, this is what required the `parse_indexers` rewrite?

bindings/python/src/metal.rs
```diff
@@ -0,0 +1,72 @@
...
+fn device() -> Result<&'static ProtocolObject<dyn MTLDevice>, &'static str> {
+    if let Some(h) = DEVICE.get() {
+        return Ok(&h.0);
+    }
+    let dev = MTLCreateSystemDefaultDevice().ok_or("MTLCreateSystemDefaultDevice returned nil")?;
+    let _ = DEVICE.set(DeviceHandle(dev));
+    Ok(&DEVICE.get().unwrap().0)
+}
+
+pub struct MtlBuffer {
```
danieldk: Nit, maybe mirror the Objective-C/Swift side and call it `MTLBuffer`? Requires scoping of `objc2_metal::MTLBuffer`, but I think that's good, makes it clearer when this file refers to the raw flavor and the Rust-managed flavor.

bindings/python/src/lib.rs
McPatate: It's not, but I needed to test the code out (I don't remember where though). Can split into a separate PR.

McPatate: It's not, but I needed to test the code out (I don't remember where though). Can split into a separate PR.

bindings/python/src/lib.rs
McPatate: yes!

danieldk (approved):
Looks really nice! Awesome improvement!

bindings/python/src/lib.rs
```diff
@@ -1847,76 +2027,88 @@ fn torch_storage_shape(dtype: Dtype, logical_shape: &[usize]) -> PyResult<Vec<us
...
+///
+/// - **MLX**: no `from_dlpack` exists; `mx.array(capsule)` fails. The
+///   public Metal-zero-copy path tracked at ml-explore/mlx#2855 isn't
+///   shipped yet (needs allocator::Buffer wrapping + Deleter plumbing).
+///   The numpy-buffer-protocol path is *not* end-to-end zero-copy:
+///   MLX copies into its own MTLBuffer on first GPU dispatch.
+/// - **Numpy / Jax / Paddle / Flax**: not relevant for MPS; none have a
+///   functioning MPS device path.
+#[cfg(all(target_os = "macos", target_arch = "aarch64"))]
+fn ingest_dlpack_mps(
```
EricLBuehler, line 2132: This relies on `torch.from_dlpack` importing an `MTLBuffer` that was allocated by our own `MTLCreateSystemDefaultDevice`, not by torch's MPS caching allocator. On older torch without support for importing `MTLBuffer` from dlpack this will raise an error, but maybe we should keep a fallback for that case?

bindings/python/src/metal.rs
```diff
@@ -0,0 +1,91 @@
...
+
+static DEVICE: OnceLock<DeviceHandle> = OnceLock::new();
+
+struct DeviceHandle(Retained<ProtocolObject<dyn MTLDevice>>);
+
+// MTLDevice is documented thread-safe.
+unsafe impl Send for DeviceHandle {}
+unsafe impl Sync for DeviceHandle {}
+
+fn device() -> Result<&'static ProtocolObject<dyn MTLDevice>, &'static str> {
```
EricLBuehler: (nit): In theory a get -> set -> get sequence can construct (and then drop) a second MTLDevice if two threads race here. I think something like `OnceLock::get_or_init` should solve this.

bindings/python/src/dlpack.rs
```diff
@@ -0,0 +1,273 @@
...
+    pub deleter: Option<unsafe extern "C" fn(*mut DLManagedTensor)>,
+}
+
+const CAPSULE_NAME: &[u8] = b"dltensor\0";
+
+pub(crate) trait AsDevicePtr: Send + 'static {
+    fn as_device_ptr(&self) -> *mut c_void;
+}
+
+impl AsDevicePtr for MTLBuffer {
```
EricLBuehler, line 85: Worth stating explicitly here that the returned ptr stays valid because of an invariant of `to_capsule`. It computes data before moving the buffer into `ManagedCtx`, and the Obj-C object doesn't move when its `Retained` is moved, instead it lives until `managed_tensor_deleter` drops the ctx.

The reasoning is correct, but I think we should probably document this invariant somewhere.

EricLBuehler (changes requested):
Looks largely good! I added one comment about a potential correctness issue and 2 nit comments.

bindings/python/src/metal.rs
McPatate: I should've caught this 😬

bindings/python/src/lib.rs
McPatate: yes definitely agree, also should've thought of that 🦭

bindings/python/src/dlpack.rs
McPatate: done!

PR #768, opened 2026-05-14
feat: add support for strided slicing

McPatate:
Avoid the need to schedule multiple `get_slice` calls and concatenate the resulting tensors, useful for distributed loading patterns.

cf code in https://github.com/huggingface/transformers/pull/45028/changes#diff-d53b11aa21cc2aff5866078be811ee195227f696ccf28a44588f2a155ccfc800R217

namely the `_slice_and_cat` function would be removed. this avoids multiple allocations and unecessary memcpys

cc @3outeille @ArthurZucker 

Snippet:
```py
  import tempfile, torch

  shape = (4096, 1024)
  src = torch.arange(shape[0] * shape[1], dtype=torch.float32).reshape(shape).contiguous()
  with tempfile.NamedTemporaryFile(suffix=".safetensors", delete=False) as fp:
      path = fp.name
  save_file({"fused": src}, path)

  # _StridedShard(dim=0, split_factor=2) for rank 0 of TP=4: interleaved quarters.
  intervals = [(0, 512), (2048, 2560)]
  device = "cuda:0" if torch.cuda.is_available() else "cpu"

  with safe_open(path, framework="pt", device=device) as f:
      got = f.get_strided_slice("fused", dim=0, intervals=intervals)
      expected = torch.cat([f.get_slice("fused")[a:b, :] for a, b in intervals], dim=0)
      assert torch.equal(got.cpu(), expected.cpu())
      print(got.shape, got.dtype, got.device)
```

### DIFF
--- a/bindings/python/py_src/safetensors/__init__.pyi
+++ b/bindings/python/py_src/safetensors/__init__.pyi
@@ -207,6 +207,47 @@ class safe_open:
         """
         pass
 
+    def get_strided_slice(self, name, dim, intervals):
+        """
+        Read `N` disjoint intervals along a single dimension and pack them
+        into one contiguous tensor — equivalent to
+        `torch.cat([f.get_slice(name)[..., a_i:b_i, ...] for (a_i, b_i) in intervals], dim=dim)`
+        but without materializing the per-interval tensors or the cat. The
+        destination is allocated once on the configured framework/device;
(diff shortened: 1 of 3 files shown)


McPatate:
parking for now, performance stays identical with or without the change. It seems like a good idea, but in practice transformers already //ises so io is overlapped.

PR #769, opened 2026-05-19
feat: slice ellipsis support

McPatate:
split elipsis support from the previous mtl handling PR. will rebase #767 when it is merged, should only keep the elipsis bits of code.


McPatate:
Ended up doing this directly in #767

PR #773, opened 2026-05-25
fix(python): correct load_model return type annotation

### LINKED ISSUE #632: Type of the return value is error.
### System Info

all system

### Information

- [x] The official example scripts
- [ ] My own modified scripts

### Reproduction

https://github.com/huggingface/safetensors/blob/8814598e41e2f840bef3a8f491a952ae999bf7ca/bindings/python/py_src/safetensors/torch.py#L189-L194

The type of the return value should be `Tuple[Set[str], Set[str]]` instead of `Tuple[List[str], List[str]]`

### Expected behavior

The type of the return value is `Tuple[Set[str], Set[str]]`

eunseo9311:
# What does this PR do?

Fixes the incorrect return type annotation on `load_model` in torch.py.

The annotation declared `Tuple[List[str], List[str]]`, but the actual return is `Tuple[Set[str], List[str]]`:
- `missing` is converted via `set(missing)` → Set
- `unexpected` is built with `.append()` → remains a List

Updated both the signature (line 201) and the docstring (line 223). Annotation-only change; no runtime behavior affected.

Note: issue #632 suggested `Tuple[Set, Set]`, but `unexpected` is a List, so the correct type is `Tuple[Set[str], List[str]]`.

Closes #632

## AI model use

Model used: Claude (via Claude Code)

### DIFF
--- a/bindings/python/py_src/safetensors/torch.py
+++ b/bindings/python/py_src/safetensors/torch.py
@@ -198,7 +198,7 @@ def load_model(
     device: Union[str, int] = "cpu",
     *,
     backend: str = "mmap",
-) -> Tuple[List[str], List[str]]:
+) -> Tuple[Set[str], List[str]]:
     """
     Loads a given filename onto a torch model.
     This method exists specifically to avoid tensor sharing issues which are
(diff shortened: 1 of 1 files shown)


eunseo9311:
Friendly bump — happy to revise if anything needs adjustment!

PR #775, opened 2026-05-29
docs(spec): document JSON whitespace tolerance in header

SMC17:
## Summary

The README format specification (lines 78-79) and the canonical Rust implementation disagree about which JSON whitespace bytes a safetensors header may contain, and where. This PR aligns the README with the implementation's actual contract and adds a Rust regression test that pins the round-trip for non-trivial payloads.

## Spec vs. implementation divergence

`README.md` lines 78-79 currently say:

> The header data MUST begin with a `{` character (0x7B).
> The header data MAY be trailing padded with whitespace (0x20).

The implementation diverges from both clauses:

1. **Position.** PR #686 ("refactor: support reading padded/trailing whitespace header", commit 96305fc) explicitly added support for leading whitespace, motivated by data-section page alignment. The regression test `test_whitespace_leading_padded_header` (`safetensors/src/tensor.rs:1545`) deserializes `\x09\x0A{}\x0D\x20` successfully, meaning the header does **not** have to begin with `{`.
2. **Byte set.** Both `test_whitespace_padded_header` (line 1535) and `test_whitespace_leading_padded_header` (line 1545) use all four JSON whitespace bytes (0x20 SPACE, 0x09 TAB, 0x0A LF, 0x0D CR), not just 0x20. The implementation delegates to `serde_json`, which accepts any JSON whitespace anywhere the JSON grammar permits.

Both clauses in the spec therefore under-specify what every alternative implementation must accept to read files produced by the canonical writer (or by other compliant alternative writers). Lock the contract in the spec.

## Fix

`README.md` is updated to:

- Document that the body, after stripping JSON whitespace, MUST begin with `{`.
- Document that JSON whitespace (0x20, 0x09, 0x0A, 0x0D) MAY appear in both leading and trailing positions, and that implementations MUST tolerate either or both.
- Keep 0x20 trailing as the writer recommendation, so existing writers stay valid and alternative writers have an unambiguous default.
- Reaffirm that the padding is counted in the `N` byte length prefix.

`safetensors/src/tensor.rs` gains `test_whitespace_padded_header_with_tensor`, which builds an `I32[1]` tensor with a header padded by `\t\n` leading and `\r ` trailing, then checks that `dtype`, `shape`, and tensor bytes round-trip. The two existing whitespace tests only cover `{}` (zero-tensor) headers, so this closes the obvious gap for alternative-implementation conformance.

## Substrate context

Surfaced while porting safetensors to Zig at `safetensors-zig` (independent implementation, AGPL-3.0). The Zig port reads the README as ground truth; when we ran behavioral tests against files emitted by the canonical writer we hit fully-spec-conformant readers rejecting files that the canonical Python/Rust reader happily accepts. PR #686's commit message acknowledges the gap ("trailing whitespace are already ignored by serde, this won't break the format") but did not update the spec.

## Test coverage

- `cargo test test_whitespace` in `safetensors/`:
  - `test_whitespace_padded_header` (existing): empty header with trailing JSON whitespace.
  - `test_whitespace_leading_padded_header` (existing): empty header with leading JSON whitespace.
  - `test_whitespace_padded_header_with_tensor` (new): non-empty I32[1] tensor with leading `\t\n` and trailing `\r ` padding; checks dtype, shape, and tensor bytes round-trip.

## CI gate verification

- `cargo fmt --check` in `safetensors/`: clean.
- `cargo test` in `safetensors/`: all tests pass (3 whitespace tests including the new one + full pre-existing suite).

## Test plan

- [ ] `cd safetensors && cargo fmt --check`
- [ ] `cd safetensors && cargo test`
- [ ] Review README diff for accuracy and tone consistency with surrounding format section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,8 +75,11 @@ with safe_open("model.safetensors", framework="pt", device="cpu") as f:
 
 - 8 bytes: `N`, an unsigned little-endian 64-bit integer, containing the size of the header
 - N bytes: a JSON UTF-8 string representing the header.
-  - The header data MUST begin with a `{` character (0x7B).
-  - The header data MAY be trailing padded with whitespace (0x20).
+  - The header data, after stripping JSON whitespace, MUST begin with a `{` character (0x7B).
+  - The header data MAY be padded with JSON whitespace (any of 0x20 SPACE, 0x09 TAB, 0x0A LF, 0x0D CR) in
+    both leading and trailing positions. Implementations MUST tolerate either or both. Writers SHOULD use
+    trailing 0x20 padding when padding is needed (e.g. to align the data section to a power-of-two boundary).
+    The padding bytes are included in the `N` byte count.
   - The header is a dict like `{"TENSOR_NAME": {"dtype": "F16", "shape": [1, 16, 256], "data_offsets": [BEGIN, END]}, "NEXT_TENSOR_NAME": {...}, ...}`,
     - `data_offsets` point to the tensor data relative to the beginning of the byte buffer (i.e. not an absolute position in the file),
       with `BEGIN` as the starting offset and `END` as the one-past offset (so total tensor byte size = `END - BEGIN`).
--- a/safetensors/src/tensor.rs
+++ b/safetensors/src/tensor.rs
@@ -1549,6 +1549,31 @@ mod tests {
     }
 
     #[test]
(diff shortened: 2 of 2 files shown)


SMC17:
Sorry — been testing some things and this leaked out. Closing.

PR #776, opened 2026-05-29
tests: add Llama-3.2-shape fixture generator (201 tensors)

SMC17:
## Summary

Adds a stdlib-only Llama-shape fixture generator and a pytest module that
round-trips the full 201-tensor layout through the canonical reader.

The existing Python test suite covers two regimes well: empty headers
(`{}`) and very small synthetic tensors (`(2, 2)` arrays). What it does
not cover is what most users actually hand the library: a real-model
header with hundreds of entries and tensor data running to gigabytes.
This gap is what surfaced the spec/impl divergence in #775. The same
fixture machinery that surfaced #775 is what this PR contributes upstream.

## What this changes

* `bindings/python/tests/fixtures/llama_generator.py`: a pure-stdlib
  generator that emits a Llama-3.2-shaped safetensors file. No
  dependency on `torch`, `safetensors`, or `numpy`; the generator writes
  the binary format directly. Shape is parameterised, default produces
  201 tensors matching the structural inventory of TinyLlama-1.1B /
  Llama-3.2-1B (embedding + 22 layers x 9 tensors + final norm + LM
  head). Determinism is controlled by a `seed` argument routed through
  `random.Random`, so identical inputs produce byte-identical output.

* `bindings/python/tests/test_llama_fixture.py`: nine test cases across
  four `unittest.TestCase` classes covering:
  - Structural integrity: 201-tensor count, contiguous non-overlapping
    offsets, dtype coverage across the Llama-relevant set
    (F16 / BF16 / F32 / I32 / I64).
  - Determinism: same-seed byte-identical output; different-seed data
    divergence while header stays identical.
  - Round-trip: canonical reader (`safetensors.numpy.load_file`) accepts
    the full 201-tensor file; raw data bytes round-trip through the
    reader; `__metadata__` round-trips through `safe_open`.
  - Header invariants: pure-ASCII JSON, well-formed envelope.

## Why this belongs upstream

`safetensors` is a load-bearing dependency for the HuggingFace model
ecosystem; the formats most users ship in are large transformer
checkpoints, not the synthetic shapes the existing suite covers. A
shared fixture generator under `tests/fixtures/` gives the project
three things it does not have today:

1. A regression seed any future PR can use to validate that a parser
   change still accepts the realistic emission order.
2. A no-torch-dependency path to test real-model shapes (the generator
   uses only stdlib, so it runs in any CI matrix configuration).
3. A reproducible source for fixtures of arbitrary size: the same code
   path produces the small fixtures used in this PR's tests and the
   gigabyte-scale fixtures useful for memory-mapping or pread testing.

## Substrate context

The generator was extracted from an independent Zig port of safetensors
(AGPL-3.0). The port reads the README as ground truth, which is how the
JSON-whitespace divergence in #775 first surfaced. The same fixture
shape used to validate that fix is what this PR contributes back. It
pairs naturally with #775: that PR documents what every reader must
accept, this PR gives the suite a non-trivial fixture to verify it
against.

## Test plan

- [ ] `cd bindings/python`
- [ ] `python -m pytest tests/test_llama_fixture.py -v` (9/9 pass)
- [ ] `ruff check bindings/python/tests/fixtures/llama_generator.py bindings/python/tests/test_llama_fixture.py`
- [ ] `ruff format --check bindings/python/tests/fixtures/llama_generator.py bindings/python/tests/test_llama_fixture.py`
- [ ] Inspect generator CLI: `python bindings/python/tests/fixtures/llama_generator.py --help`

## Local verification

Run on Python 3.12 / safetensors 0.7.0 with `numpy` extra installed.

```
$ python -m pytest tests/test_llama_fixture.py -v
============================= test session starts ==============================
collected 9 items

tests/test_llama_fixture.py::LlamaFixtureStructureTestCase::test_default_shape_produces_201_tensors PASSED
tests/test_llama_fixture.py::LlamaFixtureStructureTestCase::test_header_dtype_coverage_includes_real_model_set PASSED
tests/test_llama_fixture.py::LlamaFixtureStructureTestCase::test_offsets_are_contiguous_and_non_overlapping PASSED
tests/test_llama_fixture.py::LlamaFixtureDeterminismTestCase::test_different_seed_changes_data_segment PASSED
tests/test_llama_fixture.py::LlamaFixtureDeterminismTestCase::test_same_seed_produces_byte_identical_output PASSED
tests/test_llama_fixture.py::LlamaFixtureRoundTripTestCase::test_canonical_reader_accepts_full_201_tensor_fixture PASSED
tests/test_llama_fixture.py::LlamaFixtureRoundTripTestCase::test_canonical_reader_round_trips_data_bytes PASSED
tests/test_llama_fixture.py::LlamaFixtureRoundTripTestCase::test_metadata_round_trips_via_safe_open PASSED
tests/test_llama_fixture.py::LlamaFixtureHeaderJsonTestCase::test_header_json_is_pure_ascii_and_well_formed PASSED

============================== 9 passed in 1.49s ===============================
```

### DIFF
--- a/bindings/python/tests/fixtures/llama_generator.py
+++ b/bindings/python/tests/fixtures/llama_generator.py
@@ -1,0 +1,397 @@
+"""Llama-shaped safetensors fixture generator (stdlib-only).
+
+Existing test coverage in this repository exercises empty-header round-trips
+(``{}``) and very small synthetic tensors. Real-model deployments hit headers
+that hold hundreds of entries and tensor data running to gigabytes; bugs that
+only surface at that scale (header parser limits, offset arithmetic, layered
+dtype mixes) are not currently covered by the suite.
+
+This module emits a Llama-3.2-shaped safetensors file using only the Python
+standard library. It depends on neither ``torch`` nor ``safetensors`` itself,
+which keeps it usable for any test that wants a real-shape input regardless of
+which framework extra is installed. The structural shape (embedding +
+N transformer layers with attention/MLP/norms + final norm + LM head) is
+parameterised so the same generator covers both a small-default fixture
+suitable for unit tests and a larger configuration if a downstream test needs
+one.
+
+The output follows the safetensors binary contract exactly:
+
+* bytes ``[0, 8)``: little-endian ``u64`` header length ``N``
+* bytes ``[8, 8 + N)``: UTF-8 JSON header
+* bytes ``[8 + N, EOF)``: contiguous tensor data, dtype-typed and aligned per
+  the writer's emission order
+
+The default shape produces 201 tensors (embedding + 22 layers x 9 tensors per
+layer + final norm + LM head) matching the structural inventory of
+TinyLlama-1.1B / Llama-3.2-1B-Instruct, scaled to a small fixture footprint.
+Determinism is controlled by an explicit ``seed`` argument routed through
+``random.Random``; the same seed produces byte-identical output, which is what
+regression tests want.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import os
+import random
+import struct
+import sys
+from dataclasses import dataclass, field
+from typing import Dict, List, Tuple
+
+
+# Element-size table covering the dtype families exercised by Llama-shaped
+# weights. Bit widths come from the safetensors README dtype table; this map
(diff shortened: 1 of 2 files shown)


SMC17:
Sorry — been testing some things and this leaked out. Closing.

PR #789, opened 2026-06-13
fix: preserve file permissions in buffered_write_to_file

### LINKED ISSUE #782: Files created by buffered_write_to_file are chmod 600
### System Info

Starting with version 0.8.1-rc1, following https://github.com/safetensors/safetensors/pull/764, files created by buffered_write_to_file() are chmod 600 whereas they were previously chmod 666.

This means for example that after using safetensors to train/fine-tune a model, the model files end up usable by the trainer only, and not by other previously allowed users.

In my particular, using another user to infer with a fine tuned model triggers `vllm.exceptions.LoRAAdapterNotFoundError: Loading lora model failed: No adapter found` even though the `model_adapter.safetensors` file exists - but it's not accessible by the other user.

That being said, it's not a bug in safetensors, but rather an undocumented behaviour change between 0.8.0-rc0 and 0.8.0-rc1

### Information

- [ ] My own modified scripts

### Reproduction

1. Use a standard peft finetuning loop to finetune a model
2. Try to infer using the finetuned model with an other unix user 

### Expected behavior

At step 2, using the finetuned model should be possible for an other unix user

discobot:
# What does this PR do?

Fixes #782.

Since #764, `serialize_to_file` creates a `tempfile::NamedTempFile` and renames
it into place. The tempfile crate always creates files with mode 0o600, and
`persist()` is just a rename, so the destination keeps 0o600 instead of the mode
a plain `File::create(path)` would have produced — 0o666 masked by the process
umask for a new file, or the existing file's mode when overwriting.

This restores both behaviors on unix before the `persist()` call: if the
destination already exists, its permission bits are copied onto the temp file;
otherwise the temp file gets `0o666 & !umask`, reading the umask with the usual
`umask(0)`-then-restore dance. The `libc` dependency is widened from macOS
(where it was already used for `F_NOCACHE`) to all unix targets. Windows is
unaffected.

Tested with a new unix-only unit test in `tensor.rs` (a fresh file is 0o644
under umask 022, and an existing file's 0o640 survives an overwrite) plus the
same checks through `save_file` in `bindings/python/tests/test_simple.py`. The
Rust test fails on unpatched `main` (mode 0o600), and the #762 mmap regression
test still passes.

## AI model use

version that you used:

Claude (Anthropic), via Claude Code.

If you ticked the last option, please list the prompt(s) that you used:

Investigate huggingface/safetensors issue #782 (files written by
`serialize_to_file` come out chmod 600 since the tempfile+rename change in
#764), find the root cause, and write a minimal fix that restores the prior
permission behavior (umask-derived mode for new files, preserved mode when
overwriting) with regression tests on both the Rust and Python sides.

### DIFF
--- a/safetensors/src/tensor.rs
+++ b/safetensors/src/tensor.rs
@@ -336,6 +336,28 @@ fn buffered_write_to_file<V: View>(
         f.flush()?;
     }
 
+    // `NamedTempFile` is always created with mode 0o600, which the rename
+    // below would otherwise hand over to `path` (#782). Restore the
+    // permissions a plain `File::create(path)` would have produced: the
+    // existing file's mode when overwriting, else 0o666 masked by the
+    // process umask.
+    #[cfg(unix)]
+    {
+        use std::os::unix::fs::PermissionsExt;
+
+        let permissions = match std::fs::metadata(path) {
+            Ok(metadata) => metadata.permissions(),
+            Err(_) => {
+                // Reading the process umask requires setting it; restore it
+                // right away.
+                let umask = unsafe { libc::umask(0) };
+                unsafe { libc::umask(umask) };
+                std::fs::Permissions::from_mode(0o666 & !u32::from(umask))
+            }
+        };
+        temp.as_file().set_permissions(permissions)?;
+    }
+
     temp.persist(path).map_err(|e| e.error)?;
 
     Ok(())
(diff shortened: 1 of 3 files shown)


safetensors/src/tensor.rs
```diff
@@ -336,6 +336,28 @@ fn buffered_write_to_file<V: View>(
...
+    #[cfg(unix)]
+    {
+        use std::os::unix::fs::PermissionsExt;
+
+        let permissions = match std::fs::metadata(path) {
+            Ok(metadata) => metadata.permissions(),
+            Err(_) => {
+                // Reading the process umask requires setting it; restore it
+                // right away.
+                let umask = unsafe { libc::umask(0) };
```
nurpax, line 353: The umask business here seems complicated (and potentially unsafe threading wise?).

I'm not a rust developer, but the tempfile crate seems to support setting umask filtered permissions for the new file:

https://docs.rs/tempfile/latest/tempfile/struct.Builder.html#method.permissions

Or is there something particular about how NamedTempFile is used above that makes this harder?

nurpax (commented):
We just hit the problem this change is fixing and definitely would support fixing this one way or another.

yz-ppl:
We also hit the same issue. Can this fix be merged and added to the next release?

PR #791, opened 2026-06-13
fix(python): restore committed Cargo.lock for --locked builds

### LINKED ISSUE #785: Cargo.lock missing in release 0.8.0
### System Info

The 0.8.0 release has removed the `bindings/python/Cargo.lock`, specifically in commit b4a9ec783cb5c6e4a5c0f788ffe8fd04f00f2282

This goes against rust best practices as it breaks building with `maturin build --locked`.

### Information

- [ ] My own modified scripts

### Reproduction

```
$ cd bindings/python
$ maturin build --locked --release --target "$(rustc --print host-tuple)" --strip
    Updating crates.io index
error: cannot create the lock file /build/python-safetensors/src/safetensors-0.8.0/bindings/python/Cargo.lock because --locked was passed to prevent this
help: to generate the lock file without accessing the network, remove the --locked flag and use --offline instead.
💥 maturin failed
  Caused by: Cargo metadata failed. Does your crate compile with `cargo build`?
  Caused by: `cargo metadata` exited with an error:
```

### Expected behavior

`maturin build --locked` should work

discobot:
# What does this PR do?

Fixes #785.

The lockfile deletion in b4a9ec7 was collateral damage: that commit is the revert of #551, and it dropped the whole file even though it had been tracked on purpose since #233. A stale `bindings/python/Cargo.lock` entry in the root `.gitignore` is what let the deletion go unnoticed.

That also answers the question in the issue thread: 0.7.0 still shipped a lockfile only because the v0.7.0 release branch re-added one in its "Set version to 0.7.0" commit (fa27fa8); `main` — and therefore the v0.8.0 tag — never got it back.

This restores the lockfile (regenerated with `cargo generate-lockfile`; it comes out as lockfile version 3 because of `rust-version = "1.74"`, the same format 0.7.0 shipped) and removes the stale ignore line so it stays tracked. The `safetensors/**/Cargo.lock` ignore for the core library crate is untouched.

Verified with cargo 1.89 / maturin 1.13.3 on macOS arm64: the issue's exact `maturin build --locked` command fails before and builds the wheel after, with the committed lockfile byte-identical after the build; `git archive HEAD` (what GitHub tag tarballs contain) now includes `bindings/python/Cargo.lock`, and building from the unpacked archive with `--locked` works; `maturin sdist` ships the committed lockfile byte-for-byte.

## AI model use

version that you used:

Anthropic Claude (Opus 4.x), driven via Claude Code.

If you ticked the last option, please list the prompt(s) that you used:

The model was asked to investigate issue #785 (Cargo.lock missing from the 0.8.0 source release breaking `maturin build --locked`), determine the root cause from git history, restore the committed `bindings/python/Cargo.lock` and remove the stale `.gitignore` entry so it stays tracked, and verify the fix end to end with the issue's exact build command before opening this PR.


McPatate:
closing in favour of #794, ty for the contribution!

PR #794, opened 2026-06-15
feat(ci): lock dependencies for reproducible release builds

### LINKED ISSUE #785: Cargo.lock missing in release 0.8.0

McPatate:
Locking deps to mitigate future supply chain attacks.

Supersedes #791

closes #785


danieldk (approved):
Once Cargo cooldown is also supported client-side, we should probably use it as well. (Not a fan of using the third-party cargo-cooldown.)

PR #796, opened 2026-06-15
feat: bump pyo3 to `0.29`

McPatate:
Required to fix the `cargo audit`. While reviewing the changelog, discovered that there was a proper `PyCapsule` type we could use instead of calling raw ffi `PyCapsule_New`.

Removes one `unsafe` block :)

cf https://pyo3.rs/v0.29.0/migration.html


danieldk (approved):
Nice!

danieldk:
Clippy wants an import fixed.

danieldk (approved):

PR #801, opened 2026-06-18
Make file preallocation best-effort in buffered_write_to_file

### LINKED ISSUE #787: serialize_to_file fails on filesystems that don't support ftruncate (via set_len)
## Summary

`safetensors::tensor::serialize_to_file` (and therefore the Python `save_file`)
fails on my system because it calls `File::set_len` for preallocation, which
maps to `ftruncate(2)`. My filesystem does not support `ftruncate`, so the call
returns an error and the whole save operation aborts.

## System info

- safetensors version: [e.g. 0.8.0]
- Filesystem: [ FUSE HDFS]
- Python version [3.11]

## Where it happens

In `safetensors/src/tensor.rs`, inside `buffered_write_to_file`:

```rust
temp.as_file().set_len(total_size as u64)?;

### Motivation

Expected behavior
Saving should succeed even when the target filesystem does not support ftruncate.

Analysis
As far as I can tell, this set_len call is purely a preallocation optimization: the subsequent sequential write_all calls already grow the file to exactly total_size bytes (N_LEN + header_bytes.len() + offset). So the preallocation is not required for correctness — the produced file is identical without it.

### Your contribution

Proposed fix
Make the preallocation best-effort instead of fatal, so it gracefully degrades on filesystems where ftruncate is unsupported, while keeping the optimization everywhere else:

// Best-effort preallocation; sequential writes below still grow the file to
// `total_size`, so skip it where ftruncate is unsupported.
let _ = temp.as_file().set_len(total_size as u64);
I'm happy to open a PR if this approach sounds reasonable.

Workaround (for other users hitting this)
Use the in-memory save() path, which never touches ftruncate, and write the bytes yourself:


from safetensors.torch import save
with open("model.safetensors", "wb") as f:
    f.write(save(tensors))

palios-taey:
# What does this PR do?

Fixes #787. `serialize_to_file` currently aborts on filesystems that don't support `ftruncate` (e.g. FUSE HDFS, some S3 mounts) because `buffered_write_to_file` propagates the `File::set_len` error. Preallocation is only an allocation hint — the sequential writes already grow the file to exactly `total_size` (the in-memory `serialize` path never preallocates) — so this makes `set_len` best-effort by ignoring its result, with a comment noting why.

Orthogonal to #793, which addresses a FUSE rename failure; this addresses the `ftruncate`/`set_len` preallocation path. Happy to coordinate if you'd prefer them consolidated.

## AI model use

### DIFF
--- a/safetensors/src/tensor.rs
+++ b/safetensors/src/tensor.rs
@@ -312,7 +312,8 @@ fn buffered_write_to_file<V: View>(
     let parent = path.parent().unwrap_or_else(|| Path::new("."));
     let temp = tempfile::NamedTempFile::new_in(parent)?;
 
-    temp.as_file().set_len(total_size as u64)?;
+    // Best-effort preallocation: ignore failure on filesystems without ftruncate support (FUSE/HDFS/S3). The sequential writes below still produce a correct file; a genuine out-of-space error surfaces there.
+    let _ = temp.as_file().set_len(total_size as u64);
 
     // Serialize tensors to a file using direct I/O (bypassing page cache) using F_NOCACHE.
     // This yields ~30% performance improvement.
--- a/safetensors/src/tensor.rs
+++ b/safetensors/src/tensor.rs
@@ -312,8 +312,14 @@ fn buffered_write_to_file<V: View>(
     let parent = path.parent().unwrap_or_else(|| Path::new("."));
     let temp = tempfile::NamedTempFile::new_in(parent)?;
 
-    // Best-effort preallocation: ignore failure on filesystems without ftruncate support (FUSE/HDFS/S3). The sequential writes below still produce a correct file; a genuine out-of-space error surfaces there.
-    let _ = temp.as_file().set_len(total_size as u64);
+    // Preallocation is best-effort only when the filesystem reports it as unsupported
+    // (e.g. FUSE/HDFS/S3 ftruncate not implemented). The sequential writes below grow the
+    // file to total_size regardless; surface all other errors such as overflow, space, or IO.
+    if let Err(e) = temp.as_file().set_len(total_size as u64) {
+        if e.kind() != std::io::ErrorKind::Unsupported {
+            return Err(e.into());
+        }
+    }
 
     // Serialize tensors to a file using direct I/O (bypassing page cache) using F_NOCACHE.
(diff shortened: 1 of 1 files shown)


Mesilenceki:
LGTM

chijinxina:
LGTM

McPatate:
Hi thanks for this PR. I think we don't want to ignore the error here, as there could be legitimate errors that'd be nice to surface early, such as:
> Also, [std::io::ErrorKind::InvalidInput](https://doc.rust-lang.org/beta/std/io/enum.ErrorKind.html#variant.InvalidInput) will be returned if the desired length would cause an overflow due to the implementation specifics.
> \-
> _https://doc.rust-lang.org/beta/std/fs/struct.File.html#errors-2_

I think that error is different than the one you're trying to ignore. I'd rather have a `match` or pattern match if on the specific error type making sure we don't ignore legitimate error cases. If you can show me / guarantee this is the case, then we can merge as is!

palios-taey:
Thanks @McPatate — agreed. Updated to match on the error kind: preallocation now degrades only when the filesystem reports the operation as `Unsupported` (the `EOPNOTSUPP`/`ENOSYS` that `ftruncate` returns when a filesystem doesn't implement it), and propagates everything else.

On your specific point — the overflow `InvalidInput` comes from `EINVAL`, which std maps distinctly from the unsupported case. In `library/std/src/sys/io/error/unix.rs`: `EINVAL => InvalidInput`, while `ENOSYS => Unsupported` and `EOPNOTSUPP => Unsupported`. So the `!= ErrorKind::Unsupported` guard surfaces overflow, out-of-space, and IO errors as before, and only skips preallocation where the OS reports it genuinely unsupported.

chijinxina:
Hi @McPatate @danieldk, friendly ping on this PR. The latest revision addresses the review concern by ignoring only `ErrorKind::Unsupported`, while still propagating `InvalidInput` and all other errors. The PR is mergeable and DCO is passing. Could one of you please take another look and let us know if anything still blocks approval or merge? Thanks!

smallfryhlx:
hi @McPatate @danieldk, Our system also does not support the ftruncate operation. The newer version of safetensor is currently unusable. Can this MR be merged?

PR #808, opened 2026-07-05
Guard against integer overflow in the parse path (TensorView::new + read_metadata)

v-code01:
## Problem

Two spots in the parse path use unchecked integer arithmetic on attacker-influenced values, so they **panic** (debug builds) or silently **wrap** (release) instead of returning an error — even though their sibling code paths are already guarded:

1. **`TensorView::new`** computes `shape.iter().product()` and `n_elements * dtype.bitsize()` unchecked. An overflowing shape panics instead of returning `ValidationOverflow` (the `deserialize` path guards the identical arithmetic with `checked_mul`).

   ```rust
   TensorView::new(Dtype::I32, vec![usize::MAX, 2], &[]); // panics
   ```

2. **`read_metadata`**'s final length check `buffer_end + N_LEN + n` is derived from attacker-controlled `data_offsets`. A crafted multi-tensor header whose contiguous offsets sum near `usize::MAX` overflows this add — `attempt to add with overflow` — while the header `stop = n.checked_add(N_LEN)` right above it is already guarded. Since safetensors deserializes untrusted files, this is a DoS on malformed input (and a release-build wrap could bypass the length check).

## Fix

Use checked arithmetic in both spots, returning `ValidationOverflow` on overflow, matching the existing guarded paths.

## Tests

- `test_tensorview_new_overflow` — element-count product and `n_elements * bitsize` overflow.
- `test_read_metadata_offset_sum_overflow` — a crafted 8-tensor header whose offsets sum near `usize::MAX`.

```
cargo test    # 38 lib tests + integration/doctests green
cargo fmt/clippy  # clean on the changed file
```

### DIFF
--- a/safetensors/src/tensor.rs
+++ b/safetensors/src/tensor.rs
@@ -417,7 +417,15 @@ impl<'data> SafeTensors<'data> {
             serde_json::from_str(string).map_err(SafeTensorError::InvalidHeaderDeserialization)?;
         let metadata: Metadata = metadata.try_into()?;
         let buffer_end = metadata.validate()?;
-        if buffer_end + N_LEN + n != buffer_len {
+        // `buffer_end` is derived from attacker-controlled offsets, so add with
+        // overflow checks (the header `stop` above is guarded the same way): a
+        // crafted header can otherwise drive this near usize::MAX and panic
+        // (debug) or wrap and bypass the length check (release).
+        let expected_len = buffer_end
+            .checked_add(N_LEN)
+            .and_then(|v| v.checked_add(n))
+            .ok_or(SafeTensorError::ValidationOverflow)?;
+        if expected_len != buffer_len {
             return Err(SafeTensorError::MetadataIncompleteBuffer);
         }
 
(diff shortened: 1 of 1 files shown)


v-code01:
cc @Narsil — small robustness fix: `TensorView::new` used unchecked `product()`/`*` for the shape and nbits, so an overflowing shape panicked instead of returning `ValidationOverflow` like the `deserialize` path already does. Mirrored the checked arithmetic + added a regression test.

Narsil:
Not a maintainer anymore, but LGTM, I'm wondering if we couldn't enforce overflow checks in release builds by default tbh.

Narsil:
@McPatate

v-code01:
Thanks! Agree it's worth considering. One caveat on `overflow-checks = true` in `[profile.release]`: that setting is only honored for the crate at the root of the build, so a downstream app that pulls `safetensors` in as a dependency and builds in its own default release profile (overflow-checks off) won't inherit it. The checked arithmetic here holds regardless of how the consumer compiles, which is why I kept it explicit at the parse boundary rather than leaning on the profile.

Happy to also add `overflow-checks = true` to the workspace's own dev/test profiles as belt-and-suspenders for CI if you'd like, but I'd keep the explicit checks either way since they're what protects downstream users.

Narsil:
Good point !
```
[workspace.lints.clippy]
arithmetic_side_effects = "deny"
```
Should be enough then

v-code01:
Yeah, that'd catch this class at lint time going forward, which is nice. Two caveats worth flagging before turning it on workspace-wide:

- `arithmetic_side_effects` is a `restriction`-group lint, so it denies *all* integer `+`/`-`/`*` (not just the genuinely-unchecked ones). It'll light up a lot of existing arithmetic that's provably safe, so it needs a pass to convert the real hotspots to `checked_*`/`saturating_*` and `#[allow]` (with a reason) the ones that are fine — a decent bit of churn.
- It only covers integer overflow, not the offset/bounds logic itself, so the explicit validation in this PR is still doing work the lint wouldn't.

I'm happy to open a follow-up that wires up the lint and does the cleanup pass across the crate if you want it — I'd keep this PR as the concrete fix and land the lint separately so the guardrail doesn't hold up the actual bug fix. Sound good?

PR #809, opened 2026-07-05
Guard slice bound resolution against index overflow

v-code01:
## Problem

Resolving slice bounds computes `*s + 1` in three places — an `Excluded` start bound, an `Included` stop bound, and `Select(s)`:

```rust
Bound::Excluded(s) => *s + 1,
Bound::Included(s) => *s + 1,
TensorIndexer::Select(s) => (*s, *s + 1, 1),
```

A `usize::MAX` index therefore **overflows** (panic in debug, wrap in release) *before* the out-of-range check against the dimension size can reject it:

```rust
slice_byte_ranges(Dtype::F32, &[10], &[TensorIndexer::Select(usize::MAX)]); // panics
```

## Fix

Use `saturating_add(1)`. A `usize::MAX` index saturates to `usize::MAX`, which the existing check then rejects with `SliceOutOfRange`. For any in-range index the behaviour is identical (`s + 1 == s.saturating_add(1)` for `s < usize::MAX`).

## Tests

`test_slice_usize_max_index_out_of_range` covers `Select(usize::MAX)` and `Narrow` with `Excluded`/`Included` `usize::MAX` bounds — all now return `SliceOutOfRange` instead of panicking.

```
cargo test    # 37 lib + integration/doctests green
```

### DIFF
--- a/safetensors/src/slice.rs
+++ b/safetensors/src/slice.rs
@@ -314,14 +314,16 @@ pub type SlicedShape = Vec<usize>;
 /// Resolve a `(start, stop)` half-open element range from slice bounds,
 /// defaulting unbounded ends to `0` and `dim`.
 fn narrow_bounds(left: &Bound<usize>, right: &Bound<usize>, dim: usize) -> (usize, usize) {
+    // saturating_add avoids overflow on a usize::MAX bound; the saturated value
+    // is then rejected by the out-of-range check against `dim` in the caller.
     let start = match left {
         Bound::Unbounded => 0,
         Bound::Included(s) => *s,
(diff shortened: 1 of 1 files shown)


v-code01:
Follow-up to the parse-path overflow guards in #808, on the user-slicing side: `slice_byte_ranges` resolved `Select(s)`/`Narrow` bounds with `*s + 1`, so a `usize::MAX` index panicked before the `SliceOutOfRange` check. `saturating_add` lets the existing bounds check reject it cleanly. cc @Narsil

Narsil:
@McPatate

PR #810, opened 2026-07-06
Raise PermissionError instead of FileNotFoundError when safe_open lacks read access

### LINKED ISSUE #805: `safe_open` raises FileNotFoundError instead of PermissionError when read access is denied
### System Info

safetensors 0.8.0, Python 3.13, Linux

### Reproduction

```python
import os, tempfile, torch

path = os.path.join(tempfile.mkdtemp(), "m.safetensors")
save_file({"w": torch.tensor([1.0])}, path)
os.chmod(path, 0o000)

safe_open(path, framework="pt")
# FileNotFoundError: No such file or directory: /tmp/.../m.safetensors
# Expected: PermissionError
```

### Expected behavior

`safe_open` should raise `PermissionError` (errno EACCES) when read access is denied, not `FileNotFoundError` (errno ENOENT). These are distinct OS errors with distinct Python exception types.

### Impact

The wrong exception type is actively misleading. We hit this in production when our quantizer (running as root) wrote `model.safetensors` with `0o600` perms (the root cause of that being #782) — vLLM in a different container, running as non-root, then failed with:

```
FileNotFoundError: No such file or directory: /llm-serving/model.safetensors
```

The file was clearly there. The error sent us chasing dependency versions (`compressed-tensors`, container images, model formats) for hours before we thought to check permissions. A correct `PermissionError` would have pointed straight at the cause.

Note: even after #782/#789 ship, this exception-type bug remains and will surface for any unrelated permission scenario (NFS, AppArmor, container UID mismatches, manual chmod).

Awshesh12:
## Summary

`safe_open` mapped **every** `File::open` failure to `FileNotFoundError`, so a file that exists but is unreadable (EACCES) was reported as a misleading `No such file or directory` instead of `PermissionError`. These are distinct OS errors (ENOENT vs EACCES) with distinct Python exception types, and callers that branch on the exception type can't tell "missing" from "denied".

Reproduction (from #805):

```python
import os, tempfile, torch

path = os.path.join(tempfile.mkdtemp(), "m.safetensors")
save_file({"w": torch.tensor([1.0])}, path)
os.chmod(path, 0o000)
safe_open(path, framework="pt")   # was FileNotFoundError, now PermissionError
```

## Change

Match on `io::ErrorKind` in `Open::new`:
- `PermissionDenied` → `PermissionError`
- `NotFound` → `FileNotFoundError` (unchanged message)
- anything else → `OSError`, preserving the underlying error text

The existing `FileNotFoundError` path and message are untouched.

## Testing

Added `ErrorsTestCase.test_permission_denied` (skipped on Windows / when running as root). Verified against a freshly built wheel:

```
tests/test_simple.py::ErrorsTestCase::test_file_not_found PASSED
tests/test_simple.py::ErrorsTestCase::test_permission_denied PASSED
```

Closes #805

### DIFF
--- a/bindings/python/src/lib.rs
+++ b/bindings/python/src/lib.rs
@@ -7,7 +7,7 @@ mod metal;
 
 use core::slice;
 use memmap2::{Mmap, MmapOptions};
-use pyo3::exceptions::{PyException, PyFileNotFoundError};
+use pyo3::exceptions::{PyException, PyFileNotFoundError, PyIOError, PyPermissionError};
 use pyo3::prelude::*;
 use pyo3::sync::OnceLockExt;
 use pyo3::types::IntoPyDict;
@@ -665,11 +665,15 @@ impl Open {
         device: Option<Device>,
         backend: Backend,
     ) -> PyResult<Self> {
-        let file = File::open(&filename).map_err(|_| {
-            PyFileNotFoundError::new_err(format!(
+        let file = File::open(&filename).map_err(|e| match e.kind() {
+            std::io::ErrorKind::PermissionDenied => {
+                PyPermissionError::new_err(format!("Permission denied: {}", filename.display()))
+            }
+            std::io::ErrorKind::NotFound => PyFileNotFoundError::new_err(format!(
                 "No such file or directory: {}",
                 filename.display()
-            ))
+            )),
+            _ => PyIOError::new_err(format!("Unable to open {}: {e}", filename.display())),
         })?;
         let device = device.unwrap_or(Device::Cpu);
         if device != Device::Cpu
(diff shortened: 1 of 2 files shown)


YusefSyed:
Hi @Awshesh12 — apologies; I independently opened #838 before noticing that this PR already covered #805. I’m closing mine in favor of your earlier work. In comparing them, I found two potentially useful refinements: PyO3’s native `io::Error` conversion retains more Python exception subclasses than a generic `OSError` catch-all, and the permission test can use effective UID plus restoration of the original mode. One caveat is that PyO3 0.29 constructs these exceptions from a single string, so neither implementation currently preserves `.errno` or `.filename`; if #805’s “errno EACCES” is intended as a structured API contract, those attributes need an explicit test and constructor adjustment. Happy to provide a focused patch or review if you or the maintainers would find that useful.

Awshesh12:
@YusefSyed Thanks for the careful comparison. Pushed both: the test now uses geteuid and restores the original mode, and PermissionError carries errno and filename

PR #822, opened 2026-07-29
Faster get_tensors pipeline for CUDA

malfet:
# What does this PR do?

Speed up `load_file` by using batching and interleaving storage IO and Host-to-Device copy

Before this change `load_file(device="cuda")` simply called `get_tensor` in a loop, that allocate a pinned buffer, fill it from the file, then transferr to GPU via blocking `.to("cuda")` call, i.e. read and copy would never overlap.

This commit allocates allocates every destination tensor upfront, stages file reads through a dual  pinned ring buffers filled by striped worker threads (mmap/torch-storage memcpy or parallel pread), and enqueues async copy_ per tensor/chunk overlap on the target device's current stream, so the CPU fills chunk k+1 while the copy engine drains chunk k. Destinations are flat uint8 allocations viewed to their real dtype/shape at the end, which keeps the views zero-copy even for dtype-misaligned file offsets, tensors larger than a ring buffer are streamed through it in chunks, so peak host memory use is bounded by the ring buffer size, not the largest tensor size.

The ring buffer size is controllable via a new keyword-only staging_bytes arg on get_tensors, forwarded from load_file (default 256MiB per buffer): lower it on memory-constrained hosts, raise it for potentially more bandwidth.

Benchmarks (benches/test_pt.py -k gpu, gpt2-12 ~550MB, warm page cache, torch 2.13, median):

| platform                            | test               | before  | after   | speedup |
|-------------------------------------|--------------------|---------|---------|---------|
| GB300 (aarch64, coherent C2C)       | sf_load_gpu (mmap) | 95.3ms  | 12.6ms  | 7.6x    |
| GB300 (aarch64, coherent C2C)       | sf_load_gpu_pread  | 51.6ms  | 11.1ms  | 4.7x    |
| RTX PRO 6000 Blackwell (x86, PCIe)  | sf_load_gpu (mmap) | 60.1ms  | 25.6ms  | 2.35x   |
| RTX PRO 6000 Blackwell (x86, PCIe)  | sf_load_gpu_pread  | 48.2ms  | 17.6ms  | 2.73x   |

Partially addresses https://github.com/safetensors/safetensors/issues/729

## AI model use

version that you used:
Claude Opus 4.8

### DIFF
--- a/bindings/python/py_src/safetensors/__init__.pyi
+++ b/bindings/python/py_src/safetensors/__init__.pyi
@@ -233,9 +233,13 @@ class safe_open:
         """
         pass
 
-    def get_tensors(self):
+    def get_tensors(self, *, staging_bytes: Optional[int] = None):
         """
         Returns every tensor in the file as a dict keyed by name.
+
+        `staging_bytes` sizes each buffer of the pinned staging ring used
+        when loading to a CUDA (or ROCm) device (default 256MiB per buffer,
+        two buffers live); other devices ignore it.
 
         Equivalent to iterating `offset_keys()` and calling `get_tensor` on
         each, but specific `framework` + `device` combinations take an internal
--- a/bindings/python/src/lib.rs
+++ b/bindings/python/src/lib.rs
@@ -650,6 +650,10 @@ impl Version {
     }
 }
 
+/// Default per-buffer size of the pinned staging ring used by the bulk CUDA
+/// load path (`get_tensors_cuda`); two buffers of this size are live at once.
+const DEFAULT_STAGING_BYTES: usize = 256 << 20;
+
 struct Open {
     metadata: Metadata,
     offset: usize,
@@ -1253,7 +1257,20 @@ impl Open {
     /// Shared `MTLBuffer`s, parallel-`pread`s into them, and hands them to
     /// torch via DLPack (1x model memory, the MTLBuffer is the destination).
     /// The `mmap` backend uses the sequential loop, reading through the mmap.
-    pub fn get_tensors(&self) -> PyResult<Py<PyDict>> {
+    /// Pytorch on CUDA uses a bulk pipelined path (`get_tensors_cuda`) whose
+    /// pinned staging ring is sized by `staging_bytes` (`None` = 256 MiB per
+    /// buffer, two buffers live).
+    pub fn get_tensors(&self, staging_bytes: Option<usize>) -> PyResult<Py<PyDict>> {
+        let staging_bytes = match staging_bytes {
+            Some(0) => {
+                return Err(SafetensorError::new_err(
+                    "staging_bytes must be a positive number of bytes",
+                ))
+            }
+            Some(n) => n,
+            None => DEFAULT_STAGING_BYTES,
+        };
+
         // The bulk parallel-pread path is gated on the `pread` backend so the
         // `backend` choice is honored: `mmap` falls through to the per-tensor
         // loop below, which reads through the mmap via `get_tensor_mps`.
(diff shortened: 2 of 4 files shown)


bindings/python/py_src/safetensors/__init__.pyi
```diff
@@ -233,10 +233,14 @@ class safe_open:
         """
         pass
 
-    def get_tensors(self):
+    def get_tensors(self, *, staging_bytes: Optional[int] = None):
```
McPatate, line 236: I'm not super stoked on making this a public parameter for the function: it's only linked to the cuda + torch special path you added and I'm not sure a user should/would know how to tune that.

I do understand that it may be needed, because different devices might have different budgets for these buffers, couldn't we manage to have a simple-ish heuristic + good defaults?

bindings/python/src/lib.rs
```diff
@@ -650,6 +650,10 @@ impl Version {
     }
 }
 
+/// Default per-buffer size of the pinned staging ring used by the bulk CUDA
+/// load path (`get_tensors_cuda`); two buffers of this size are live at once.
+const DEFAULT_STAGING_BYTES: usize = 256 << 20;
```
McPatate, line 655:
```suggestion
const DEFAULT_STAGING_BYTES: usize = 256 * 1024 * 1024;
```
reads a bit better imo

bindings/python/src/lib.rs
```diff
@@ -1253,7 +1257,20 @@ impl Open {
     /// Shared `MTLBuffer`s, parallel-`pread`s into them, and hands them to
     /// torch via DLPack (1x model memory, the MTLBuffer is the destination).
     /// The `mmap` backend uses the sequential loop, reading through the mmap.
-    pub fn get_tensors(&self) -> PyResult<Py<PyDict>> {
+    /// Pytorch on CUDA uses a bulk pipelined path (`get_tensors_cuda`) whose
+    /// pinned staging ring is sized by `staging_bytes` (`None` = 256 MiB per
+    /// buffer, two buffers live).
+    pub fn get_tensors(&self, staging_bytes: Option<usize>) -> PyResult<Py<PyDict>> {
```
McPatate, line 1263:
```suggestion
    pub fn get_tensors(&self, staging_bytes: Option<NonZeroUsize>) -> PyResult<Py<PyDict>> {
```

bindings/python/src/lib.rs
```diff
@@ -1331,6 +1352,241 @@ impl Open {
...
+    /// overlap is enqueued on the target device's current stream: while the
+    /// copy engine drains chunk `k`, worker threads fill chunk `k+1`. Peak
+    /// extra host memory is the pinned ring (`CUDA_STAGE_*`), and one final
+    /// event sync materializes everything before returning.
+    fn get_tensors_cuda(&self, staging_bytes: usize) -> PyResult<Py<PyDict>> {
+        // Two pinned ring buffers of `staging_bytes` each (default 256 MiB:
+        // large enough that per-chunk costs — stripe thread spawns, event
+        // syncs — amortize, small enough that the transient host allocation
+        // stays negligible next to the model).
+        const CUDA_STAGE_BUFFERS: usize = 2;
```
McPatate, line 1371: why would the size be a public configuration knob, but not the number of buffers?

McPatate (commented):
Thank you for the contribution! This is a very torch-centric implementation 😄 

I'm a bit worried that we have the GIL everywhere.

For the cuda fast path, I'm inclined we go the same road as we did for the mps backend: we create custom native cuda buffers, and do the framework handover with dlpack. It'll be much easier to support new frameworks that way, since we don't have a hard dep on framework internals. It's probably more work, but I think this is more aligned with safetensors' initial mission which is to bring safe tensor loading to all libraries, even if pytorch is in fact probably our biggest target!

Addendum: I also think it'll be easier to maintain over the long term to have the native buffer handles and I believe performance will most likely be better

PR #828, opened 2026-08-18
feat: add prefetch

McPatate:


bindings/python/src/engine/cuda.rs
```diff
@@ -0,0 +1,338 @@
...
+
+fn open_cudart() -> *mut c_void {
+    unsafe {
+        if let Some(path) = find_loaded_cudart() {
+            let h = libc::dlopen(path.as_ptr(), libc::RTLD_NOW | libc::RTLD_NOLOAD);
+            if !h.is_null() {
+                return h;
+            }
+        }
+        for name in ["libcudart.so.13", "libcudart.so.12", "libcudart.so"] {
```
takeshi-yoshimura, line 252: This looks fairly ad hoc: it hard-codes CUDA 13 and 12 so names, misses older and future versions, and may load a runtime different from the one used by the framework.

Could the runtime library name/path be resolved by the Python/framework layer and passed explicitly to api(), as done in fastsafetensors?

bindings/python/src/engine/cuda.rs
```diff
@@ -0,0 +1,338 @@
...
+pub type Event = *mut c_void;
+pub type MemPool = *mut c_void;
+type Err = c_int;
+
+const CUDA_MEMCPY_HOST_TO_DEVICE: c_int = 1;
+const CUDA_EVENT_DISABLE_TIMING: c_uint = 0x2;
+const CUDA_STREAM_NON_BLOCKING: c_uint = 0x1;
+const CUDA_MEMPOOL_ATTR_RELEASE_THRESHOLD: c_int = 4;
+
+macro_rules! cuda_fns {
```
takeshi-yoshimura, line 16: I know this PR focuses on a CUDA specific configuration, but the macro currently derives every symbol with stringify!($name), so for example, it cannot load HIP equivalents such as hipHostMalloc without duplicating or restructuring the function table.

bindings/python/src/engine/cuda.rs
```diff
@@ -0,0 +1,338 @@
...
+
+pub fn api() -> Option<&'static CudaApi> {
+    static API: OnceLock<Option<CudaApi>> = OnceLock::new();
+    API.get_or_init(|| unsafe {
+        let h = open_cudart();
+        if h.is_null() {
+            return None;
+        }
+        let f = Fns::load(h)?;
+        let sym = CString::new("cudaGetErrorString").unwrap();
```
takeshi-yoshimura, line 271: Is cudaGetErrorString loaded separately only because cuda_fns! assumes every function returns Err? Possible extensions in near future like GDS/HIP will probably require different symbol names as well, so it may be worth making the function-table loader handle explicit symbols and arbitrary signatures.

bindings/python/src/engine/cuda.rs
```diff
@@ -0,0 +1,338 @@
...
+                }
+                eprintln!("skipped: could not load CUDA runtime");
+                None
+            }
+        }
+    }
+
+    #[test]
+    fn test_idempotent_api_load() {
+        let a = api();
```
takeshi-yoshimura, line 303: The current tests only cover repeated api() calls and standalone CUDA operations. Could we add subprocess tests for both load orders: framework/vLLM before safetensors, and safetensors before framework/vLLM? dlopen behavior can differ once multiple CUDA consumers coexist, and OnceLock prevents testing both orders reliably in one process.

bindings/python/src/engine/unix.rs
```diff
@@ -0,0 +1,906 @@
+use std::{
```
takeshi-yoshimura, line 1: Most of this file appears platform-independent; only the positional file read is Unix-specific. Could we keep LoadPlan, Loader, iteration, and sink orchestration in a common module and isolate read_exact_at behind a small platform abstraction?

bindings/python/src/engine/windows.rs
```diff
@@ -0,0 +1,66 @@
+use std::fmt::Display;
```
takeshi-yoshimura, line 1: This appears to be an unsupported-platform stub rather than a Windows implementation. Could it be named accordingly and gated explicitly with cfg(windows)? Most of the engine is platform-independent, so moving the common logic out of unix.rs would also leave only the positional-read implementation to add for Windows.

bindings/python/src/engine/mod.rs
```diff
@@ -0,0 +1,12 @@
+#[cfg(unix)]
```
takeshi-yoshimura, line 1: cfg(unix) currently enables the CUDA engine on macOS, while cfg(not(unix)) routes every other platform through a module named windows. Could we separate OS-specific file I/O from accelerator backends and use explicit target cfgs?

bindings/python/src/lib.rs
```diff
@@ -709,14 +712,39 @@ impl Open {
...
+                    return Err(SafetensorError::new_err(
+                        "prefetch is only compatible with cuda devices",
+                    ));
+                };
+                Some(Loader::load(
+                    file.clone(),
+                    &metadata,
+                    offset,
+                    device_idx as i32,
+                    8,
```
takeshi-yoshimura, line 736: Prefetch currently allocates GPU storage ahead of consumption and has no memory budget or backpressure, so a live loader may retain the entire file on the GPU. tensor_stream() also keeps the loader alive when iteration stops early.

Also, each prefetched file currently creates eight dedicated worker threads, so concurrently opening N shards may create 8N workers. The shared slab pool limits transfer concurrency but not thread creation, file I/O scheduling, or GPU allocation across loaders.

The main concern is that these resource limits are currently hard-coded and cannot be controlled by the caller. Could we expose configuration for worker count and pinned-memory capacity? A higher-level multi-file helper can be added later, but it needs these knobs to bound total concurrency and memory usage safely.

bindings/python/src/lib.rs
```diff
@@ -2038,6 +2168,70 @@ fn torch_storage_shape(dtype: Dtype, logical_shape: &[usize]) -> PyResult<Vec<us
...
+    let shape_i64 = storage_shape.iter().map(|&n| n as i64).collect();
+    let view_target = if dlpack::dlpack_supported_native(dtype) {
+        None
+    } else {
+        dlpack::torch_view_target(dtype)
+    };
+    let dl_dtype = if view_target.is_some() {
+        dlpack::uint8_dlpack()
+    } else {
+        dlpack::dtype_to_dlpack(dtype)
```
takeshi-yoshimura, line 2218: torch_view_target() returning None does not indicate an error; unsupported types fall through to here, which maps them to OpaqueHandle. Could we distinguish “native DLPack”, “uint8 view workaround”, and “unsupported”? Also, can we fail explicitly before starting prefetch?

takeshi-yoshimura (commented):
Hi, @McPatate. This is cool!

Here is my review based on lessons from my fastsafetensors development. In my understanding, there is still room for speeding up load and reducing memory, but this is a nice step towards them. I would like to know which hardware environment this PR focuses on especially the number of CPUs, GPU memory capacity, and storage speed.

bindings/python/src/engine/cuda.rs
McPatate: Good point, although I think we can address this once we add AMD support!

bindings/python/src/engine/cuda.rs
McPatate: Interesting too, I think we can tackle that further down the road!

bindings/python/src/engine/unix.rs
McPatate: This was done temporarily to make sure nothing crashes on windows and make clippy happy. It's obviously not definitive and once and if we tackle windows, we'll merge everything back in the same module and keep the platform dependant code separate

bindings/python/src/engine/cuda.rs
McPatate: This is fallback in case we didn't find anything in `/proc/self/maps` first. We reopen the already mapped libcudart with `RTLD_NOLOAD`. We also always import torch before creating the loader, so we should always attach to the same runtime.

bindings/python/src/engine/cuda.rs
McPatate: I don't think the issue you're mentioning is possible with the existing API, I don't think we need the test

bindings/python/src/engine/mod.rs
McPatate: > cfg(unix) currently enables the CUDA engine on macOS

there are some cases where this is still relevant, although old intel + nvidia macs are getting rarer 😄 

and for the rest of the split, I'll change it to `unsupported_platform_stub` to make it more explicit

bindings/python/src/lib.rs
McPatate: The idea for this is atm we put the responsibility on the consumer to control how many files they want to open at any given time, this is how you do resource management.

Thinking of adding a process-global worker/thread pool later too which will definitely help, also the multi-file API entrypoint is already on my todo!

bindings/python/src/lib.rs
McPatate: good catch, fixing

bindings/python/src/engine/cuda.rs
```diff
@@ -0,0 +1,343 @@
...
+
+    pub fn set_device(&self, d: i32) -> Result<(), CudaError> {
+        self.check(unsafe { (self.f.cudaSetDevice)(d) })
+    }
+
+    pub fn device_guard(&'static self, d: i32) -> Result<DeviceGuard, CudaError> {
+        let mut prev = 0;
+        self.check(unsafe { (self.f.cudaGetDevice)(&mut prev) })?;
+        self.set_device(d)?;
+        Ok(DeviceGuard { api: self, prev })
```
danieldk: So what if you make two device guards and drop them in a different order? Don't we need e.g. a move or a mutable borrowing to avoid that the caller cannot call it twice and mess up the ordering?

(That or a e.g. mutex-guarded stack to ensure correct reversal order.)

bindings/python/src/engine/cuda.rs
```diff
@@ -0,0 +1,343 @@
...
+
+impl Drop for DeviceGuard {
+    fn drop(&mut self) {
+        let _ = self.api.set_device(self.prev);
+    }
+}
+
+#[cfg(target_os = "linux")]
+fn find_loaded_cudart() -> Option<CString> {
+    let maps = std::fs::read_to_string("/proc/self/maps").ok()?;
```
danieldk: Alternative:

https://man7.org/linux/man-pages/man3/dl_iterate_phdr.3.html
https://docs.rs/libc/latest/libc/fn.dl_iterate_phdr.html

Probably works also if systems have `/proc` hardening.

PR #848, opened 2026-09-16
feat: support prefetch plan

McPatate:
new `safe_open::prefetch(plan)` API

this allows users to specify slices of tensors they want to get, akin to `get_slice` but declared once ahead of time.

This yielded massive perf boosts on loading glm5.2, x3 from the previous prefetch-load-all-for-every-rank (which makes sense, thanks to that we cut the I/O by a large chunk).


bindings/python/py_src/safetensors/__init__.pyi
```diff
@@ -193,10 +205,34 @@ class safe_open:
         device=...,
         *,
         backend: str = "mmap",
-        prefetch: bool = False,
     ):
         pass
 
+    def prefetch(self, plan: Optional[PrefetchPlan] = None) -> None:
```
danieldk, line 211: I have looked mostly from a high level at the API (can a do a longer review of the details when my brain is less fried from a cold 😁). I think separating out a prefetch as a function is better than as an init argument, however since `get_tensor` can only be called once per handle, I think the API does not really help avoiding issues.

I would rather expect something along the lines of:

```python
f = safe_open(self.path, framework="pt", device="cuda:0", backend="pread")
with f.prefetch() as p:
   # Fine:
   f.get_tensor("fp32_2d")
   f.get_tensor("fp32_2d")

   # Fine:
   p.get_tensor("fp32_2d")

   # Breaks
   p.get_tensor("fp32_2d")
```

Similar for the Rust side, where the `prefetch` method could return `Prefetch` object.

I think it's generally best to avoid cases where a container object becomes (sorta) invalid after calling a method.

McPatate: Sorry just saw your comment! We're aligned, I've made the change in #859 !

